PR #4, opened 2017-05-04
added folder for vocabulary files

mariapoveda:
added folder and cvs file for the used namespaces


mariapoveda:
Added the ttl code according to the proposed model. Still pending:
- make a decision about the href attribute in DataSchema
- add contributors and metadata, change publisher
- work on the HTML documentation and exammples

sebastiankb:
Many thanks for updating the ontology :-)

Just 2 short comments /  questions:
1) the diagram seems not to reflect your latest updates, right? https://github.com/mariapoveda/wot-thing-description/blob/master/OnToology/vocabulary/td.ttl/diagrams/ar2dtool-class/td.ttl.png

2) for simplicity, it would be good if we only have a folder called "ontology" that contains only  the turtle file + a UML / figure diagram in png format (can be reused for the TD specification). Would it possible to modified your pr?

mariapoveda:
Hi, 

we can ignore the contents of /diagrams/ for now and keep using the UML alike one.

Regarding "2" I think we also need the HTML documentation to be published together witht he ontology and to include there the UML diagram. What do you think?

sebastiankb:
> we can ignore the contents of /diagrams/ for now and keep using the UML alike one.

I think, the simple way would be if you update your pr and remove all not needed information / files. In general, this github space is the official place where the latest information about the Thing Description should be available. Alternatively, I can merge your pull pr and make the changes. You can decide how we proceed.

> Regarding "2" I think we also need the HTML documentation to be published together witht he ontology and to include there the UML diagram. What do you think?

I think, it is not necessary. The model explanation should be done here on the official page: https://w3c.github.io/wot-thing-description/ 
I would like to avoid multiple places where you can find some explanation of something. We have already the current practice document where you can find the latest approach for the PlugFest. Based on the success of the PlugFest parts of the current practice document are taken over into the official Thing Description document. If you have time for the new TD ontology description I would like to ask you to make a pr on the current practice document. What do you think?

mariapoveda:
I don't see how the content negotiation will be done to provide HTML for the ontology terms having it merged with the current practices. Maybe the ontology will have a section or appendix to describe the terms? We can decide that in a call and ask also @draggett

draggett:
We can use HTTP's support for negotiating content types so that a client can request a resource in turtle rather than HTML.  The HTML resource provides human meaningful definitions and can describe the different vocabularies and link to their specifications. For the work that I was doing with Linked Data for the Web of Things, there is a need to define a number of new RDF concepts. There is also a need to define contexts for mapping terms in JSON to terms in RDF.

Note that we can introduce sub-directories or paths for grouping terms as needed.

mariapoveda:
> We can use HTTP's support for negotiating content types so that a client can request a resource in turtle rather than HTML. 

Right, my question is about how to mix that HTML with the documentation at https://w3c.github.io/wot-thing-description/ . Will they both share namespace? I might have misunderstood it.

PR #28, opened 2017-08-18
Added proposed alignment to W3C&OGC SOSA/SSN as per ACTION-115

maximelefrancois86:
See also https://github.com/w3c/wot/pull/348


sebastiankb:
Wrong repository? I think this is related to Web of Things Linked Data and Semantic Processing IG, right?

maximelefrancois86:
Yes indeed. This needs to be validated first within the IG.

PR #54, opened 2017-09-14
Proposed specification for input/output type definition

vcharpenay:
See [`ontology/td-types/`](https://github.com/vcharpenay/wot-thing-description/tree/master/ontology/td-types).


vcharpenay:
It now also includes a def for 'observable' as decided in last week's telco.

https://www.w3.org/2017/10/06-wot-td-minutes.html

vcharpenay:
As per last TD call, I reverted the last commits (introducing 'observable'). The changes have been moved to the Current Practices document.

PR #72, opened 2018-01-10
link -> form

mariapoveda:
Link replaced by form
Delete prefixed not used

### DIFF
--- a/ontology/td.ttl
+++ b/ontology/td.ttl
@@ -1,362 +1,277 @@
 @prefix : <http://www.w3.org/ns/td#> .
 @prefix dc: <http://purl.org/dc/elements/1.1/> .
 @prefix vs: <http://www.w3.org/2003/06/sw-vocab-status/ns#> .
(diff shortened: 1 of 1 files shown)


sebastiankb:
Hi Maria,

many thanks. One point: There is still 'link' as label in

` rdfs:label "form"@en , "link"@en .  `

I think, 'link' can be removed here.

Another favor would be if you can introduce the new optional element "rel" into the form class.  In the TD core model of TD4Prague branch I have already embedded this: [TD4Prague ](https://rawgit.com/w3c/wot-thing-description/TD4Prague/index.html#overview)

The description text can be taken from the latest version of the [WoT Protocol Template deliverable](https://rawgit.com/mjkoster/wot-binding-templates/initial-spec-contribution/index.html#relation-types).

Would it be possible that you update the PR to the TD4Prague branch?

Many thanks
Seb

mariapoveda:
Done. 

I've changed the Link class and rel to Form, however I think it might be a good idea keeping the old "link" ones as deprecated for informative purposes. What do you think?

PR #97, opened 2018-02-20
changed URI of w3c-wot-td-context.jsonld

takuki:
https://w3c.github.io/wot/w3c-wot-td-context.jsonld was changed to:
https://www.w3.org/ns/td/w3c-wot-td-context.jsonld

### DIFF
--- a/index.html
+++ b/index.html
@@ -35,22 +35,23 @@
                 }
               ]
             , localBiblio: {
-                "draft-fge-json-schema-validation-00": {
-                  title:    "JSON Schema: interactive and non interactive validation"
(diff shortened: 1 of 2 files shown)


sebastiankb:
Thanks for doing the changes. However, we should wait until @draggett has fix the redirection issue. The major idea is that https://www.w3.org/ns/td/w3c-wot-td-context.jsonld points to the latest working assumption that can be find here https://w3c.github.io/wot/w3c-wot-td-context.jsonld.

draggett:
I believe that this will need help from the W3C Systems team due to the challenge of interaction between the apache configuration rules.

PR #102, opened 2018-03-06
WIP: Simplified TD with experiemental JSON-LD 1.1 features

mkovatsc:
See https://rawgit.com/w3c/wot-thing-description/simplified-td-1/proposals/simplified-td/index.html

The array version for Interactions was dropped.

The Example 4 now explores JSON-LD 1.1 applied to the Data Schema, potentially making it JSON Schema compatible. There are still some issues.

### DIFF
--- a/proposals/simplified-td/index.html
+++ b/proposals/simplified-td/index.html
@@ -612,7 +612,7 @@
           <p>
             On the top level, <code>@id</code> will become important for
             security and identity management. It identifies the instance
-            of Thing and also serves as the <code>@base</base> for all
+            of Thing and also serves as the <code>@base</code> for all
             other <code>@id</code> definitions (e.g., of Interactions).
             Only this fully enables the object notation of Interactions
             (opposed to the previous array notation).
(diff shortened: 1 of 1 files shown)


proposals/simplified-td/index.html
```diff
@@ -612,19 +635,65 @@ <h3><code>@id</code></h3>
...
+          </p>
+
+          <pre class="example" title="Simplified TD Example">
 {
   "name": "lamp",
   "@id": "urn:dev:wot:com:example:servient:lamp",
   "base": "https://servient.example.com/",
+  /* optional, only if wanted, cf. Dave's proposal! for free by JSON-LD */
+  "definitions": {
+    "range": {
```
zolkis: maybe the name could be less generic, e.g. range_0_100

mkovatsc: It is up to the TD author how they want to name their custom types in the TD. It is just an optimization to not repeat definitions (=schemas).

When people start to define "public types" that would be imported via an external context (e.g., domain-specific .jsonld file given), they should put such thoughts into the term design.

benfrancis:
Interesting!

benfrancis:
Just a note to say that Mozilla's WoT implementation is moving somewhat towards the direction of this experimental simplified example. Our [latest internal spec](https://iot.mozilla.org/wot/) uses the array syntax for the top level [`links`](https://github.com/mozilla-iot/wot/pull/74) member and the JSON Schema syntax for property `type`, [action `input` and event payloads](https://github.com/mozilla-iot/wot/pull/80).

It doesn't currently use the HyperSchema [`links`](https://github.com/mozilla-iot/wot/issues/73) for properties and events or [`forms`](https://github.com/mozilla-iot/wot/issues/79) for actions but we are discussing that.

We'll be testing out these changes in our NodeJS, Python, Java, C++ and Rust implementations to see how it works out.

We don't currently implement an `@id` or `base` member.

mkovatsc:
Sorry for the chaos caused by the accidental merge. I am a bit too short on time the moment to clean this up.

@benfrancis Thanks for your feedback. I definitely will continue working on the "experimental Simplified TD".

Issue #123, opened 2018-04-09
Renaming of the term 'name' at the interaction level in 'label'


sebastiankb:
JSON-LD 1.1 allows the use of properties, actions, and event names as keys. Therefore, the _name_ term is obsolete at the interaction level. However, key names can not be human-readable (for example, no spaces allowed), the (optional) term _label_ can be used as an alternate interaction name and used, e.g., for UI labels.

benfrancis:
I think the term "name" is fine for human-readable strings. I like that "name" and "description" can be consistent across properties, actions, events and the top level of the thing. "name" is what we are currently using in Mozilla's implementation.

sebastiankb:
ok, thanks for this input. We will discuss this in one of the next TD web meeting.

takuki:
My understanding was that JSON key names can contain special characters such as spaces. This is purely from syntactic viewpoint. Are there any other reasons that we should consider to avoid spaces in names? (e.g. Scripting API)  

A name can be considered to be part of a definition, and it can also be considered external to the definition.  When it is important to distinguish a "type" from a named use of the "type", it should be considered external so that a definition of the type can be shared among multiple named uses of the "type". 

On the other hand, I agree that consistency (e.g. Thing and interaction patters) is also an important factor.

takuki:
The TD TF discussed the issue today. The Thing level "name" should really be "label", and the TD no longer has "name" field. Thing, Property, Action and Events this way will only have ID and optional "label". The TF discussed at lengths and found it prefers to use "label" for human-readable statement than to use "name".

benfrancis:
Can you explain the reason this change of terminology is "preferred"?

It seems natural to give web things a human readable "name" and "description", as is also the case for web apps (https://w3c.github.io/manifest/#name-member). This is distinct from a machine readable unique identifier in the form of a URI.

zolkis:
At top level (Thing) I'd also prefer `name`.

On interaction level (Property, Action, Event), the interactions are represented as dictionaries, where the keys are the interaction names, the values are the representations. Therefore inside the representation `label` is used instead of duplicating `name`. Also, a `label` can be internationalized. 
I guess the reasoning was to harmonize with the convention `label` is used at interaction level.

My only gripe with `label` is either that why limiting the number of labels to one, or (in another interpretation) why are we using the term `label` for something that really is a name.

mkovatsc:
[...]

PR #139, opened 2018-05-18
Thing class should have "label" instead of "name".

takuki:
This is for addressing issue https://github.com/w3c/wot-thing-description/issues/123.


zolkis:
So we need to update the Scripting API as well.

PR #150, opened 2018-06-06
TD and datatypes diagram

mariapoveda:
Proposal of TD and datatypes diagram and questions about previous model


mmccool:
Note that the pending security metadata PR (https://github.com/w3c/wot-thing-description/pull/144) also modified the diagrams.  Therefore, we have a conflict between these two PRs.   Also, if we include the security metadata, the overall diagram will get a lot bigger (and busier).   It's not just two "modules", but three.

I personally think three separate diagrams (one for the Thing, then one each for DataSchema and SecurityScheme) is fine and easier to read (not as cluttered).   That said, a number of the other improvements make sense but probably should use the new diagrams in the Security Metadata PR as a baseline.

mariapoveda:
Thanks for the pointer @mmccool I wasn't aware. We can have indeed one diagram per module, 3 in total, which makes sense. 

Usually with ontology networks I also try to keep a common view, so that you see the whole picture, even if it is not as complete as the join of every module, but that is optional.

mmccool:
I think the first diagram (for the Thing) gives the big picture, since it includes placeholders for the other two classes (DataSchema and SecurityScheme).

mmccool:
BTW right after I made the above comments I realized I forgot to fully update the diagrams in the security PR.  Done now :-P.

sebastiankb:
@mariapoveda Thank you for this input. I would like to look together in tomorrows TD web meeting. I hope, you are joining.

PR #170, opened 2018-07-20
Automate Normative Assertion Extraction, Round 2

mmccool:
An improvement on my previous attempt to automate normative assertion extraction.  This version...
1. now processes the HTML structure instead of just using grep, so the normative assertion can be broken across lines
2. does not require the rfc2119 category to be marked in any special way (except being in caps)
3. stores the test specification in a separate file

I have only marked up a few assertions and have actually included some intentional omissions (missing test specs, missing ids) to check my error checking.   After (if) this PR is accepted, someone needs to go through and mark up all the normative assertions.  [Update: all assertions now marked up]

See testing/plan.html to see the output.


mmccool:
I marked up the normative assertions in index.html.template.  However, this basically only captures the normative assertions for serialization, not the various assertions required for eg default and mandatory values, which are all defined in tables that are automatically generated.  In the spec document itself we also need RFC2119 assertions for these things.

PR #171, opened 2018-07-26
Restructured JSON Schema and Security ontology modules

vcharpenay:


mkovatsc:
Needed to clean up the namespaces for modular vocabularies - merging.

Issue #165, opened 2018-07-19
Should "security" be mandatory?


mmccool:
We should decide whether "security" should be mandatory.   This can be combined with a "none" scheme.  The intent is that Things with no security will still have to explicitly declare they don't have any.

My main concern with this is whether we can combine it with multi-level overriding.   In particular I'd like it so that if a "security" tag is given at the Thing level, that satisfies the "mandatory" requirement and it then does not need to be given at the "forms" level.

This would make it clear that a Thing has no security (if that is the case) without adding an extra security definition to every form.

For slightly more information, see the last slide in [this presentation](https://github.com/w3c/wot-security/blob/master/presentations/W3C%20WoT%20Security%20Definitions%20Proposal.pdf).

mkovatsc:
As commented during the Bundang F2F, I prefer making `security` mandatory and require "no sec" to be declared explicitly, e.g., with scheme `none` or `nosec`.

benfrancis:
Is the idea that a client should refuse to parse an otherwise valid Thing Description if it doesn't define a security member?

What's the rationale for not just assuming "none" as a default?

takuki:
Another option is we state Security is a should-have, and warn that it should be specified for interoperability reasons. We do not need to specify what it means by default when there is no security specified.

mmccool:
I think there is also a difference in the requirements for H2M and M2M use.   For M2M use, we want to know for sure in advance all the interface requirements, including security.   If something unexpected happens at runtime, and in particular if an unanticipated request for credentials occurs without a human in the loop, it would likely break the system. For H2M, asking for credentials "unexpectedly" just means a small diversion in the browser to prompt for credentials.

So one thing we could do here is "RECOMMEND" (RFC2119 style) the of a "security" declaration for any system meant to be used in an M2M fashion, and should consider the TD a "binding contract" so if we declare "none", it really should mean "none".  Leaving "security" off should only be used for Things that are not used as part of an M2M system.

I *personally* think making "security" mandatory, and "binding", would be safer and would support better interoperability.  This is because even if a Thing is designed for H2M use, *someone* is going to want to use it in a M2M use case, if only for automation.

mmccool:
@takuki , your comment is not entirely off-topic since if we make security mandatory, then one form of “incomplete” TD would be one without a security definition.   At any rate, the going name for incomplete TDs seems to be “TD template”.

danielpeintner:
[...]

PR #174, opened 2018-07-31
Make security configuration mandatory

mmccool:
As discussed and resolved in Issue 165 (https://github.com/w3c/wot-thing-description/issues/165), this PR specifies that security configuration should be mandatory.   Technically this means that each form needs a security configuration; this is combined with overriding however, so in practice there is not a significant additional verbosity burden.   Unfortunately this style of "mandatory" cannot be covered by local rules and so I have left the tables and diagrams showing "security" as optional and given the additional rule as an additional normative statement.

This PR modifies a number of examples to include the now-mandatory security, and also updates existing examples to include proper use of the new "none" scheme.   The "psk" and "none" schemes are also mentioned in the list of schemes allowed in the definition of allowed values for that tag.

There is one other addition in this PR: there is now a blanket RFC2119 statement that all restrictions given in the tables are normative.  We may want to discuss how to deal with this in a better way.


mmccool:
Regarding "does this reflect the Bundang plugfest": not exactly, it is a little more restrictive.  In particular there were many TDs without security configurations at the plugfest.  However it would be easy to update them, and we can also allow node-wot to issue just a warning if it sees TDs without a security configuration, so on balance I think we should merge this and get it into the next release.

benfrancis:
@benfrancis wrote in #165:
> Is the idea that a client should refuse to parse an otherwise valid Thing Description if it doesn't define a security member?
> What's the rationale for not just assuming "none" as a default?

@mmccool I can't tell whether you answered this question or not. What is the harm in assuming a default of "none"? Making security mandatory makes all existing Thing Descriptions invalid and it seems like unnecessary verbosity if the default is in fact no security. My recommendation is that it's optional.

sebastiankb:
As discussed on Wednesday in WoT web meeting this PR should come into "bundang" version. 

@benfrancis The idea is to increase the awareness of security in IoT application and that developer has to select explicit "none" security if really no security handling is needed. 

For continuing this discussion please use one of the existing [issues ](https://github.com/w3c/wot-thing-description/issues?q=is%3Aissue+is%3Aopen+label%3Asecurity) or create a new one.

PR #177, opened 2018-08-02
Rename "pname" parameter to "name" for API Key (and other security schemes)

mmccool:
Upon request from Hitachi, adding "name" parameter to API key to indicate custom header when "in" equals "header".  Actually, an appropriate parameter already existed, but was called "pname" (in order to avoid a conflict with "name" used as the name of a security configuration... not a problem with the current structure).   This PR simply renames "pname" to "name" to do what people expect.   It is in the security namespace, however, so no conflicts with other uses of "name" exist.


vcharpenay:
I included this change in #180, too. This PR can be closed without merging (to avoid conflicts).

mmccool:
Reviewed changes, doesn't look like there are any other changes not covered by VC, so closing this PR to avoid confusion.

PR #183, opened 2018-08-03
Security example fixes

mmccool:
As discussed in the TD meeting, this PR fixes various things:
1. Examples 1 and 2 now use "basic" authentication with "https" and point readers at Section 7 for coaps examples.
2. Examples in section 7 now use the default coaps port (that is, no port is given in the URL) and specifiy the "psk" scheme (with no identity)
3. Ontology has been edited so "psk" scheme no longer has a "psk" field.
4. Figure 3 updated to include "none" and "psk" schemes.   It also converts "pname" to "name" assuming PR 177 (https://github.com/w3c/wot-thing-description/pull/177) will also be merged.
Note done (yet): expanded descriptions of scope; emphasizing the special rule for "security" in information model section.


mmccool:
Please go ahead and merge both these and PR177.   I want to deal with the scope and mandatory note in separate PR because I'm worried about this PR and PR177 conflicting.   So I'd like to deal with those issues after they are merged...

sebastiankb:
I cannot merge the PR due to the conflicts. Please can you check this.

mmccool:
Fixed merge conflicts.   One was in index.html and was probably due to the fact the render produces huge single lines that confused git... anyhow, "npm run render" fixed it.   The other was due to a renaming of security.ttl to wot-security.ttl; I deleted my copy of security.ttl to resolve the conflict.
Not quite ready to merge... will try to address the other issues noted above first.

mmccool:
I fixed the issues raised by Sebastian in Figure1 (it now uses anyURI) and the indentation problem in Example 15 (tabs are evil).   It looks like the OAuth->OAuth2 change was already taken care of; it is also correct in the Figure.   If you see any instances of plain "OAuth" please change them to OAuth2.

I did make one more change: NoneSecurityScheme -> NoSecurityScheme.   "None" just sounds wrong here (but "none" is still used as the short form; grammatically, this makes sense, because English).   This change affects the ontology, the validation, and the render.js.   Actually, in the latter I added a rule (which was not there before) to sort the "NoSecurityScheme" so it now appears first in the lists of schemes.

Ready to merge.   In fact, I'm going to ASSUME it's going to be merged and am starting another branch with a set of edits to handle security definitions, as discussed.

mmccool:
BTW, one more change I am contemplating (but have not implemented) is removing "Url" from all the names that use them, eg. authorizationUrl, proxyUrl, etc.   For a few reasons: the Url term is inconsistent with URI; making things shorter; consistency with other items that have URI values.   The "Url" terms come from OpenAPI, but we don't have to follow this slavishly.

However, I don't want to make this change without first getting input from people, and it should not be done in this round anyway (which is supposed to be Bundang-compatible).

sebastiankb:
many thanks for the changes!

PR #188, opened 2018-08-08
Removed superfluous "SimpleDataSchema" class

vcharpenay:


sebastiankb:
I cannot merge since there is a conflict. Can you check this. Thanks

PR #192, opened 2018-08-13
Move list of allowed rel values to ed note

mmccool:
The attribute rel is used in both forms and links, but the list of restricted values for rel only belong in forms.  Declaring the restrictions in the ontology applies it in both places, which is wrong.

This PR moves the list of restricted values to an Ed note for now.  However, we need to fix this properly, probably by having different definitions for "rel" for forms and links.  Right now the description of rel is also specific to forms, and needs some differentiation too, but for now I have only reverted it.

I note as well that the domain of rel is listed as Link, which is also an issue...


sebastiankb:
thanks

PR #195, opened 2018-08-13
Editorial changes

mkovatsc:
* Clarified default values (not from informational WoT Binding Templates)
* Fixed Eample 1 (Event over HTTP needs `subProtocol`)
* Fixed Action and Event assertions on "collecting" in respective fields (`actions`, `events`)
* Added experimental CoAP Content-Format (65100)
* Added reference to JSON (RFC8259)
* Improved wording on JSON terms
* Clarified serialization is JSON that allows JSON-LD processing
* Changed JSON-LD "keys" to "keywords" (correct wording from JSON-LD spec)
* Added Editors' Note on CBOR as potential default media type
* Fixed section linking (don't give text within `<a></a>`, so that ReSpec fills it in)
* Made JSON keys to appear as `code`

### DIFF
--- a/index.html
+++ b/index.html
@@ -134,10 +134,9 @@
   Thing Descriptions provide a set of interactions based on a small vocabulary 
   that makes it possible both to integrate diverse devices and 
   to allow diverse applications to interoperate. 
-  Thing Descriptions, by default, are encoded in JSON-LD. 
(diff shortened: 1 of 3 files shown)


sebastiankb:
Thanks.

PR #198, opened 2018-08-17
Add CoAP/DTLS "cert" and "public" security schemes

mmccool:
Add two other CoAP/DTLS security schemes defined in the CoAP standard.   
**Not yet ready for merge:**
1. Parameters may be incomplete.   Just using "identity" as with the "psk" scheme
2. HTML is not rendering properly.  New enumerated values and sections not showing up.

Note: I noticed while editing this that "psk" was missing in a few places too.


sebastiankb:
many thanks

PR #199, opened 2018-08-17
Add subProtocol

mmccool:
Add new key to "form" to allow specification of sub-protocols.
1. For now the only valid value is "LongPoll"; we can add more as necessary.
2. Unlike other enumerated values, uses camel-case.
As an aside, also fixed some missing "psk" enumerated values for SecurityScheme and got rid of the redundant list of schemes in the description.


mmccool:
Going to go ahead and merge, since we already discussed these changes.

PR #203, opened 2018-08-22
Add cert and public security schemes to validation

mmccool:
Fix an oversight in a previous PR.   I forgot to add shape definitions for the cert and public security schemes to the validation files, so as a result the new sections did not get created in index.html.   This adds those definitions and a rendered index.html file with the new sections talking about the cert and public schemes.

While I was at it, I reran the assertions extractor and updated the testing plan.


sebastiankb:
PR looks ok. Do you want to use the same PR for the JSON schema update?

mmccool:
Sure, I think the JSON schema update is a small change, so I'll work it into this PR as well.

mmccool:
Done: subProtocol now listed as an optional element under form_element in JSON Schema.  Ready to merge.

sebastiankb:
thanks

PR #205, opened 2018-08-27
Oracle Review Comments on TD spec draft

mlagally:
Oracle review comments on release candidate for virtual plug-fest.

This PR contains various bug fixes as well as multiple review comments.
All review comments are enclosed in <span color=red> sections.

Various fixes in the TD model in section 5 must be retrofitted to the ontology model.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,3 @@
+# Code of Conduct
+
+All documentation, code and communication under this repository are covered by the [W3C Code of Ethics and Professional Conduct](https://www.w3.org/Consortium/cepc/).
--- a/assertions.js
+++ b/assertions.js
@@ -48,13 +48,14 @@ src_dom('span[class="rfc2119-assertion"]').each(function(i,elem) {
 
 // Merge assertions and test specs into plan
 plan_dom('head>title').append(src_title);
-plan_dom('body>h2').append(src_title);
-plan_dom('body').append('<dl></dl>');
+// plan_dom('body>h2').append(src_title);
+// plan_dom('body').append('<dl></dl>');
 for (a in assertions) {
     console.log("Processing assertion "+a);
 
-    plan_dom('body>dl').append('<dt></dt>');
-    let plan_dt = plan_dom('body>dl>dt:last-child');
+    // Appendix
+    plan_dom('#testspecs').append('<dt></dt>');
+    let plan_dt = plan_dom('#testspecs>dt:last-child');
     plan_dt.append('<a href="../index.html#'+a+'">'+a+'</a>');
 
     let category = undefined;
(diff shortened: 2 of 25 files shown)


mmccool:
Unfortunately the updates need to be on index.html.template.html since "npm run render" generates index.html from other files, so this PR would be overwritten the next time someone executed that.   Some of the changes might also have to be made in other places, e.g. vocabulary changes might have to be made in the files under "ontology" or "validation".

mlagally:
@mmccool Thanks for pointing out. This PR is not meant to be merged, but as a single file to collect  all review feedback.  I suggest we review the PR as a whole and discuss which changes to incorporate, since I'm not sure that the tooling for generating the vocabulary chapter is available.

mlagally:
@takuki There are significant changes also in this section, which should be reviewed. 
Unfortunately the diff tool cannot ignore line breaks and the original index file contains
only a few very long lines. I'll try to create a better usable diff.

mlagally:
I have created a PDF version of the reviewed spec which includes change marks.

It is available at:
https://github.com/mlagally/W3C-Work/blob/master/Oracle%20TD%20review.pdf

vcharpenay:
Few comments:

> (5.2) The references for the types don’t point to a proper type definition

They do, in fact, point to their machine-readable definition (the XML schema namespace). It is then, in theory, the responsibility of the namespace holders to redirect to the adequate human-readable documentation. Here, they chose to expose a plain HTML documentation with links to the XML schema normative specifications. If you think it is still not enough, maybe an explanatory note would help?

> (5.3.5, 5.3.7) Only applicable for associated number or integer types

I am not sure I understand your remark. The current specification precisely distinguishes between numbers (eq. to double-precision floating point numbers here) and integers for fields `minimum` and `maximum`. Do you mean that only the text should be changed?

> (6.1.1) This paragrah appears to be normative and not an editors note. Section 3 mentions RDF as an alternative format.. what accept header is required?

This part deserves clarification. It is indeed normative but it is also already included as such in the JSON-LD spec. It is probably not necessary to mention it again in the TD spec. Moreover, the JSON-LD and Turtle documents under http://w3.org/ns/td are of different nature, they do not contain the same information. One is a context file, the other is an ontology (or vocabulary) file. It would be good to consistently use either 'ontology' or 'vocabulary' throughout the document, to avoid confusion.

mlagally:
@vcharpenay 

#5.2. The spec will be read by humans, therefore also the links should refer to human readable information. I would expect to read a regular expression if I follow the link of anyURI.

#5.3.5 / 5.3.7
we should have another look at this section. Strictly speaking JSON does not have an integer type. 

#6.1.1
IMHO we should not require the reader to read the entire JSON-LD specification to make the section understandable. Perhaps an editor's note would be suitable to resolve potential confusion.

mlagally:
this PR is a mix of review comments from the last plug fest version and a new chapter for a thing model.
I'll create a new PR for the latter.

sebastiankb:
based on today's TD web meeting:
AI: Please add an example to have a better understanding how this can work. 
AI: Renaming "Thing model" to "Thing Template"

PR #207, opened 2018-08-28
Security Considerations section

mmccool:
This is just a rough draft, and I expect a lot of discussion, as it makes specific statements and normative assertions that I expect we will have to discuss before it can be merged.  This addresses the following issues:
https://github.com/w3c/wot-security/issues/70
https://github.com/w3c/wot-security/issues/61 
It also depends on an external document for "security best practices" that is still under development but that we will have to complete and publish as a note.


mlagally:
There are several normative requirements in this PR that require a lot of discussion, including the following issues:

1. user consent
Many devices won't have a UI, so there is no user consent possible. This is also the case for typical registration flows. 

2. changing the user id
The paragraph attempts to obfuscate the author of a TD by changing the ID. Security by obscurity seldom works.

3. exchange of TD via secure channels / hiding device interfaces by not exposing the TD
I would not assume these secure channels as the normal case. Again, security by obscurity does not work. Since we have various authentication schemes, a device can protect itself.

4. TD tampering
What is the attack scenario? What can be done with a tampered TD, if proper authentication is required from the device?

mmccool:
I agree; the point of this PR is to kick off these discussions, and hopefully put to bed a number of issues raised in wot-security at once.   I'm sure this PR will go through a number of revisions before it is mergable.  Note this PR is NOT intended to be merged into the Bundang version.

mmccool:
Anyhow, let me try to address each of these points:

1. User consent should be obtained when the device is installed, OR, if it is a question of how the data generated by the device will be used, at the point of subscription to data from the device.   It may not be the devices' job to obtain user consent itself, but it needs to be obtained somewhere in the system before the data is used, stored, etc.  Note that the assertion just says the law must be followed... which is pretty much a tautology.  The problem is that we can't say the laws "should" be satisfied, obviously, but we can't also make blanket statements as the actual requirements would vary from place to place.  However one substitution here is a set of more generic "should" assertions.   As a case study we should probably look at the EU laws and make sure we have some suggestions on how to satisfy them.  Leaving this out is not really an option: we have to say *something* about user consent.  We might be able to get away with making non-normative statements, though.

2. Note that this is to address privacy, not security.   The point of changing the id is that if a device is being tracked by id, then you might want to change the id to break the association of that device with that person (and with previous data associated with the device).   Of course this can "break" other associations as well, so a way for a device to notify registered (and trusted...) parties that its id has changed would be useful.

3. There are many risks (again, mostly to privacy, not security) that are amplified if TDs are visible publically.   For example, tracking, or characterization of a person's property. ("Hey, this person has a nice TV model, they would be worth robbing...").   So this recommends just that TDs be shared with authorized parties.  This could be implemented, for instance, by having a Thing Directory that requires credentials before it can be used, probably a good idea to avoid DOS anyway.  The other point of transmitting a TD by authenticated means is to prevent hackers from "faking" or modifying a TD (for example, setting up a fake device with the intent of capturing security credentials that a system could use to try to hack into other devices, or intercepting and tweaking a TD to add a "proxy" that is meant to log access and capture data).

4. Actually I mentioned a couple of attack scenarios resulting from the ability to modify a TD above.   Another one would be just to mess with people, for instance, swapping the "on" and "off" actions for a light by switching URLs in the TD.  Hilarity (and annoyance...) ensues.

mlagally:
@mmccool Thanks a lot for all your clarifications.

Let me add a few comments:

#1: It depends on the scenario, however I believe in many cases this user consent is not handled at *machine/device* level, rather during *service subscription by a human*, when a user has to pass a click-through and accept the terms of the service. We should not introduce normative requirements here.

#3: I agree with your concerns, however I'm not sure whether a thing directory will be a way to resolve it for the home use cases.
Assuming that there is a centralised directory, where a (home) user registers all his devices, you end up with a lot of questions:

- What is the business model for hosting a thing directory for the home scenarios?
- Who is hosting the thing directory for the home scenario, i.e. whom do you trust to know about all devices (including security systems) in your home? 
- Typically a thing directory needs fine grained access for different groups (e.g. one for your lamps, one for your kitchen devices, one for your security systems), who does the admin work?

Some of these questions becomes easier for industrial scenarios, where you have a trust relationship between exposed devices and consumers (cloud service).

You could consider the Oracle IoT Cloud Service as this centralised repository, where all devices of a service subscriber are registered and you can query device details, which includes the device model[s] via a REST API:
https://docs.oracle.com/en/cloud/paas/iot-cloud/iotrq/op-iot-api-v2-devices-device-id-get.html

mmccool:
1. My intent was only to say that user intent should be obtained, not how.  I actually mention both by the device and by the service using the devices as possible points where this can happen.  That said, we discussed this in the security call and the "legal" part of this is troublesome.  We may have to turn this into an informative statement: "There may be legal requirements in the use of personal information that require devices or the services using them to obtain user consent..."
3. I only mentioned the use of a Thing Directory as one possible implementation.  There are many ways TDs could be distributed securely.  If there is a key infrastructure in place already for example, or some other way to identify authorized users (eg possession of tokens), then devices can still self-host TDs as long as they only provide them to authorized users (as confirmed by the keys or tokens previously noted).  A central repository is not needed in this case.  I will have to think about whether additional examples should be given...

mmccool:
Last few commits have updated the consent normative assertion.

takuki:
WoT TD (Thing Description) is a JSON-LD document. Thefore, security consideration inherent to JSON-LD is also relevant to TD. JSON-LD specification has a [security consideration section](https://w3c.github.io/json-ld-syntax/#iana-considerations). In particular, it is advised that "any application that depends on a remote context for mission critical purposes vet and cache the remote context before allowing the system to use it".

mmccool:
Anyway, we discussed this at TPAC and I recommend we merge this as it stands.   There are more changes to be made but they can be layered on top of this basis.

mmccool:
Note also that "Security Definitions" are a completely orthogonal topic.  This is about the "Security Considerations", which is prose section "about" the use of the specification, and does not affect the actual content of the TD.

mmccool:
Although Taki, good point, your comment relates to the privacy discussion we had at TPAC.  I'll update the PR to reflect that point.

mmccool:
I added a point about the privacy caused by dereferencing @context and a mitigation (cachi)ng).  I did not add a reference to the JSON-LD 1.1 doc since it does not (currently) discuss this risk.  If the 1.0 doc does discuss this we could reference it instead...

mmccool:
Latest commit removes all SHOULD statements (since the normative SHOULD does not have quite the meaning we want here and has testing implications) and restructures context as a list of risks and suggested mitigations.  As the sections are quite long and we may want to reference them from elsewhere, subsections were added with unique ids to act as hyperlink anchors.

mmccool:
I also updated the name of this PR as I no longer consider it a draft, but a contribution which should be merged as soon as possible.

mmccool:
@takuki Re-reading your comment I realized that you are talking about the risk of someone *tampering* with the context files.   That is a security risk.   The comment I added was about the *privacy* risk of someone observing what you are referencing (allowing them to infer things about your devices).   I do mention another tampering risk about TDs themselves... maybe I can add a sentence to that risk, since tampering with the vocabulary files is similar.

The mitigation (caching, making context files immutable) is more like the mitigation for the privacy risk, however.

takuki:
@mmccool , it is all clear to me. Thank you.

mmccool:
conflicts resolved, ready to merge

sebastiankb:
there is a new conflict. Maybe caused by another merged PR.

mmccool:
Conflicts fixed.   Please merge (do scopes first, then this one, then definitions...)

mmccool:
I merged this myself with the permission of Sebastian so I could resolve conflicts (arising just from entries in the change log).

Issue #215, opened 2018-08-31
scopes should be array of strings


mlagally:
There are two sections where the description of the scopes indicates an array, however the type is a plain string.
Should this be an *array of string*?

mmccool:
Yes, they should be.   I'll create a PR to fix this.

PR #219, opened 2018-09-03
Make scopes an array of string

mmccool:
Address a mistake in the ontology/JSON validation identified by Michael Lagally: the value of "scopes" is supposed to be an array of strings, but was mistakenly declared as just a string.  This PR resolves [Issue 215](https://github.com/w3c/wot-thing-description/issues/215).  This PR also removes the "potentially" qualification from the normative assertion stating that "security" MUST be serialized as an array, like the other array-valued things.


mlagally:
Looks good!

mmccool:
Actually, I noticed a few more places where "potentially" was used and so I updated my PR to remove them.  Also, I added the requirement that "scopes" be serialized as an array to be consistent with other array-valued entities.

There is still one thing bothering me.  These normative assertions all start with "The type of ...".  Shouldn't that be "The value of ..." or "The content of ..."?

Issue #214, opened 2018-08-31
Interaction pattern - clarify unique name


mlagally:
>InteractionPattern 
> <snip>
>Each instance of a Property , Action , and Event class has a (unique) name. See Section Representation Format for more details about JSON-LD 1.1 identifiers. 

Should be reworded to:

Each instance of a Property , Action , and Event class **MUST have** a unique name. See Section Representation Format for more details about JSON-LD 1.1 identifiers. 

It should also be clarified whether the same name can be used for actions and properties and events.

mmccool:
See [PR221](https://github.com/w3c/wot-thing-description/pull/221).  I also added "globally" in front of "unique", and updated the normative assertions extraction.

sebastiankb:
non active issue anymore. PR #221 was merged.

PR #221, opened 2018-09-03
Assertion for unique-within-a-TD identifiers for interaction patterns

mmccool:
This converts a "note" about the need for unique names for interaction patterns to a *normative assertion* about the need for *globally unique names*.
One issue: this assertion is in the vocabulary section; do normative assertions belong here?
Also, the term "globally unique" is meant to capture the fact that eg you cannot have a property and an action with the same name (I think), because JSON-LD puts all ids in the same namespace.   I'm not sure the term "globally unique" captures this and another (better) description may be needed.

This addresses [Issue 214](https://github.com/w3c/wot-thing-description/issues/214).


mlagally:
The language now asks for a "*globally*" unique name. I believe we want to say that all key names *within a TD* are unique.

vcharpenay:
I would also favor "unique within the scope of a Thing Description document". But it is not exactly true. Interaction pattern names actually resolve as IRIs (in RDF). The default behavior should be that the resolve relatively to the Thing's `id` and this behavior should be included somewhere in the spec.

However, so far, there has been no restriction on the characters allowed for a key name, which means that properties, actions and event keys may also be absolute IRIs or more complex relative IRIs that do not follow this default behavior. E.g.:
```js
{
  "id": "urn:example:something",
  "properties": {
    "urn:example:somethingElse": {} // absolute IRI
  }
}
```
```js
{
  "id": "urn:example:root/path/to/something",
  "properties": {
    "/somethingElse": {} // resolves as urn:example:root/somethingElse
  }
}
```

To me, there is no need to have a detailed description of all possibilities, it is enough to just mention that names resolves as (relative) IRIs and that the well-defined mechanism of RFC3986 applies. It is also probably a good idea not to restrict these key names since:
 - most programming languages have ways to deal with that;
 - it does not prevent simpler names to be used;
 - it provides greater modeling freedom to implementers, especially if they want to use the RDF representation of a TD natively.

mmccool:
I agree that "global" is too broad; it should be unique within a TD.  I like Victor's wording and will update the PR to use it.  Regarding use of URIs that ARE actually globally unique: this normative assertion would not forbid that, since actual globally unique names would also of course be unique within a TD.

takuki:
Merged. "index.html" still needs to be re-generated from index.html.template.html.

PR #264, opened 2018-10-29
Added axioms to the JSON schema vocab for readOnly, writeOnly

vcharpenay:
Following the decision to replace `writable` with `readOnly` and `writeOnly`.


mmccool:
I'm a little confused why we have both readOnly and writeOnly.  I only remember discussing readOnly at TPAC, not writeOnly.

I guess it would be clearer if we had both "readable" (can be read) and "writable" (can be written) and these are just the negations of those... but as far as I know we never had "readable".

sebastiankb:
@vcharpenay  please check the conflict

vcharpenay:
PR #304 merges changes on `writeOnly` (and other changes made since then).

PR #268, opened 2018-10-30
Remove URL postfixes from all security terms

mmccool:
Remove "Url" postfixes from all security terms as it is (a) unnecessarily verbose (b) inconsistent with usage elsewhere (other URI-valued terms, eg href, do not have postfixes) (c) it is confusing to use Url instead of Uri.

Note that this change is inconsistent with OpenAPI, but there are enough other inconsistencies that one more won't matter too much (more).

This is based on PR https://github.com/w3c/wot-thing-description/pull/265 so that should be merged first (if this one is merged, it will include all the changes in that PR, though).


sebastiankb:
Can you quick check and resolve the conflict?

mmccool:
conflict resolved (trivial issue in release notes), ready to merge

sebastiankb:
thanks

PR #277, opened 2018-11-05
Security definitions

mmccool:
Add basic definitions for a "securityDefinitions" section.  This is not yet complete however as I don't know how to declare that values in "security" array can be either strings (for using previously declared definitions) or objects (for in-place specification of security configurations).

This is based on top of the "security cleanup" PR (https://github.com/w3c/wot-thing-description/pull/277) so should not be merged before that one is (if it is, it will include all the changes in the security cleanup).


mmccool:
Ah... it would have been better not to merge this, as I said it was still under discussion and was not ready.   It was not really complete.  Oh, well....

mmccool:
I'm going to submit another PR to finalize this.   It will be based on this same branch.

PR #290, opened 2018-11-19
Even More Updates to Implementation Report Generator

mmccool:
NOTE: READY TO MERGE.  Updates to index.html.template.html have been limited to bug fixes:
1. fixing a duplicate of td-action-object as an id that should have been td-event-object.
2. adding securityDefinitions to the assertion listing what fields should be serialized as objects.

Original description:
This PR includes updates to the implementation report generator but ALSO includes updates to the text of the specification (to clarify some information that was missing when the assertions were put into the IR table) that at the very least needs to be carefully reviewed before merging.


mmccool:
I'm going to go ahead and merge this as others need the new structure under "testing" to provide input for the implementation report.   I have reverted my changes to the index.html.template.html to limit it to the two "bug fixes" noted in the opening comment.

PR #294, opened 2018-11-22
Yet another update to the implementation report generator

mmccool:
This update includes:
1. Merging of implementation descriptions into the generated report.   Note a div with the id is needed; see the template.  This also automatically adds the implementations to the TOC.
2. Extraction of uniquely named assertions for all the table entries.
3. A way to track dependencies, both general/specific and contextual (see plan.html and README.md for an explanation)
4. Tabulation of interop test results.  Not technically necessary to include, but aligns with plugfest results.


mmccool:
Note that I intentionally extract the "tabulated" assertions from the index.html rather than from the RDF.  I do this for two reasons: 1. it's easier 2. it ensures the implementation report reflects what is actually in the spec.

mmccool:
Ready to merge, should be done as soon as possible.

mmccool:
As discussed with Taki, I am merging after updating to the latest master.  This includes several updates to catch up with recent changes by @vcharpenay on how templates are handled, since I needed to include unique IDs for table entries.  I also did a few cleanups on files that are checked in ignored, in package.json, and in the README.

PR #324, opened 2018-12-12
remove writable key from property interaction type and remove readOnly and writeOnly from the required list

egekorkan:


sebastiankb:
thanks

PR #330, opened 2018-12-21
Update MultiLanguage

yamagile:
* jp -> ja based on [BCP47](https://tools.ietf.org/html/bcp47)
* I'm not sure which files should be updated, please comment if there's something.


sebastiankb:
looks good, thanks

PR #342, opened 2019-01-15
fix: there is no contentType "application/text" --> use "text/plain" ...

danielpeintner:


sebastiankb:
Unfortunately, I saw this PR too late. I already fixed this bug in one of my last commits. Thanks again for mention this

PR #352, opened 2019-01-18
Fix Json Typo

egekorkan:
there was a comma missing in example 14


mmccool:
That is indeed a typo, but you need to change it in index.template.html and execute "npm run render" to update index.html.  Otherwise your edit will just get overwritten.

egekorkan:
I see, I am closing this PR and making a new one

PR #356, opened 2019-01-19
fix examples and update validation schema

egekorkan:


egekorkan:
I think it works now. I have also updated the schema and added my proposal to change an assertion id

egekorkan:
I have updated the schema and changed id

PR #357, opened 2019-01-20
Change DataSchema mention in td-security assertion to SecurityScheme

egekorkan:


mmccool:
Yeah, that's a bug alright, but the edit needs to be made to "index.template.html" not "report.html", then "npm run render" and "npm run assertions" need to be run to update index.html and report.html.   I'll make a separate PR to fix this.

PR #358, opened 2019-01-21
change security assertions in td-vocab to MUST

egekorkan:


mmccool:
I have updated all the security vocab with default values to "mandatory".   This should update the corresponding assertions in the implementation report table.   Some of the other elements (those without default values) in some of the security schemes are actually optional, however.
Ege, can you please review the latest master and check if the updates have addressed this issue?   Note as well though that some of the default values for certain parameters of security schemes, in particular for "in", are dependent on the scheme in which they are used.   The tables do not correctly capture this contextual information (yet...).

mmccool:
PS I don't recommend that this PR be merged.  Since the file being edited is autogenerated, the changes would just get overridden.  We have to address the issue elsewhere in the spec.

mmccool:
Looking more carefully at the changes:
- My recent edits have made securityDefinitions mandatory, so that is fixed.
- The "security" field is mandatory at the Thing level, but not at the other levels (properties, actions, events, or forms).   The reason it shows up as MAY in the table is that my tool extracting the information does not understand the context in which a particular term is declared in the input.   It is actually declared in multiple places, the first being mandatory and the rest being optional... but the last definition seen overrides the others.   This is a bug in my assertion-extraction tool that I will have to fix.

mmccool:
So... I have been thinking about how to solve this problem and here is my plan. 

The fundamental issue is that vocab terms can appear more than once in different contexts, with “security” being an example.  I think these should be turned into multiple rows in the results table, each with a different context and id.   When an assertion appears with multiple variants like this, I will add a numerical suffix to the id.  In particular the “mandatory” variant of td-security (eg at the Thing level) will become td-security-1 and the “optional” variant will become td-security-2.   Also td-vocab-in will become td-vocab-in-1 and td-vocab-in-2 for the two variants with different default values.   There may be others...

I will update my table generator ASAP to automatically detect variants and create these entries.  We may have to do some tweaks to the playground to line up the ids used in the CSV files, but the above should at least let you do the right thing for “security” and “in”.   I will NOT change ids for assertions without variants.

Unfortunately it is not easy to fix the fact that there will be duplicate ids in the spec itself for these vocab terms, as the “render” script is a lot harder for me to modify, so hyperlinks from the table back to the spec may go to the wrong place, and of course ids should not be duplicated in HTML either.  I’m going to ignore that (for now...) as I think it’s a second-order problem.

mmccool:
I think “name” may also appear in multiple contexts...

egekorkan:
I agree with this, we can simply have numbers to indicate different instances and then one would look at the context of the assertion. Another possibility would be to add the context in the id, like td-name_basic

mmccool:
OK, I updated my code to capture the variants.  See https://github.com/w3c/wot-thing-description/pull/361.   There were a lot more than I expected...

mmccool:
BTW some of the id duplicates are simply mistakes, eg a cut-and-paste error causing an accidental reuse of an id.

mmccool:
OK, I fixed all the ID duplicates that seemed to have been accidental, namely all the ones that were not vocabulary items.   The remaining duplicates are all due to the context issue.   See PR https://github.com/w3c/wot-thing-description/pull/362.   I went ahead and merged this too...

There is one that is kind of weird: minimum/maximum appear as duplicates for the different types they apply to (integer and number/double).

PR #360, opened 2019-01-22
Behavioral section and assertions

mmccool:
This PR adds a new section for behavioral assertions, moves the existing security behavioral assertions to it, add adds several new assertions relating to compliance with data schemas.   Note: we should collect compliance data on these assertions at the test fest, even though this PR comes in after the "finalization" date for the TD, since it will be a lot harder to collect information on these assertions later.


egekorkan:
Is there any assertion that has been removed with this PR?

mmccool:
No assertions have been removed or reworded.   I did add several new ones around behavior... none of which can be tested with the playground, though.   It is notable that several of the new assertions related to how data sent to and returned from a Thing have to follow the data schema in the TD, though, which is something we *could* build a tool to check.  I'm thinking something like a "validating proxy" that can look at the data going past on a connection (at least, when it's not encrypted...).

mmccool:
PPS I keep forgetting to type “behavior” instead of “behaviour”.  In the spec itself I folllow the American spelling...

PR #413, opened 2019-02-07
Remove scopes and security at interaction level

mmccool:
As discussed in the Princeton F2F, "scopes" and "security" at the interaction level were not being used, and would have been marked at-risk, so we decided to remove this capability.   Security configurations can still be used at the Thing and Form levels, with the latter overriding the former, so fine-grained configuration is still possible.

The draft implementation report was also updated, but note there are now some parents with only one child (eg. td-arrays only has one child, td-arrays_forms, which has the exact same definition).  To complete this change we will also have to update the AssertionTester to remove these unnecessary child assertions.

This changes the definitions under validation as well, including the JSON schema, and also changes the text in index.template.html to describe the new overriding rules in terms of only two levels.


mmccool:
Note this merge was done AFTER the highlighting tooling so the report and the spec now have the new highlighting as well.

PR #414, opened 2019-02-07
Tooling to highlight at-risk assertions

mmccool:
Updated render.js and assertions.js to automatically highlight (with a bright yellow background, although this can be easily changed) at-risk assertions.   Assertions are highlighted both when they are given explicitly as spans or when they appear in tables.

The way this works is when "npm run assertions" is run it reads testing/inputs/atrisk.csv (which lists all the at-risk assertions) and generates testing/atrisk.css.   When "npm run render" is run it reads in "testing/atrisk.css" and sticks it into the generated HTML inside a "style" block.  Note to generate the implementation report you do things in the opposite order, so in theory you need to use a render/assertions/render sequence to get the correct final output, which should not be a serious problem since the render/assertions sequence is typically run over and over.

I didn't add any text anywhere to explain the highlighting, but there should be something in the intro or something.  If people don't like the CSS style I used this can be easily changed in assertions.js but the yellow is highly visible without impacting readability, so...


mmccool:
Bundled in a few extra things to this PR: cleaned up formatting of report (for instance, making colors more reasonable, including making at-risk assertions the same color as the highlighting in the spec) and updated categories and contexts so they are at least semi-correct (but these are specified manually, and so may contain errors and may get out of sync with the spec if there are changes to it).

mmccool:
Few more things: reran report using results from new directory; for "child" rules, regenerated parent results using min-pass, max-fail (and max-notimpl) over ALL implementations, not just one at a time (so we don't need every implementation to implement every child...).

Issue #434, opened 2019-02-14
Assertions in section 7.1 redundant?


vcharpenay:
[Section 7.1](https://w3c.github.io/wot-thing-description/#note-json-processing) contains assertions that seem redundant with the beginning of [section 6.1](https://w3c.github.io/wot-thing-description/#json-serializiation-section). Is there a consensus on that?

If so, it could be refactored in the form of an "algorithm" to add default values, with the same structure as section 7.2.

sebastiankb:
You are right. However, we should wait since the defaults values will be moved back to Section 5. 

Also see https://github.com/w3c/wot-thing-description/issues/337

vcharpenay:
The issue was addressed in PR #492.

PR #435, opened 2019-02-15
WIP: Modified assertions on @context

vcharpenay:
Assertions were put in `td-validation.ttl`.

This PR is also related to #434, which should be resolved before merging. Given that default values are provided for `@context`, half of the RDF transformation algorithm is superfluous.


mmccool:
As I mentioned in the Feb 15th meeting, I think putting all the defaults in one place, and including `@context` and `@type` in the information model, is cleaner and less confusing.  A system only has to act AS IF it filled in the defaults.   If it does not use a value like `@context` then equivalent behavior is obtained by simply ignoring it, so I don't see this being a problem in practice.

I'm not convinced we should nest assertions in table entries, however.   It creates nested assertions in the Implementation Report, since each row of a table is treated as an assertion defining a vocabulary term.   I think we should move those assertions out of the table and put them in the text somewhere (specifically, keep the first paragraph in the definition of @context in the table, move the second paragraph out to the text, and while you are at it, make the phrase about "optional vocabularies" its own sentence and reword it as a MAY assertion.

vcharpenay:
> pecifically, keep the first paragraph in the definition of @context in the table, move the second paragraph out to the text

this is the case with the changes I made.

> while you are at it, make the phrase about "optional vocabularies" its own sentence and reword it as a MAY assertion.

Just did that (see 5d927094779474ecd0f035e88b3e34d24795e7bd). Is the PR ready for merging now?

(Apologies: I messed up with my branches, this PR contains unrelated changes, which I unsuccessfully tried to revert. There shouldn't be any conflict if this PR is merged with #445)

takuki:
I think it reflects what we discussed in TD telecon last week, and looks great.

takuki:
The type of @context is now just an array without qualification on what is inside the array. It used to be "an array of JSON-LD contexts". I think we should qualify like that, instead of just saying it is an array.

vcharpenay:
good point, @takuki. I removed it because there was no definition of what a JSON-LD context is in the document. What about reformulating `@context` assertions (and keeping simply "array" in the table)? We could include details on the content of that array. Something like:

_The mandatory string `"http://www.w3.org/ns/td"` MUST be the first item in the `@context` array. It MAY be followed by one or more context definitions as strings or objects. If a context definition is a string, it SHOULD be a dereferenceable URL, as per [the JSON-LD specification](https://www.w3.org/TR/json-ld/#the-context). If a context definition is an object, it SHOULD be a map of namespaces, in which keys are namespace prefixes. Each optional namespace specified in the `@context` array of a Thing MUST be bound to an unique prefix._

takuki:
Thank you @vcharpenay . I added the suggested explanation in PR #452.

Issue #292, opened 2018-11-22
Identifiers for Properties, Actions, Events


maximelefrancois86:
Spec says: 

> Each instance of a Property, Action, and Event class MUST have an identifier that is unique within the context of a Thing Description document. See Section Representation Format for more details about JSON-LD 1.1 identifiers.

https://github.com/w3c/wot-thing-description/blob/807ca425dbce8868dbfd7f7d675d9ae99684512a/index.html#L487

- It is not said that the identifiers of  Property, Action, Event, MUST be JSON-LD 1.1 identifiers. What is meant by JSON-LD 1.1 identifier? How are these identifiers computed/defined.

- defined ->  it could be made more clear that the identifier is to be provided with `@id` key? maybe add this key explicitly in the Information Model?
- computed -> (in the RDF generation process, I heard @vcharpenay would like to compute unique identifiers concatenating the URI of the Thing with the JSON Pointer to the Prop/Act/Event) . How compatible is this with explicitly defined identifiers?

mkovatsc:
I agree we need to more explicit explanation and definition for this. Aspects from the top of my head (maybe incomplete or incorrect):

* Properties, Actions, and Events MUST have a Thing-local identifier, whose value also must be restricted to a valid JS object attribute name (proper term and reference needed here), e.g., must not contain slashes, must start with alpha-num or underscore, etc.
* This Thing-local identifier is mapped to the `@id` of the respective node (which is done automatically through `@container: @id` and the expected future `@base: @id`.
* In `@id` form, it is a IRI.
* There must be an algorithm that applies the `@container: @id` and `@base: @id` processing, so a JSON-LD 1.0 processor can work with it.

draggett:
Does the TD spec allow the same name to be used for properties, actions and events, or is the requirement only that you can't have two properties with the same name, and likewise for actions and events? 

What other constraints do we put on names for properties, actions and events?  I would suggest that we preclude names that start with "@" as this would otherwise make TDs harder to read.

I would assume that each property has exactly one name, but may have multiple descriptions when the developer wants to document a thing in more than one human language. I think we are waiting on guidance for the mechanism for doing that.

Note the terminological question about names and identifiers.  I've used name in the sense of an identifier above, but we could allow developer's to provide localised identifiers as metadata once we have guidance as how to do so.

vcharpenay:
[...]

PR #436, opened 2019-02-16
Remove td-unique-identifiers assertion

mmccool:
Since @vcharpenay updated the RDF conversion to generate unique IDs for `@id` based on JSON pointers this (annoying) assertion is no longer necessary.  This PR removes it, and should resolve Issue https://github.com/w3c/wot-thing-description/issues/292.

I also modified several assertions that had '(unique)' in them to just say 'unique', that is, I removed the parenthesis... which I felt gave the assertion a vague meaning.  I still find these assertions a little ambiguous since the scope of the uniqueness is not stated.  With the update by @vcharpenay the scope should just be the map in which the keys are used...  However when we add a "map of ..." to all the objects which are really maps, when we define "map" we should just state that the keys need to be unique within each such map, and that will take care of any remaining ambiguity.

I also fixed a minor capitalization inconsistency.


mmccool:
Note I added PR https://github.com/w3c/wot-thing-description/pull/437 to resolve the remaining ambiguity noted.

sebastiankb:
there are conflicts

mmccool:
Holding off on merging until I understand what we are doing about aligning with JSON 1.1... although I do think we don't *really* want the old behavior of requiring global uniqueness in the JSON serialization.

mmccool:
Not going to merge as it seems there are some spurious changes against the master now.   Going to redo this PR though since we DO want to remove that assertion.

Issue #345, opened 2019-01-17
Map of type


danielpeintner:
The information model definitions (according to my understanding) are not 100% clear about which types are actually used.

Please take a look for example at [Thing](https://w3c.github.io/wot-thing-description/#thing) definition. It states that `securityDefinitions` is of type `SecurityScheme`. I think this is not correct, it should read `map of SecurityScheme` similar to the `array of` types.

e.g.,
```
"securityDefinitions": {
      "basic_sc": {"scheme":"basic", "in":"header"}
  }
```

`SecurityScheme` only defines the *inner* part: `{"scheme":"basic", "in":"header"}` and not that it needs to be wrapped in a map with a key.

A similar case applies for `properties`, `actions`, and `events`.

ToruKawaguchi:
"version" also, right?

ToruKawaguchi:
Additionally, there are explanations at "securityDefinitions and security" and "version" which are unclear what "JSON name" points. For example at <https://w3c.github.io/wot-thing-description/#security-serialization-json> : 
> Each mandatory and optional vocabulary term as defined in the class SecurityScheme, MUST be serialized as a JSON name.

There are similar descriptions at "property", "action" and "event" which are relatively clearer. For example, at "property" <https://w3c.github.io/wot-thing-description/#property-serialization-json>: 
> Properties (and sub-properties) offered by a Thing MUST be collected in the JSON-object based properties member with (unique) Property names as JSON names.

danielpeintner:
> "version" also, right?

Mhh, I am not sure about `version`.
In this case it seems OK to me but let's clarify this also...

ToruKawaguchi:
@danielpeintner thanks. Understood that version is not the case for "map".

egekorkan:
I was handling this in the validation schema but just with human understanding skills, it should be clearer like mentioned :)

egekorkan:
Following the discussion in Princeton Testfest:
"Map Of" seems to be accepted even though it is not an XML Schema. We have to handle it in a custom way, like a table that contains these cases. 
Also we would have to explain what "Map Of" means. 
Arrays would not be possible since we use real arrays for other vocab terms

mmccool:
[...]

PR #437, opened 2019-02-16
Mark and Define Maps

mmccool:
This PR attempts to resolve Issue https://github.com/w3c/wot-thing-description/issues/345 by adding a rendering rule to templates.txt for specific vocabulary terms, namely: properties (both the InteractionPattern/InteractionAssertion one and the DataSchema one), actions, events, securityDefinitions, and uriVariables.   Note that definitions and titles are NOT marked as "map of" since MultiLanguage is itself a container with specific rules on the keys used.

The templates.txt lists the specific values this applies to.  There may be a more elegant way to handle this, eg by defining a container class somewhere.   If so, feel free to suggest an improvement to this rule, but this gets the job done.

There is also a definition of "map of" at the top of the Core Vocabulary Definition section.   I also put in a definition of "array of" here.   [EDITED] These definitions are intentionally abstract since other constraints or conventions (for example, that keys should be JSON names, or that arrays or maps should not be empty and should simply be omitted if there are no elements, or that arrays of one element can be replaced with that element) should be listed in the Serialization section.


takuki:
The fix looks comprehensive. Also, the paragraph defining "Map of X" looks very nice!

vcharpenay:
See PR #445 (already merged) that made the appropriate changes in the table. I rely on the context definitions in `/context` instead of hard-coding values in `templates.txt`. It is maybe more robust to changes.

I haven't textually defined "map of", thanks for that. Can you revert your changes in `templates.txt` and merge the rest?

mmccool:
OK, I will revert my changes, since @vcharpenay seems to have fixed the rendering of maps in a better way, I will update this PR to include only the paragraph defining "map of" and "array of", and merge.

mmccool:
No longer needed, @vcharpenay addressed already, so am going to close this PR without merging.   However I did add the paragraph defining what "map of" and "array of" mean in a separate PR.

Issue #448, opened 2019-02-21
Example1, Example2 and Example35 contain validation errors


mlagally:
When validating these two examples with the playground, I noticed some validation errors.

These relate to a missing "id" field at the top level and an extraneous "@id" field in the
security definitions and a missing array.

sebastiankb:
I just tested Example1 and Example2 with the TD playground and both tell me to be correct.

Example35 indeed have an error.

mlagally:
Are we using the same playground?

sebastiankb:
example 1:
![image](https://user-images.githubusercontent.com/13832739/53172981-ff17a280-35e6-11e9-84c9-82ec2cedb548.png)

example 2:

![image](https://user-images.githubusercontent.com/13832739/53173019-122a7280-35e7-11e9-9b17-7a2bbf3a02fc.png)

sebastiankb:
Please use only the TD simple or full version as defined in the TD spec. The JSON-LD form is non-normative yet.

mlagally:
You are right, I validated the JSON-LD versions.
My PR fixes the problems and they can be validated.

(On example 2 there's still the problem with the @context contents which I described In #447,
but this needs to be addressed in the schema)

sebastiankb:
https://github.com/w3c/wot-thing-description/issues/447 is solved and this issue can be closed

Issue #444, opened 2019-02-20
7.2 Transformation to JSON-LD & RDF seems unnecessary


msporny:
The entirety of Section 7.2 [Transformation to JSON-LD & RDF](https://w3c.github.io/wot-thing-description/#note-jsonld10-processing) feels unnecessary for at least one of these reasons:

* Developers often go from JSON/JSON-LD to RDF, rarely in reverse. If they do go in reverse, they pick and choose the RDF and output some simple JSON/JSON-LD. Having an algorithm to go from RDF to JSON has never been needed in the other specs/deployments using JSON-LD AFAIK. So, the fact that you're in unchartered waters here is a sign that something is off.
* If you do the JSON-LD Mapping correctly, you often never have round-tripping problems, so if you need such a complex algorithm, it's probably a sign that your data model is funkier than it needs to be.
* You may want to take a look at [JSON-LD Framing](https://w3c.github.io/json-ld-framing/) as it was designed to enforce layouts in ways that JSON developers like and some systems require when processing JSON-LD documents using simple JSON processors and hard coded Javascript object paths.

I suggest an immediate review by the JSON-LD WG for the spec as there are a number of areas in the specification that could benefit from their input, AFAICT.

vcharpenay:
Thank you, @msporny, for your review and apologies for not attending [today's call](https://www.w3.org/2019/02/20-wot-minutes.html). Looking back at the section, it deserves further case. Yet, I don't think we should completely remove it from the spec.

>  Developers often go from JSON/JSON-LD to RDF, rarely in reverse.

The algorithm presented in Section 7.2 is a transformation from JSON to RDF, not the contrary. The main use case is the integration of TD documents with contextual info in a large RDF store. I can remove references to the "bidirectionality" of the transformation. We haven't tested it in fact.

>  If you do the JSON-LD Mapping correctly, you often never have round-tripping problems

Is there a way to keep `@index` keys when transforming to RDF? This is our main problem, in fact. We suggested [to index by IRI](https://github.com/json-ld/json-ld.org/issues/430) instead, to keep index values in the RDF graph, but it comes with other issues like custom `@base` values.

sebastiankb:
In tomorrow's TD web conference we should discuss the new option that @msporny was recommended to us:

1) going back and claim the usage of JSON-LD 1.1 style again as we had for the Lyon f2f meeting (https://www.w3.org/TR/wot-thing-description)
[...]

PR #449, opened 2019-02-21
Fix syntax errors in several examples

mlagally:
This PR resolves:
https://github.com/w3c/wot-thing-description/issues/448


sebastiankb:
thanks. You just fixed example 35 and not 1 and 2, right?

vcharpenay:
I'm not sure I understand why you attempted to "correct" the JSON-LD version of Ex. 1. Did the simple/full examples change? If not, your changes make the example not consistent with the algorithm of Sec. 7.2, which is not exactly a "correction". But regardless of that, there is a discussion about Sec. 7.2, which might lead to rewriting the section and all JSON-LD examples (#444). The present PR should then remain open until a resolution is taken regarding that issue.

Otherwise, in Ex. 35, `components` is never specified and not included in the JSON-LD context. It is therefore not exactly a "JSON-LD 1.1 serialization", contrary to what the title says. What was your intention?

mlagally:
@vcharpenay The examples are simply wrong, since they don't validate, i.e. don't pass the playground.
It is not the intention of this PR to solve the JSON-LD serialisation.
The @components tag is needed to make the template conformant to JSON.
We need a way to structure the thing templates to enable defining a thing that consists of several components.

Issue #395, opened 2019-01-31
6.1 talks only about unsigned int


egekorkan:
Following Princeton Testfest:
6.1 Basic Representation Format Assumptions, the bullet points before the table talk about unsigned integers but not of signed integers. This means that if one implements a signed integer, we cannot serialize it.

takuki:
Now the spec says:
Vocabulary terms that use simple type integer MUST be serialized as JSON number.

egekorkan:
This seems to be fixed, closing

PR #455, opened 2019-02-22
Clarified JSON number representation used for representing integers.

takuki:
As discussed during 2/15 TD telecon.


takuki:
This is about Issue #395.

PR #474, opened 2019-03-04
Add context to vocab ids

mmccool:
Several errors in the test report were discovered that stemmed from duplicate ids in the specification.  These were the result of the use of the same vocabulary terms in different contexts, such as "security" (shows up in both Thing and Forms), "in" (shows up in multiple SecuritySchemes) and the like.  The previous attempt to solve this by scanning the document and adding numbers in post-processing was not working out; among other things, the numbering was not stable.   Secondly, duplicate IDs in the spec HTML is not conformant with the HTML standard; which broke other things, such as hyperlinks to specific assertions and CSS formatting to mark assertions as being as at-risk.

The best solution to this is to give all assertions unique IDs.  Therefore I updated templates.txt to add the context to the ID, using the name of the class preceeded by a "--" in the id.   For example, "security" is now labelled with two separate ids in the two places it appears: "td-vocab-security--Thing" and "td-vocab-security--Form".   Note that these new ids-with-context apply to ALL vocabulary terms, as there is no convenient way to special-case only the terms that appear in more than one place in a single pass.

This solves the problem BUT we need to update the AssertionTester to use the new IDs for ALL vocabulary terms.   I will do that shortly and create a PR against thingweb-playground.   When THAT is done and merged, then I can update the tests in wot/testfest/2019-03-online and regenerate the implementation report.   In the meantime while I am fixing all this be aware that the vocab section of the implementation report will not reflect the correct results.


mmccool:
I am currently working on updates the to AssertionTester.   This is still a WIP; it turns out I have to refactor a lot of the schemas to test each vocabulary term in one specific context to be compatible with the specification.

egekorkan:
Thank you @mmccool, this makes a lot of things easier

mmccool:
@egekorkan please take a look at the WIP PR to update the AssertionTester and let me know if I'm going in the right direction... I plan to plow through all the rest of the vocab tests tomorrow and finish this off before the Wed meeting.

mmccool:
Tests have mostly been updated and atrisk.csv has been updated to reflect the new test status.  Should be merged to act as a basis for discussion in plugfest meeting.  NOTE: missing tests are identified as "TEST" in atrisk.csv.   There are about a dozen of these but in most cases the missing tests are pretty simple to fix; for example, there is no test for the use of "properties", "events", or "action" objects in a Thing...

Issue #479, opened 2019-03-06
Allow multi-language descriptions in SecuritySchemes


mmccool:
For consistency, SecuritySchemes should also allow multi-language "descriptions".   Currently they only have "description".

mmccool:
See https://github.com/w3c/wot-thing-description/pull/480

PR #480, opened 2019-03-06
Allow MultiLanguage "descriptions" in SecuritySchemes

mmccool:
For consistency with other places where a "description" is allowed, also allow a multilanguage "descriptions" object.    Note the ontology was also refactored so wot-security refers to the MultiLanguage definition in the TD ontology, which avoids redundancy but which may or may not undesirable for other reasons.   However, it renders properly, so...  I guess technically "internationalization" definitions should be in their own standard ontology defined outside of WoT; something to bring up with the internationalization WG.


mmccool:
This addresses Issue https://github.com/w3c/wot-thing-description/issues/479

PR #489, opened 2019-03-08
Add At-risk Highlighting to "Status of This Document" section

mmccool:
I added a description of the yellow highlighting for at-risk items to SOTD section.  This PR also includes the updated test results.

However, I was unable to update the process document reference (the current only is 1 March 2019) because that reference is automatically inserted by ReSpec...  the same is true for the Architecture document, BTW.


mmccool:
The process document reference should be https://www.w3.org/2019/Process-20190301/, but that text is generated by ReSpec, so...

k-toumura:
@mmccool, From my browser, process document reference is rendered correctly, as shown below.

screen shot 2019-03-08 at 18 50 54

Unfortunately, your commit https://github.com/w3c/wot-thing-description/commit/ca33ce76c1dcd2b0c53d25bdbc61b8500f512c37 reverts a my PR #476.   I'll make the PR again.

PR #506, opened 2019-03-16
Explainer Updates

mmccool:
Updates to explainer document.   Ready to merge; please review.
1. New narrow waist/hourglass figures and explanation of how the WoT narrow waist is not about a "common protocol", but about a "common set of abstractions"
2. Moved "JSON-LD 1.0" design alternative here from Arch explainer (a separate PR deletes it from the Arch explainer)
3. Filled in implementation section; also listed companies that did implementation, and the number of implementations (15) and instances (65).
4. Filled in at-risk section using lastest test results.   Note that there are two categories: those that will probably still be at risk for the CR transition, and those I think we can fix prior to it.
5. Removed "Other information in TD" section


mmccool:
Ready to merge.  This version feels "complete" to me and would be suitable (IMO) for sending to TAG.   However, members should review ASAP.

PR #515, opened 2019-03-22
Update validation schema

egekorkan:
I have changed the description field to mention that it validates the plain JSON TD


sebastiankb:
thanks

Issue #473, opened 2019-03-01
Should the Full Validation Schema be referenced


egekorkan:
Given that we have some mandatory items that are not mandatory outside of default value generation, I had created two versions of the validation schema. The one in the specification is the simple version that does not oblige the keys that would be generated with "a default value generator". Should the full schema be also referenced (not the entire schema), maybe with default values?

* Full Schema: https://github.com/thingweb/thingweb-playground/blob/master/WebContent/td-schema-full.json
* Simple Schema: https://github.com/thingweb/thingweb-playground/blob/master/WebContent/td-schema.json

sebastiankb:
please note that the "full" TD is now JSON-LD 1.1 compliant. Lets decide this in tomorrows TD meeting

sebastiankb:
In today's TD meeting we decided to have also the "full" JSON Schema in the TD annex. It should be stated what is the different between both schemas and its purpose. 

@egekorkan can you make a PR?

egekorkan:
Done via #516

PR #516, opened 2019-03-22
Create jsonld TD validation

egekorkan:
This was referred to as full schema before.


egekorkan:
In response to #473

sebastiankb:
thanks

PR #523, opened 2019-03-24
Fixed issue in rendering nested context information

vcharpenay:
Among others, `ObjectSchema.properties` would not render properly in Section 5.


sebastiankb:
thanks

Issue #536, opened 2019-03-28
@context for security and data schema


egekorkan:
`@context` has been removed from the vocabulary definitions of Data Schema and Security Scheme but its assertions and thus the associated paragraphs are still present at the end of section 5.3.1 and 5.4.1.

sebastiankb:
good point! This has to removed.

mmccool:
I was wondering... I was trying to update the implementation report just now (to match the at-risk highlighting in the TAG version, among other things) and ran into this just now.   How about I include removing these as part of my PR to update the atrisk highlighting?

mmccool:
OK, PR https://github.com/w3c/wot-thing-description/pull/537 removes those assertions.
Note I left the "sh:description" for SecurityScheme there, but blank, in case we need to add back some other assertions.

egekorkan:
This has been fixed

PR #537, opened 2019-03-29
Update at-risk assertions

mmccool:
Update at-risk assertions, removing those which we expect to pass (we just need tests) which were manually removed by editing atrisk.css in the TAG-review version of the report.  However, this PR updates these in a way (modifying the testing/inputs/atrisk.csv file rather than the testing/atrisk.css file) that also leaves the highlighting out in the Implementation Report and also does not break when the Implementation Report is regenerated (since that process writes to testing/atrisk.css).

As noted in Issue https://github.com/w3c/wot-thing-description/issues/536, there were also some assertions left in the text regarding the (now removed...) `@context` fields for SecurityScheme and DataSchema.  I have removed them (they were in validation/td-validation.ttl, by the way, which seems like an odd place to put *content*, but oh, well).

This PR also includes an updated rendering of the index.html doc and the implementation report with the above assertions removed and the highlighting updated.  I did a diff with the TAG review version and the highlighting of this version is now consistent; the only changes were a new conformance paragraph in the TAG-review version (which maybe should be in the master... another issue) and the sections I removed.


mmccool:
HTML diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.staticaly.com%2Fgh%2Fw3c%2Fwot-thing-description%2FTD-TAG-review%2Findex.html%3Fenv%3Ddev&doc2=https%3A%2F%2Fcdn.staticaly.com%2Fgh%2Fmmccool%2Fwot-thing-description%2Fmaster%2Findex.html%3Fenv%3Ddev

Issue #501, opened 2019-03-14
Clarify how new security schemes can be given in extensions


mmccool:
Currently it looks highly likely at least some of the security schemes will be at-risk and in danger of being dropped.   This includes potentially useful schemes like PSK and OAuth2.   This is not a big deal IF we can define these later as needed with extensions.   However, we need to clarify exactly how this will work, and maybe provide an example in the text.

One issue is that the definitions of `scheme` and `@type` overlap.   Right now `scheme` is mandatory and must be one of a specific set of values.  But if we use an extension to define a new scheme, we would include an `@context` to pull in the new vocabulary and use an `@type` to identify the class.   Say for example we had to get rid of OAuth2, and support it in an extension.   Then a security scheme using that extension might look like

```
{
     "@context": [ "http://...."],
     "@type": ["OAuth2SecurityScheme"],
     "scheme": "?",
     ...
}
```

What should `scheme` be in this case?  If it is `"oauth2"` where does that value come from?  How can we ensure that `scheme` and `@type` are consistent?  Isn't this redundant?   What if a user makes a mistake and uses something else for `scheme`, like (for example) `"digest"`?  Is that an error, or does one of the two values take precedence?  

I see a couple of solutions to this:

1. Make `scheme` optional, but allow it to be omitted (and perhaps require it to be omitted to avoid redundancy and possible conflicting information) if there is an `@type`.   
2. Get rid of `scheme` and make `@type` mandatory inside a `SecurityScheme`; its functionality therefore replaces `scheme`.

Option 2 would be the most consistent but would require some pretty major changes to all our current TDs, and goes against the convention we have been using that values beginning with `@` can be omitted if you are not doing semantics.  So I prefer option 1.  We should also make sure that we can drop the string values referring to at-risk items if we drop those at-risk schemes.  I also prefer *requiring* `scheme` to be omitted if there is an `@type`; then we avoid the whole issue of having to define new string values for `scheme` and also avoid the possibility of conflicting definitions.   In this case `scheme` is just a short-hand way of defining the `@type` so the normalization (during conversion to JSON-LD) would be to delete `scheme` if it is given and generate the corresponding `@type` value.

mmccool:
The tricky bit here is making `scheme` semi-optional... what we really want is to say there is either an `@type` OR a `scheme` but never both (eg a oneOf relationship), and never neither.

mmccool:
Yes, I know this is unnecessarily complicated.   If someone has a better idea...

[...]

Issue #470, opened 2019-03-01
scopes explanation in form is not linked to oauth2 in securityDefinitions


egekorkan:
The explanation of scopes in Form is currently the following:

> Set of authorization scope identifiers, provided as an array. These are provided in tokens returned by an authorization server and associated with forms in order to identify what resources a client may access and how.

Which doesn't link where the scope values are defined. For example, for security we have currently the following text:

> Set of security definition names, chosen from those defined in securityDefinitions. These must all be satisfied for access to resources at or below the current level, if not overridden at a lower level.

Here we say that a string given in security array should be defined in securityDefinitions. Since scopes is used only with oauth2, we should say something like the following: 

> The set of scopes names chosen from those defined in an "oauth2" security scheme defined in securityDefinitions.

mmccool:
Addressed in PR https://github.com/w3c/wot-thing-description/pull/538.

egekorkan:
Seems to be fixed

Issue #350, opened 2019-01-18
security in forms mentions lower levels


egekorkan:
The security key at form level has the following text:

> Set of security definition names, chosen from those defined in securityDefinitions. These must all be satisfied for access to resources at or below the current level, if not overridden at a lower level.

However, I think it should be:

> Set of security definition names, chosen from those defined in securityDefinitions. These must all be satisfied for access to resources at the current level.

This can be confusing due to two reasons:

1. In an interaction, there is nothing lower than the forms field.
2. In case there is forms field in the thing level to describe readallproperties etc. with a security scheme, one can understand that a single action in the "lower level" can require this security scheme.

mmccool:
Addressed in PR https://github.com/w3c/wot-thing-description/pull/538.   I actually made it even simpler: "These must all be satisfied for access to resources."   The overriding behavior is specified by another set of assertions, so defining it here was redundant anyway.   But we do need to state the "all must be satisfied"

mmccool:
PR merged, so this should be taken care of now.

egekorkan:
Seems to be fixed

PR #538, opened 2019-03-29
Security extension example and namespaces

mmccool:
NOTE: this is based on top of PR https://github.com/w3c/wot-thing-description/pull/537, so includes those edits as well.

This PR addresses Issue https://github.com/w3c/wot-thing-description/issues/501.  It provides an example of how to use an extension to implement a new security scheme, using OAuth2 as an example. 

Also addresses https://github.com/w3c/wot-thing-description/issues/345,  https://github.com/w3c/wot-thing-description/issues/350, and https://github.com/w3c/wot-thing-description/issues/470.

NOTE: OAuth2 is important but is also likely to be removed from the specification due to lack of implementation experience.   So if we do remove it, this example can remain and still be valid (although we might have to remove the immediately preceding example).


mmccool:
Originally I had an assertion defining a standard namespace for security schemes but decided to drop it and update the example to use an "example.org" URL.   We probably SHOULD define a standard namespace for security schemes but we can publish that separately and put it into a W3C NOTE (maybe the WoT Security Best Practices note).   So this PR really only includes an example.  In the end I decided to say nothing about the overlap of `@type` and `scheme`...

mmccool:
Extended to also take care of Issue https://github.com/w3c/wot-thing-description/issues/470.

takuki:
Does this mean that some level of JSON-LD processing will be needed to understand security schemes?

mmccool:
Also addressed https://github.com/w3c/wot-thing-description/issues/350.
Replying to Taki's comments: JSON-LD processing will not be required to understand built-in schemes.  It will be needed to understand extensions, including extensions defining security schemes.   Which is why it is still a good idea to have built-in schemes.  I suppose an implementation could take a middle ground though and understand "well-known" security schemes.   We could publish a bunch in a note...

mmccool:
Also fixed https://github.com/w3c/wot-thing-description/issues/345

mmccool:
An HTML diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.staticaly.com%2Fgh%2Fw3c%2Fwot-thing-description%2FTD-TAG-review%2Findex.html%3Fenv%3Ddev&doc2=https%3A%2F%2Fcdn.staticaly.com%2Fgh%2Fmmccool%2Fwot-thing-description%2Fsecurity-extensions%2Findex.html%3Fenv%3Ddev

mmccool:
OK to merge, but we will have to clean up the definition of extensions, perhaps in another document.  We will merge, @vcharpenay will define some issues, @mmccool will work on them and define a PR to discuss in the main call next week, and we can aim to merge any updates by the next TD call.

PR #571, opened 2019-04-15
Remove full schema validation related information

egekorkan:
I have reorganized the validation section:
Removed the full schema data from the spec and from github
Remove mentions of full schema or any difference


sebastiankb:
thanks

PR #581, opened 2019-04-25
Updated test results - 25 April 2019

mmccool:
Updated to latest version of AssertionTester, re-ran implementation report generator, updated list of atrisk items and categories to track new assertions.
Note: mostly this update only affects files in the testing subdirectory.  However, I had to modify index.template.html to fix some duplicate ids in the tabulated assertions for the new http-default-method assertion.   This should be "invisible" to the rendered version, however.


mmccool:
updated with reordered columns as discussed in April 25 TD call.  The Context now comes before the assertion to make contextual assertions clearer.

Issue #588, opened 2019-04-30
Proper IANA Considerations section


mkovatsc:
We need a proper top-level section for the IANA registration, following their template:

https://tools.ietf.org/html/bcp13#section-5.6

JSON-LD 1.1 has a good example:

https://w3c.github.io/json-ld-syntax/#iana-considerations

mkovatsc:
Once done, I also need to send the e-mail to media-types@iana.org following https://tools.ietf.org/html/bcp13#section-5.1

mkovatsc:
Done: https://mailarchive.ietf.org/arch/browse/media-types/

takuki:
@mkovatsc , we need to update IANA with the CR version URL.

mkovatsc:
So far, I only sent out the Preliminary Community Review, for which the Editor Draft still is technically up to date.

It is a pitty that the Link under IANA Considerations was not updated in the CR version... We have to make sure we reference the right document when sending the actual registration request to IANA.

See #560 for IANA progress.

Issue #560, opened 2019-04-05
Make a request for TD media type registration at IANA


sebastiankb:
We have to start the process to register `application/td+json` at IANA. This has to be announced in the TD specification.

sebastiankb:
@ashimura will talk with Ralph about this

ektrah:
FWIW, from the IETF side, this should be as easy as filling out this form: https://www.iana.org/form/media-types.

[RFC 6838](https://tools.ietf.org/html/rfc6838#section-3.1):

> Standards-tree registrations from recognized standards-related organizations are submitted directly to the IANA, where they will undergo Expert Review prior to approval.

mkovatsc:
The e-mail to media-types@iana.org for the Preliminary Community Review is out. Track on https://mailarchive.ietf.org/arch/browse/media-types/

mkovatsc:
@ektrah Is there something similar for CoAP Content-Fortmat IDs? I included that information in the mail to IANA. We want to go for the 2-byte range in 256-1000, since TDs are usually quite large anyway.

ektrah:
@mkovatsc I assume you mean the media-types@ietf.org mailing list for community review; or have you contacted IANA already directly? You could send the content format registration request also to the core-parameters@ietf.org mailing list for community review. But since there's not much to review in a content format registration (the meat is in the media type registration), you could simply include this when sending the request to IANA.

mkovatsc:
https://tools.ietf.org/html/bcp13#section-5.1 lists "@iana.org", but I guess it is the same mailing list.

I included the intent for Content-Format there, so I will do both in the IANA request.

mkovatsc:
Update:

Besides the mail from Mark on the [mailing list](https://mailarchive.ietf.org/arch/browse/media-types/), I received a comment about "clipboard flavor names". They help in operating systems to preserve the format when copying & pasting between applications. The commentator has been busy, so I am still waiting for some consensus on that.

mkovatsc:
I just checked [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml) and in many cases the reference points to the CR document.

Unfortunately, the reference in the IANA section of the TD CR was not updated.
@ashimura Can we fix this? Or ideally, we should give the updating https://www.w3.org/TR/wot-thing-description/ as reference for IANA, no?

mkovatsc:
Unclear what "Reference" IANA expects:

[...]

PR #594, opened 2019-05-02
Add IANA section

mkovatsc:
Added Section 10 IANA Considerations.

I followed the template provided by IANA and rendered it as in other current W3C specifications (JSON-LD 1.1, HTML). Note that the links still point to the Editor's Drafts and need to be fixed later.

The section includes both media type registration and CoAP Content-Format registration.

Once merged, I will send the e-mail to IANA (see #588).

### DIFF
--- a/index.html
+++ b/index.html
@@ -3045,22 +3045,18 @@ If included namespaces are based on class definitions such as those provided by 
 
  </section>
 
-  <section id="media-type-section">  
-  <h1>Media Type</h1>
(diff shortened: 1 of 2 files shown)


takuki:
Once CR is published (expected as soon as the week of May 20th), do we need to notify IANA of the URI for CR version of TD spec?

mkovatsc:
Please see #560 for IANA topic.

Issue #599, opened 2019-05-03
Example 2 title could be improved


jmcanterafonseca:
Current figure title:

EXAMPLE 2: Thing Description with context extension

Suggested: 

EXAMPLE 2: Thing Description with semantic annotations using the SAREF ontology (@context extension)

sebastiankb:
The examples should just show that there is this context extension feature in the TD. I think the title is enough for that. Further details are described in the paragraph above.

jmcanterafonseca:
could it be "Thing Description with context extension using SAREF" ?

mkovatsc:
@jmcanterafonseca , are there hard feeling for this? (e.g., for marketing SAREF :P )
Otherwise I would close.

jmcanterafonseca:
@mkovatsc intention was not to promote SAREF specifically :), but to promote semantic annotation as a way of showing the potential of TDs ... having said that you can close or add somehow the "semantic annotation" slogan to the example title :) 

thx!

mkovatsc:
Hmm, valid point -- we have been working too hard on hiding this for those who do not like RDF semantics..
Maybe "Thing Description with context extension for semantic annotations" would do something positive :)

mkovatsc:
See #646

jmcanterafonseca:
@mkovatsc thanks! LGTM in #646, once that lands I think we can close this.

PR #601, opened 2019-05-03
Fix #65

### LINKED ISSUE #65: Do we need an "observeData" field?
Currently we have "outputData" and "inputData", corresponding to payloads for read and write operations or inputs and outputs from Actions, or output data from Events.

However, if we add an "observe" capability to Properties, then there is no guarantee that a given protocol will use the same payload format as a read.   As an example, the Mozilla proposal wraps the payload for an observe message in an additional JSON object that includes a timestamp, so the structure (not just the fields) are different from that returned from a GET.   This example shows that we can't necessarily just ignore fields to deal with differences, as the structure itself might be different.

So, we need a separate way to give payloads for observe events.   One suggestion that generalizes inputData and outputData to multiple roles is given in issue https://github.com/w3c/wot-thing-description/issues/66.  Of course we could also just add an "observeData" tag.

jmcanterafonseca:
I don't know if I could only modify index.template.html ... What is the right procedure? 

What's the difference between index.html and index.template.html ?


vcharpenay:
You should first modify `index.template.html` and then run a rendering script that will automatically generate `index.html`. See the repo's [README](https://github.com/w3c/wot-thing-description/#rendering) for more details.

sebastiankb:
please see my comment https://github.com/w3c/wot-thing-description/issues/599

Issue #549, opened 2019-04-02
Use JSON-Schema Draft07 in JSON Schemas


jmcanterafonseca:
I think it would be good to use draft07 of JSON Schemas.

jmcanterafonseca:
I can contribute if agreed

egekorkan:
What keywords of Draft 7 do you want to use and what will be their advantage?

jmcanterafonseca:
I think taking the last version is safer in terms of tool support. For instance, format: iri it is a good addition, btw are TD ids to be URIs or IRIs?

egekorkan:
> I think taking the last version is safer in terms of tool support.

I agree with this.

> btw are TD ids to be URIs or IRIs?

I am not completely sure on this. After your comment I have checked this and we should have been using IRIs due to https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#anyURI

If we confirm this, I can do the change.

sebastiankb:
@handrews  would be great if you can comment on this?

vcharpenay:
> btw are TD ids to be URIs or IRIs?

`id` being an alias for `@id`, values can very well be IRIs, as [specified by the JSON-LD recommendation](https://www.w3.org/TR/json-ld/#iris). The ultimate answer depends on the media type associated to individual TD documents:

 - if the media type is `application/ld+json` (with a TD profile), the value may be an IRI and it should be understood by clients (but it is not defined whether the TD can be serialized as a pure TD, see below)
 - if it is `application/td+json`, then the value for `id` should be a URI, as per the information model.

As far as I can tell, the current plan is to allow both options.

egekorkan:
A tested version with draft 7 is found here: https://github.com/egekorkan/wot-thing-description/blob/master/validation/td-json-schema-validation.json

sebastiankb:
issue is solved by PR above. TD spec uses 07 draft

PR #603, opened 2019-05-03
Fix JSON Schema to match latest spec version

jmcanterafonseca:


mkovatsc:
@egekorkan These are valid changes. Unfortunately, we cannot merge due to the failed IPR check. Could you please apply these fixes, best already including the new context URI `https://www.w3.org/2019/td/v1`. There should also only be one choice for now, i.e., a single enum entry.

@jmcanterafonseca Thanks for spotting and fixing!

jmcanterafonseca:
@mkovatsc @ashimura @draggett is there a way to overcome the IPR issues? Just an statement saying we don't want to claim any IPR (We are a non for profit org), 

I would like to contribute or if I contribute, I would like to be listed as contributor ... :)

egekorkan:
There was a recent change by @vcharpenay that puts a pattern for the context value so is 
> There should also only be one choice for now, i.e., a single enum entry

still valid?

egekorkan:
@jmcanterafonseca you have an extra comma at the end of `$ref` after `contains`. 

I have taken your proposal and updated the schema to also support draft 7 like we talked in issue #549. I didn't do a PR yet since I find the issue #628 blocking. However, you can find the schema on my branch here: https://github.com/egekorkan/wot-thing-description/blob/master/validation/td-json-schema-validation.json

jmcanterafonseca:
@egekorkan cool, thanks!

ashimura:
Thanks a lot for your input @jmcanterafonseca !

Regarding how to resolve this pullrequest, theoretically we could mark your PR as "non-normative" on the [repository manager https://labs.w3.org/repo-manager/pr/id/w3c/wot-thing-description/603] if it's OK by the Editors/Chairs.

However, my tendency is only those who made an IPR commitment makes pullrequests for WG repos. You can still create an issue whenever you want.

sebastiankb:
@jmcanterafonseca please see https://github.com/w3c/wot-thing-description/pull/632
If ok we can close this PR

Issue #596, opened 2019-05-03
Remove Direct Reference to wot-security-best-practices


mmccool:
The WoT Security Best Practices document, as well as the WoT Security Testing Plan, have not yet been published as W3C Notes.   As such we cannot technically refer to them from a REC.   As we did with the WoT Architecture document, we should remove direct references to these from the TD spec but instead refer only to the WoT Security and Privacy Considerations document.   This document, which HAS been published as a Note (and so is a valid citation) can then indirectly refer to these documents.

See also https://github.com/w3c/wot-architecture/issues/284 in the WoT Architecture issue tracker.

mmccool:
See PR https://github.com/w3c/wot-thing-description/pull/619.

PR #619, opened 2019-05-05
Remove reference to wot-security-best-practices

mmccool:
Since the "WoT Security Best Practices" document has not been published yet as a Note, we cannot cite it.  Therefore this citation is removed.   Note it is still available indirectly through the citation to WoT Security and Privacy Considerations.

Resolves Issue https://github.com/w3c/wot-thing-description/issues/596.


mkovatsc:
Looks good.

@mmccool It would be better not to include the other generated files in the commit, when you only change the HTML.

sebastiankb:
thanks @mmccool

Issue #598, opened 2019-05-03
Add text direction definition


mmccool:
State that default text direction is to be inferred from the default language but that this places some restrictions on strings (they should not start with text, such as numbers, that are not in the same direction as the default language).

Please comment if you find a suitable citation or text in a similar specification that we can reuse.

mkovatsc:
I think we should pick this solution: https://www.w3.org/TR/string-meta/#script_subtag

* We are based on JSON-LD, which "prevent the use of metadata" for direction metadata
* Default is LTR
* Langauge tags in `MultiLanguage` names MAY indicate RTL
* We must state that the `MultiLanguage` names MUST provide the [script subtag](https://tools.ietf.org/html/bcp47#section-2.2.3) when it changes the direction

mkovatsc:
Also make a note that this puts constraints on strings such as a RTL string cannot start with a multi-digit number. Here we might want to ask for help to be correct.

mmccool:
So, after reading over the section noted in string-meta, I think the "don't start with a number" issue is not actually a problem if we state that the base direction is ALWAYS inferred from the script sub-tag.  There are cases however where numbers with embedded spaces are displayed in the wrong order but this seems to be a problem with all encodings.  However, I noticed an additional problem: we have to limit the strings to which this is applied to the human-readable entries, i.e. title, description, etc.  For example, we would not want to apply it to "scheme", even though it is a "string".  

One troublesome field though is "support".  This is a "string", but is "free-form": it could be a URL, or it could be human-readable text, or it could be a phone number (which could run into the embedded spaces issue...).   Solutions include:
- Saying it must always be a URI
- Removing it
- Saying it must always be human readable
- Saying the strong-first algorithm must be used on it instead of the default base direction

I personally am in favor of the first option.   There is also no "supports" which would be weird anyway.  But if we give a URI then what it points at can do the heavy lifting.

mmccool:
The Thing "name" is also troublesome.  I assume this is meant to be human-readable text, eg for display on web dashboards, and is not meant to be unique (we have the id for that).   So I'm assuming that we should apply the base direction to it.

However, there is no "names".   As a separate issue (which I will file), I think there should be, so that when language negotiation is used a name in an appropriate target language can be provided.

mmccool:
[...]

PR #620, opened 2019-05-05
Add assertions for language objects and text direction (issue 598)

mmccool:
- Be specific about where `@language` tag can appear to simplify processing (this is an extension of issue noted below...)
- Add assertions to indicate how base text direction is to be inferred from default language outside of MultiLanguage contexts, including requirement for script subtags if necessary (btw, an example of the latter is az-Latn vs az-Arab)
- Make LTR the default text direction if no default language is given
- Script subtag required in MultiLanguage.  I also moved the assertions regarding language tags from ontology/td.ttl to validation/td-validation.ttl to be consistent with other such assertions, and marked them up with rfc2119-assertion spans.

Addresses Issue https://github.com/w3c/wot-thing-description/issues/598


index.html
```diff
@@ -1401,9 +1401,47 @@ <h2>Core Vocabulary Definitions</h2>
 <p>
      <span class="rfc2119-assertion" id="td-context-default-language">
      In the <code>@context</code> array, the optional default language that is 
-     used throughout a TD document is defined within an object as a member with 
+     used throughout a TD document for human-readable strings outside of a <a href="#multilanguage">MultiLanguage</a> object
+     is also defined within an object as a member with 
      a key <code>@language</code> whose value MUST be a string representing a [[!BCP47]]
```
mkovatsc: Should be "the name `@language` whose ... language tag."

index.html
```diff
@@ -1401,9 +1401,47 @@ <h2>Core Vocabulary Definitions</h2>
 <p>
      <span class="rfc2119-assertion" id="td-context-default-language">
      In the <code>@context</code> array, the optional default language that is 
-     used throughout a TD document is defined within an object as a member with 
+     used throughout a TD document for human-readable strings outside of a <a href="#multilanguage">MultiLanguage</a> object
+     is also defined within an object as a member with 
      a key <code>@language</code> whose value MUST be a string representing a [[!BCP47]] 
      language code.</span>
+     <span class="rfc2119-assertion" id="td-context-default-language-last">
+     The object used to define the default language MUST be the last object in
```
mkovatsc: MUST be the last _element_ in

mkovatsc: However, this appears to be a serialization issue to me...

index.html
mkovatsc: Talking about "member" would make it a serialization assertion.

index.html
```diff
@@ -1401,9 +1401,47 @@ <h2>Core Vocabulary Definitions</h2>
...
+     the `@context` array.
+     </span>
+     <span class="rfc2119-assertion" id="td-context-default-language-unique">
+     There MUST be only one object defining a default language in an `@context` array.
+     </span>
+     <span class="rfc2119-assertion" id="td-context-default-language-separate">
+     The object used to define the default language MUST NOT be combined with any other
+     objects defining context extensions (additional vocabulary terms).
```
mkovatsc: This is an issue when a JSON-LD library is used to serialize a TD. I would rather relax this. (Also applies to the position in the array...)

index.html
```diff
@@ -1401,9 +1401,47 @@ <h2>Core Vocabulary Definitions</h2>
...
+     </span></li>
+     <li><span class="rfc2119-assertion" id="td-context-default-language-direction-inference">
+     In cases where a language can be written in more than one script which use 
+     different writing directions, the value used for `@language` MUST
+     include a script subtag so that an appropriate base direction can be inferred. 
+     </span></li>
+     <li><span class="rfc2119-assertion" id="td-context-default-language-direction-default">
+     If no default language is given, the base text direction MUST be assumed to be LTR (left-to-right).
+     </span>  This implies that if the language used in human readable text uses a script
+     that is written RTL (right-to-left), the default language needs to be explicitly specified.</li>
```
mkovatsc: the default language, _potentially including a script subtag_, needs to be ...

index.html
```diff
@@ -1401,9 +1401,47 @@ <h2>Core Vocabulary Definitions</h2>
...
+     <li><span class="rfc2119-assertion" id="td-context-default-language-direction-independence">
+     Each human-readable string MUST be processed independently, starting from the base
+     direction defined for the TD.</span></li>
```
mkovatsc: Where does this come from? Does it mean the direction has to be computed every time, although the default language cannot have changed?

index.html
```diff
@@ -1494,12 +1532,24 @@ <h2>Core Vocabulary Definitions</h2>
...
-<section><h3><code>MultiLanguage</code></h3><p>Container to provide human-readable text in different languages using language tags described in [[BCP47]]. Each used language tags code MUST be decleared as vocabulary term member of this container (e.g., en, de, ja, zh-Hans, zh-Hant). See <a href="#titles-descriptions-serialization-json"></a> for example usages of this container in a Thing Description instance.</p>
+<section><h3><code>MultiLanguage</code></h3><p>Container to provide human-readable text in different languages using language tags described in [[BCP47]]. See <a href="#titles-descriptions-serialization-json"></a> for example usages of this container in a Thing Description instance.</p>
 <p>
+<span class="rfc2119-assertion" id="td-context-ns-multilanguage-language-tag">
+Each language tag code used MUST be declared as vocabulary term member of this container
```
mkovatsc: Strike _member_ here

index.html
```diff
@@ -1494,12 +1532,24 @@ <h2>Core Vocabulary Definitions</h2>
...
+<span class="rfc2119-assertion" id="td-context-ns-multilanguage-language-tag">
+Each language tag code used MUST be declared as vocabulary term member of this container 
+(e.g., <code>en</code>, <code>de</code>, <code>ja</code>, <code>zh-Hans</code>, <code>zh-Hant</code>,
+<code>az-Latn</code>, <code>az-Arab</code>). </span>
+<span class="rfc2119-assertion" id="td-context-ns-multilanguage-text-direction-infer">
+The base text direction MUST be inferred from the language tag code.
+</span>
+<span class="rfc2119-assertion" id="td-context-ns-multilanguage-text-direction-script-subtag">
+If a language can be written in more than one script with different base directions 
+then a script subtag MUST be used so that the base direction can be inferred.
```
mkovatsc: ... MUST be _included_ so that ...
?

index.html
```diff
@@ -1494,12 +1532,24 @@ <h2>Core Vocabulary Definitions</h2>
...
-<section><h3><code>MultiLanguage</code></h3><p>Container to provide human-readable text in different languages using language tags described in [[BCP47]]. Each used language tags code MUST be decleared as vocabulary term member of this container (e.g., en, de, ja, zh-Hans, zh-Hant). See <a href="#titles-descriptions-serialization-json"></a> for example usages of this container in a Thing Description instance.</p>
+<section><h3><code>MultiLanguage</code></h3><p>Container to provide human-readable text in different languages using language tags described in [[BCP47]]. See <a href="#titles-descriptions-serialization-json"></a> for example usages of this container in a Thing Description instance.</p>
 <p>
+<span class="rfc2119-assertion" id="td-context-ns-multilanguage-language-tag">
+Each language tag code used MUST be declared as vocabulary term member of this container 
+(e.g., <code>en</code>, <code>de</code>, <code>ja</code>, <code>zh-Hans</code>, <code>zh-Hant</code>,
+<code>az-Latn</code>, <code>az-Arab</code>). </span>
+<span class="rfc2119-assertion" id="td-context-ns-multilanguage-text-direction-infer">
+The base text direction MUST be inferred from the language tag code.
```
mkovatsc: strike _code_

index.html
```diff
@@ -1494,12 +1532,24 @@ <h2>Core Vocabulary Definitions</h2>
...
+then a script subtag MUST be used so that the base direction can be inferred.
+</span>
 <span class="rfc2119-assertion" id="td-context-ns-multilanguage-content-negotiation">
-Note that in protocols such as HTTP that have an ability to negotiate content based on languages, 
-those terms (e.g. <code>description</code> and <code>title</code>) intended for carrying texts in a default language SHOULD be used instead of <code>MultiLanguage</code> container
-when the texts are served as a result such negotiation.
+In protocols such as HTTP that have an ability to negotiate content based on languages, 
+those terms (e.g. <code>description</code> and <code>title</code>) intended 
+for carrying texts in a default language SHOULD be used instead of <code>MultiLanguage</code> container
+when the texts are served as a result of such negotiation.
```
mkovatsc: Only works for dynamically serialized TDs...

index.html
mmccool: No, it means you can't carry forward changes in direction from one string to another, or take the order that strings in the document appear as part of the processing.   Each string is processed independently, starting from the same base direction.  Which is what I said.

index.html
mmccool: Sure, means basically the same thing, but I can make the change.

index.html
mmccool: Actually, I don't think I changed this, it just got moved or the formatting got changed.   What you say is true, but I don't think it changes the assertion.

mmccool:
I agree there are a lot of issues that are serialization related.   On the other hand I think being specific about things like where the `@language` tag can appear in the `@context` will make things more consistent and easier to parse.  

I'll see if I can find a good place to move the serialization constraints that are more logical.   I can update the PR tomorrow... I have to anyway to resolve the conflicts now.

index.html
mmccool: OK; consistency.

index.html
mmccool: This actually relates to some points in the internationalization questionnaire.   Anyway, it seems obvious, but I added it to make SURE that no one does this.   You could imagine someone doing something silly like processing the entire JSON document as an string...

index.html
mmccool: OK.

index.html
mmccool: Hmmm.... good point, I guess.   And to the other point.   Well, I thought I would throw this out there, I'm not married to it.

index.html
mmccool: See also the next item.  I can remove this if it would cause problems with TD generators based on a JSON-LD library.

index.html
mmccool: I'll see what I can do to clean this up.

mkovatsc:
+1 to define the allowed location of `@language` and its meaning of default language in the model (Section 5).

The serialization-specific issues including the i18n aspect should go into Section 6. There I would rename the titles and descriptions section to "Human-readable Metadata Serialization".

index.html
mkovatsc: I easily understood the explanation with "can't carry forward changes in direction from one string to another", opposed to the assertion text... maybe we can improve accordingly

index.html
mkovatsc: Lucky you :D
Good remove it.

index.html
mkovatsc: Victor updated the Model definitions. I think the terminology of this is now quite off. MultiLanguage is now a "Map", an "Object" with only "Terms", not "Vocabulary Terms" as names.

Each name of the <code>MultiLanguage</code> <a>Map</a> MUST be a language tag as defined in [[!BCP47]] (e.g., <code>en</code>, <code>de</code>, <code>ja</code>, <code>zh-Hans</code>, <code>zh-Hant</code>,
<code>az-Latn</code>, <code>az-Arab</code>).

index.html
mkovatsc: I think there should be a comment after the assertions that informs that this does not mean that Things should be able to dynamically generate their TD. This might be hard for constrained devices.

index.html
mkovatsc: This is a good candidate for Section 6.

index.html
mkovatsc: I easily understood the explanation with "can't carry forward changes in direction from one string to another", opposed to the assertion text... maybe we can improve accordingly.

This should go to Section 6.

index.html
mmccool: Done.

index.html
mmccool: Removed.   I also removed the assertion that was an aim to avoid duplicates.   However, this is a serialization issue and should be handled by JSON-LD.

index.html
mmccool: Phrase added.

index.html
mmccool: Removed.

index.html
mmccool: Yeah, some of this text is not mine, but needs to be updated for consistency.  Done.  A few other related things also updated.

index.html
mmccool: OK, I can see where "included" can make more sense since it is always part of the language tag.  Anyhow, done.

index.html
mmccool: Code markup removed.   I also added Azeri as an example of where this is needed.  I am not sure this goes in Section 6, as it is not a serialization issue.   Text direction is how the text is displayed when it is rendered, i.e. how the data is to be interpreted.   Therefore you also need the text direction in the text's information model.

index.html
mmccool: OK, I can add that comment.   SHOULD gives us some wiggle room.

mmccool: I added the following sentence immediately after this assertion (it is not itself an assertion, just a note): "Note however that constrained devices may choose to not support dynamically-generated TDs and do not need to support this feature."

mmccool:
I'm not convinced these text direction issues are serialization issues.  Text direction affects how the text is displayed, but does not affect how the bytes are ordered in storage or when communicated (which is what "serialization" means).   Given that, do you still feel I should move the points I left open above to Section 6?

mmccool:
I took a look at section 6 and noticed some new assertions had been added at some point that were redundant with the ones I had added under MultiLanguage, eg the use of language tags and inference of text direction.   However, I still assert that these are NOT serialization issues since the text direction does not affect how the data is stored, it affects how it is *interpreted*, so it's legitimately part of the information model.  However, redundancy is bad, so I removed the extra assertions (actually, they were tacked onto the end of another assertion and not given their own id, which was another problem...).   I did rename the section to "Human-Readable Metadata" and added a note about the fact the `@language` can be serialized in its own object or combined with others.

mmccool:
But HTTP language negotation is arguably part of serialization, so I did move that assertion to section 6.

index.html
mmccool: I left this in the MultiLanguage section, and as discussed below, removed a redundant assertion in Section 6.

mmccool:
Also moved assertion for "independent processing" to Section 6.

index.html
mmccool: I reworded the assertion a little but, but also added an "In other words..." statement after it to clarify.  Also moved to Section 6 since this is a serialization (or rather, a deserialization) concern.

mmccool:
I think all the issues raised in the review have been addressed, and this is ready to merge.

mmccool:
I created an additional PR to add a definition for "supportLanguage":https://github.com/mmccool/wot-thing-description/pull/8.    This is a modification of this PR.

index.html
```diff
@@ -2968,26 +3000,47 @@ <h3>Data Schema Serialization</h3>
...
   Another possibility is to set the default language with
-  the language negotiation mechanism of HTTP
-  (i.e., the <code>Accept-Language</code> header field).
+  a language negotiation mechanism, such as 
+  the <code>Accept-Language</code> header field of HTTP.
+  <span class="rfc2119-assertion" id="td-ns-multilanguage-content-negotiation">
+  In protocols such as HTTP that have an ability to negotiate content based on languages,
+  terms (e.g. <code>description</code> and <code>title</code>) intended
+  for carrying texts in the default language SHOULD be used instead of <code>MultiLanguage</code> container
+  when the texts are served as a result of such negotiation.</span>
```
mkovatsc: In protocols such as HTTP that have an ability to negotiate content based on languages,
TD documents SHOULD only include JSON members based on the default language
(e.g., <code>description</code> and <code>title</code>)
and no <code>MultiLanguage</code> objects,
while the <code>@language</code> member is omitted.

index.html
```diff
@@ -2968,26 +3000,47 @@ <h3>Data Schema Serialization</h3>
...
+  Note however that constrained devices may choose to not support dynamically-generated TDs
+  and do not need to support this feature.
+  </p>
+  <p>
+  As noted earlier, the base text direction used to display human-readable text is also inferred
```
mkovatsc: strike "also" before inferred

index.html
```diff
@@ -2968,26 +3000,47 @@ <h3>Data Schema Serialization</h3>
...
+  <span class="rfc2119-assertion" id="td-context-default-language-direction-independence">
+  Each human-readable string value MUST be processed independently, using the base
+  direction defined for that string.</span>  In other words, a processor cannot
+  carry forward changes in direction from one string to another, or infer direction
+  for one string from another one elsewhere in the TD.
   </p>
-
   <p>
   A TD snippet using <code>title</code> and <code>description</code> is 
   shown below. The default language is set to <code>"en"</code> through the
```
mkovatsc: remove quotes around `en` in code tags.

index.html
```diff
@@ -2968,26 +3000,47 @@ <h3>Data Schema Serialization</h3>
...
-  <code>@language</code> member within the <code>@context</code>.
+  definition of the <code>@language</code> name within the <code>@context</code>.
```
mkovatsc: <code>@language</code> JSON member within the <code>@context</code>.

index.html
```diff
@@ -2968,26 +3000,47 @@ <h3>Data Schema Serialization</h3>
...
+  This definition can be serialized either in its own object or
+  combined with other definitions.
```
mkovatsc: I would remove these two lines, as they only add confusion.

index.html
```diff
@@ -3018,17 +3071,20 @@ <h3><code>title(s)</code> and <code>description(s)</code></h3>
 
   <p>
   The <a>vocabulary terms</a> <code>titles</code> and <code>descriptions</code>
```
mkovatsc: The <a>Vocabulary Terms</a> ...

index.html
```diff
@@ -3018,17 +3071,20 @@ <h3><code>title(s)</code> and <code>description(s)</code></h3>
...
-  enable human-readable texts in multiple languages using language tags defined in [[!BCP47]].
+  enable human-readable texts in multiple languages using language tags defined in [[!BCP47]]
+  as described for the <code>MultiLanguage</code> Map.
```
mkovatsc: ... <a>Map</a>.

index.html
```diff
@@ -3018,17 +3071,20 @@ <h3><code>title(s)</code> and <code>description(s)</code></h3>
...
   MUST be serialized as JSON objects.
+  </span>
+<!-- The following is redundant given the definitions in MultiLanguage
   The names of their members MUST be language tags as defined in [[!BCP47]]
   (e.g., "en", "de", "ja", "zh-Hans", "zh-Hant")
   and the values JSON strings in the language indicated.
   If a script subtag changes the text direction of the JSON string,
   it MUST be included in the name of the member.
-  </span>
+-->
```
mkovatsc: Remove commented out lines (i.e., whole HTML comment)

mkovatsc:
@mmccool can you make sure that the default language also applies to the `name`? See comment in #630...

index.html
```diff
@@ -1400,9 +1400,34 @@ <h2>Core Vocabulary Definitions</h2>
 <p>
      <span class="rfc2119-assertion" id="td-context-default-language">
      In the <code>@context</code> array, the optional default language that is 
-     used throughout a TD document is defined within an object as a member with 
-     a key <code>@language</code> whose value MUST be a string representing a [[!BCP47]] 
-     language code.</span>
+     used throughout a TD document for human-readable strings outside of a <a href="#multilanguage">MultiLanguage</a> object
```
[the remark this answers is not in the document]
mmccool: So you are just saying expand "TD" to "Thing Description", and "object" to "instances", right?   I can do that.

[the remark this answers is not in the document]
mmccool: Done: changed to
"used throughout a Thing Description for human-readable strings outside of MultiLanguage instances"

index.html
```diff
@@ -1400,9 +1400,34 @@ <h2>Core Vocabulary Definitions</h2>
...
+     <li><span class="rfc2119-assertion" id="td-context-default-language-direction-inference">
+     The base text direction MUST be inferred from the default language.
+     </span></li>
+     <li><span class="rfc2119-assertion" id="td-context-default-language-direction-script">
+     In cases where a language can be written in more than one script which use 
+     different writing directions, the value used for `@language` MUST
+     include a script subtag so that an appropriate base direction can be inferred. 
+     </span></li>
+     <li><span class="rfc2119-assertion" id="td-context-default-language-direction-default">
+     If no default language is given, the base text direction MUST be assumed to be LTR (left-to-right).</span>
```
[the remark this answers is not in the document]
mmccool: No problem there.  Done.

index.html
```diff
@@ -1400,9 +1400,34 @@ <h2>Core Vocabulary Definitions</h2>
...
+     is also defined as a name 
+     <code>@language</code> whose value MUST be language tag from [[!BCP47]], 
+     possibly including a script subtag.</span>
+</p>
+<p>
+     Identification of the base direction of human-readable text in TDs 
+     is based on the following set of rules:
+     <ul>
+     <li><span class="rfc2119-assertion" id="td-context-default-language-direction-inference">
+     The base text direction MUST be inferred from the default language.
```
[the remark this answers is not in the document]
mmccool: Well... this is for texts outside of MultiLanguage, as stated in the introductory paragraph.   However, it is true that this assertion is not standalone.   Also, if we add my supportLanguage extension, there will be another exception.   However, I can replace this with your sentence (and extend it again in my extended PR for supportLanguage...)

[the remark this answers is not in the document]
mmccool: I changed this to
"Outside of  MultiLanguage instances the base text direction MUST be inferred from the default language tag."
and added the clarifying text after the list of rules:
"Rules inside MultiLanguage instances are similar but use the language tags given inside such instances instead of the default language tag."
Later, under MultiLanguage, another set of assertions is given.

index.html
mmccool: fixed.

index.html
mmccool: Eh, sure.

index.html
mmccool: Huh.  Not even sure that was my text, but I will fix it.

mmccool: Upon searching for where you noted this issue I notice a bunch of other places where Vocabulary Terms was not upper-cased and/or anchored.   Since I assume this was supposed to be a global convention for special terminology I updated them all (at least the ones in index.template.html; there may be others pulled in as part of rendering).

index.html
mmccool: Done.

index.html
mmccool: Not sure what you wanted here, however this was meant to be a reference back to the MultiLanguage section.   I updated it so it is an actual section reference; it now renders as the following:

The Vocabulary Terms titles and descriptions enable human-readable texts in multiple languages using language tags defined in [BCP47] as described in § 5.3.1.7 MultiLanguage.

index.html
```diff
@@ -1400,9 +1400,34 @@ <h2>Core Vocabulary Definitions</h2>
...
+     different writing directions, the value used for `@language` MUST
+     include a script subtag so that an appropriate base direction can be inferred. 
+     </span></li>
+     <li><span class="rfc2119-assertion" id="td-context-default-language-direction-default">
+     If no default language is given, the base text direction MUST be assumed to be LTR (left-to-right).</span>
+     This implies that if the language used in human readable text uses a script
+     that is written RTL (right-to-left),
+     the default language, potentially including a script subtag, needs to be explicitly specified.</li>
+     </ul>
+     Consumers and producers of TDs should be aware of certain special cases that can arise
```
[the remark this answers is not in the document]
mmccool: Ah, I see TD Processor is defined in Terminology to basically mean producer or consumer.  OK, done.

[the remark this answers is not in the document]
mmccool: Also added an anchor to refer back to terminology.

index.html
```diff
@@ -1400,9 +1400,34 @@ <h2>Core Vocabulary Definitions</h2>
 <p>
      <span class="rfc2119-assertion" id="td-context-default-language">
      In the <code>@context</code> array, the optional default language that is 
-     used throughout a TD document is defined within an object as a member with 
-     a key <code>@language</code> whose value MUST be a string representing a [[!BCP47]] 
-     language code.</span>
+     used throughout a TD document for human-readable strings outside of a <a href="#multilanguage">MultiLanguage</a> object
+     is also defined as a name
```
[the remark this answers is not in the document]
mmccool: Changed.

index.html
```diff
@@ -1493,13 +1518,20 @@ <h2>Core Vocabulary Definitions</h2>
...
-<section><h3><code>MultiLanguage</code></h3><p>Container to provide human-readable text in different languages using language tags described in [[BCP47]]. Each used language tags code MUST be decleared as vocabulary term member of this container (e.g., en, de, ja, zh-Hans, zh-Hant). See <a href="#titles-descriptions-serialization-json"></a> for example usages of this container in a Thing Description instance.</p>
+<section><h3><code>MultiLanguage</code></h3><p>Container to provide human-readable text in different languages using language tags described in [[BCP47]]. See <a href="#titles-descriptions-serialization-json"></a> for example usages of this container in a Thing Description instance.</p>
```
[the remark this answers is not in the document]
mmccool: Sure.   I also cleaned up the English in this sentence a little bit, specifically, it now says that this provides a *set* of texts.

index.html
```diff
@@ -1493,13 +1518,20 @@ <h2>Core Vocabulary Definitions</h2>
...
 <p>
-<span class="rfc2119-assertion" id="td-context-ns-multilanguage-content-negotiation">
-Note that in protocols such as HTTP that have an ability to negotiate content based on languages, 
-those terms (e.g. <code>description</code> and <code>title</code>) intended for carrying texts in a default language SHOULD be used instead of <code>MultiLanguage</code> container
-when the texts are served as a result such negotiation.
-</span>
+<span class="rfc2119-assertion" id="td-context-ns-multilanguage-language-tag"> 
+Each name of the <code>MultiLanguage</code> Map MUST be a language tag, potentially including
```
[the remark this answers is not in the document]
mmccool: Done.   Actually I linked all instances of Map in this section.

index.html
```diff
@@ -2968,26 +3000,47 @@ <h3>Data Schema Serialization</h3>
...
   The vocabulary terms <code>title</code> and <code>description</code> are 
   used to provide human-readable texts for titles (e.g., a display
   text for UI) and additional information useful for <a>Consumers</a> of TDs.
-  When a default language is specified by assigned a value to <code>@language</code>
+  When a default language is specified by assigning a value to <code>@language</code>
   in the <code>@context</code> of a <a href="#sec-thing-as-a-whole-json">Thing object</a>,
   that default language is associated with the string values assigned to
   <code>title</code> and <code>description</code> within the TD
-  and to <code>name</code> in the Thing object.
+  and to <code>name</code> and <code>support</code> in the Thing object.
```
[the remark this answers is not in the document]
mmccool: Yes, I think URLs just look at the order of bytes in memory, so the *rendering* direction does not matter.   So it's ok to say that the text direction of "support" is given by the default language even if the value is a URL.

index.html
mmccool: I am fine with adding an assertion to leave out the MultiLanguage objects.   I'm not convinced it is a good idea to leave out @language since if the TD gets forwarded somewhere it would be useful.  I think instead we should add an assertion that if it is included, it MUST be consistent with the negotiated default language.

mmccool: I added the following two assertions; notice the first one mandates leaving OUT MultiLanguage objects which seems simpler.
* In contexts where the default language has been negotiated, TD documents SHOULD NOT include any MultiLanguage objects. 
* In contexts where the default language has been negotiated, a @language definition, if present, MUST be consistent with the negotiated default language.

index.html
mmccool: sure.

index.html
mmccool: Actually, elsewhere you had me remove the "code" markup around language tags, too.   We should be consistent.   It looks a little odd with nothing around it, and it is a string value.   I think the other places language tags are given as examples they should get the "code" treatment to be consistent.

mmccool: OK, fixed... but I also went back and put the language tag examples in the "Multilanguage" section inside "code" markup to be consistent.

mmccool:
The intention is that the default language also applies to "name".   The text includes the following (in "Human-Readable Metadata"):
When a default language is specified by assigning a value to @language in the @context of a Thing object, that default language is associated with the string values assigned to title and description within the TD and to name and support in the Thing object.

index.html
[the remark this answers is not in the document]
mkovatsc: No, important is to drop "document", because "TD document" means serialized form.

index.html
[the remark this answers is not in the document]
mkovatsc: We should make "support" of type anyURI and make it independent of the default language.

index.html
mkovatsc: Good

index.html
mkovatsc: Sorry, I cannot remember saying that the code markup should be removed from language tag examples. As they are Terms, they should have code markup.

mmccool:
Note I have not yet updated the supportLanguage sub-PR.   I will do that soon, I have to fix some conflicts now.

mkovatsc:
Arg, GitHub eats up the HTML tags I wrote in the comments.. thus confusion in places where I wanted you to add the markup such as anchoring to the definition...

mkovatsc:
We have a few comments on skipping on the supportLanguage feature.

The name must be defined to underlie the default language. The other way round, the name given must be in the default language (i.e., no English name when comments are in Arabic).

Support must be independent of default language (to become an anyURI).

index.html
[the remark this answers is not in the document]
mkovatsc: Okay, is fixed correctly in the changes.

index.html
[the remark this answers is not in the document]
mmccool: Well, I did suggest that as an alternative solution for "support".  Personally I think it is cleaner.  It does not resolve the default-language problem for "name", though.

mmccool:
+1 for making support an anyURI.   Probably we should create another issue for it, although I could certainly add that change to this PR.

The default-language-changing-upon-negotiation problem still exists for "name" but if I make a few small changes to my sub-PR (eg changing "supportLanguage" to "nameLanguage") it would be one possible solution.   Note in the issue I created I made a list of other possible solutions.   One simply one would be just to ignore the default language for "name" and use a strong-first rule to determine the text direction.   That is a little fragile (and is "not recommended" in the i18n guidelines) but would be simple.

My preference though would be to define "nameLanguage" and in (an update to...) my sub-PR.

mkovatsc:
See #630 comment. It does not make sense to introduce `names` and nameLanguage when we have `titles` already.

index.html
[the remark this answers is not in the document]
mkovatsc: Let's put "internationalization of support fields" under "next TD spec version".
Need to address "name" issue, see #630.

mkovatsc:
We should merge this soon, as it has lots over good and important changes!
Can you fix

> <code>title</code> and <code>description</code> within the TD
  and to <code>name</code> and <code>support</code> in the Thing object

to only `title` and `description`? Then we should be good unless there are other places that indicate `name`.

index.html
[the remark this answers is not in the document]
mmccool: We can deal with this in a separate issue.

mkovatsc:
@mmccool Have you actively edited any of the ontology files or is it just render noise?

mmccool:
My edits were in index.template.html, validation/td-validation.ttl, and ontology/td.ttl.   Unfortunately there is some content hidden away in the ontology files that needed to be updated.  Everything else is render noise (specifically all of ontology/*.html and visualization/*.svg).

mmccool:
The last open issue in the reviews was about support.   I marked it as "resolved" but the true resolution will be to make it an anyURI.

I think this is ready to be merged now.

mmccool:
I will let you do the merge when you are ready.

mkovatsc:
Waiting for @sebastiankb as I am not officially an Editor of this draft...

mmccool:
Ah... well, I *am* an editor.  Let me go ahead and merge, then.

PR #623, opened 2019-05-05
Update terminology and references

mkovatsc:
I updated the term definitions imported from WoT Architecture based on which are used in the TD spec.

I updated the WoT Architecture reference and removed the wrongly normative reference to WoT Binding Templates. It is enough to reference the Binding Templates from the Arch spec.


mkovatsc:
At some point, I would like to make a full pass over the document to make defined terms consistent (capitalized, linked to Section 2).

PR #626, opened 2019-05-06
WIP: Major editorial pass

mkovatsc:
I implemented many editorial clean-ups. I marked defined terms consistently (might still have missed some).

I checked @vcharpenay 's updates to the TD Information Model definition and resulting changes to the serialization.
* A resulting glitch is now that there are no assertions on how to serialize specific class instances. The general instruction to serialize them as JSON objects is not an assertion. This is particularly bad for the serializations of (model) assigned Map and Array values.
* Assertions with "where the name is a <a>Vocabulary Term</a> included in the <a>Signatures</a> of ..." require an OR, not AND -- please verify
* Superclass is not defined, but used once
* "Every Simple Type mentioned in Section § 5.3 Class Definitions maps to a JSON type" -- unfortunately there is no list of the Simple Types

We need to check for a consistent usage of
* WoT Thing Description -- this should be only for referring to the building block or document
* Thing Description
* Thing Description instance
* TD
* TD instance
* TD Document -- this should be used primarily to refer to the JSON-based serializations
* TD Serialization -- this secondary
* TD model -- makes it clear we are in this self-defined Object world
* TD format
* TD representation format -- good to refer to the specified representation format
* ... more variations

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -1434,8 +1434,8 @@ In summary, this requires the following:
 
   <p>
   <span class="rfc2119-assertion" id="td-context">
-  The root JSON object of a TD serialization MUST include
-  a member with the name <code>@context</code> and a value of type
+  The root element of a <a>TD Serialization</a> MUST be a JSON object and it MUST
+  include a member with the name <code>@context</code> and a value of type
   string or array that contains <code>https://www.w3.org/2019/td/v1</code>.
   </span>
   In general, this URI is used to identify the
@@ -1510,8 +1510,10 @@ In summary, this requires the following:
   The value assigned to <code>properties</code> in a <code>Thing</code> instance
   is a <a>Map</a> of instances of <code>PropertyAffordance</code>.
   <span class="rfc2119-assertion" id="td-properties">
-    A map of <code>PropertyAffordance</code> instances MUST be serialized as an object
-    with a (unique) JSON name to identify each Property affordance object.
+    All name-value pairs of a <a>Map</a> of <code>PropertyAffordance</code> instances
+    MUST be serialized as members of the JSON object that results from serializing the <a>Map</a>;
+    the name of a pair MUST be serialized as a JSON string and the value of the pair, an
+    instance of <code>PropertyAffordance</code>, MUST be serialized as a JSON object.
   </span></p>
 
   <p>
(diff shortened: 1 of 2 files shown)


vcharpenay:
> Assertions with "where the name is a Vocabulary Term included in the Signatures of ..." require an OR, not AND -- please verify

it should be an AND. Assume the following abstract data structure (in some sort of mathematical notation):

```
{
  < "type", "string" >,
  < "observable", "true" >
}
```

If the assertion were with OR, it would suffice to serialize terms from `PropertyAffordance`to meet the constraint:

```json
{
  "observable": true
}
```

while the complete serialization should be:

```json
{
  "type": "string",
  "observable": true
}
```

vcharpenay:
@mkovatsc please review. (I haven't addressed your last comment on "TD *" usage.)

mkovatsc:
> where the name is a Vocabulary Term included in the Signatures of A and B

Signature of A: { "a", "aa" }
Signature of B: { "b", "bb" }

A Vocabulary Term "a" would only be included in the Signature of A, not B. "included in A and B" would be hence false. Thus, I still think it must be "included in Signatures of A or B", e.g., included in Signature of DataSchema or PropertyAffordance.

vcharpenay:
> A Vocabulary Term "a" would only be included in the Signature of A, not B. "included in A and B" would be hence false. Thus, I still think it must be "included in Signatures of A or B", e.g., included in Signature of DataSchema or PropertyAffordance.

Oh well, here comes the ambiguity of natural language. OK, I can do that change.

index.template.html
```diff
@@ -1282,6 +1282,10 @@ <h2>Mapping to JSON Types</h2>
         Values that are of type <code>double</code>
         MUST be serialized as JSON number.
   </span></li>
+  <li><span class="rfc2119-assertion" id="td-boolean-type">
+        Values that are of type <code>boolean</code>
+        MUST be serialized as JSON boolean.
```
mkovatsc: I think there was no JSON boolean, only the literal true and false

mkovatsc:
The other changes look good except:
* no JSON boolean IIRC
* what does the default value function return, when there is no default value? It says partial function, but not sure this explains enough...

mkovatsc:
Oh good, looking at RFC 8259, this boolean issue seems fixed!

Issue #611, opened 2019-05-04
Default value of "contentType" in "response"


mkovatsc:
Currently, the contentType of possible responses defaults to "application/json".

I think it would be better if it defaults to the contentType of the containing form.

If there is agreement, we would need an additional row in the table in 5.2 for `contentType` in Context `ExpectedResponse`.

sebastiankb:
I would propose to have such an entry in the table + a separate assertion explaining that the default value of the response contentType is always the same as that specified by the contentType in the Form class.

mkovatsc:
The current solution has the following problem:
There is a (fixed) Default Value for `contentType` in `reponse`. However, in the absence of `response`, we expect `contentType` in `reponse` to equal `contentType` in the Form object. This Default Value for `response` is not defined, as it is also dynamic...

I the following path:
1 Define the Default Value for `contentType` in `reponse` dynamically based on `contentType` in the Form object (i.e., dynamic). However, `reponse` in the Form object would still be undefined, so is there actually a value for `contentType` in `reponse`?
2. Define a Default Value for `reponse` in `Form`, which is also dynamic based on `contentType` in the Form object. However, it would be a complex type, a Class instance -- is that valid?

mkovatsc:
Good:
* Remove ExpectedResponse contentType from the (static) default value table
* make a simple assertion that in the absence of `repsonse` the expected response content type will be equal to the form content type

mkovatsc:
How could I move the comment back in time before yours?!

mkovatsc:
![image](https://user-images.githubusercontent.com/664090/57412457-023b2b80-71f1-11e9-8315-59e3570f1ba3.png)

Time traveling was also resolved. Closing.

PR #629, opened 2019-05-06
Default value table + reponse term handling in Forms

sebastiankb:
[The default value table](https://w3c.github.io/wot-thing-description/#sec-default-values) provides an empty cells for contentType, safe and idempotent. This PR assign the corresponding class + explains that the response term has to be used when the response content type differ from the form content type.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1641,7 +1641,8 @@ operations types SHOULD NOT be arbitrarily set by servients
 </p>
 <p>
 In some use cases such as for an <code>ActionAffordance</code> there is a different expectations of content types for the <code>input</code> 
-and <code>output</code>. In such a case the <code>response</code> term from the <a href="#expectedresponse"></a> class can be enabled to decleare different content types for requests and 
+and <code>output</code>. In such a case the <code>response</code> term that is typed with the <a href="#expectedresponse"></a> class can be enabled to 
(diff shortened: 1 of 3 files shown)


sebastiankb:
also see https://github.com/w3c/wot-thing-description/issues/611

index.html
```diff
@@ -1668,111 +1679,110 @@ <h3>Default Value Definitions</h3>
...
+                <td><code>ActionAffordance</code></td>
+                <td>
+                    <code>idempotent</code>
+                </td>
+                <td><code>false</code></td>
+            </tr>
+            <tr class="rfc2119-default-assertion" id="td-default-op-properties">
+                <td><code>PropertyAffordance</code></td>
+                <td>
+                    <code>op</code>
```
mkovatsc: Does not work for "Class". Did work when it was context, as op is not part of *Affordance...
Seems to be a big with the change from context.

sebastiankb: agree. should be changed back to Context

sebastiankb:
thanks @mkovatsc for your comment. new commit is pushed

sebastiankb:
as agreed with @vcharpenay Im going to merge the PR. Eventually @vcharpenay does some further updates on the table for default values

Issue #628, opened 2019-05-06
Inconsistency on accepted values for @context


egekorkan:
In the spec, there are multiple values for the `@context`. Below are the different ones:
* 5.3.1.1 Thing, below the table; Example 5: The mandatory context definition of the string "https://www.w3.org/2019/wot/td/v1" MUST be the first item in the @context array. ( I think that this is a bug, there shouldn't be a `wot`)
* Example 1 and 2 and pretty much everywhere: "@context": "https://www.w3.org/2019/td/v1" ( I think this is the correct one)
* Validation schema: 
```
                {
                    "type": "string",
                    "pattern": "^https?://(www\\.)?w3\\.org/2019/td/v1$"
                }
```

( This is a flexible version of the second one)
* In the comments of PR #603 @mkovatsc said that we have only the value with www which contradicts the previous point.

In the end, I think it should be clear if `@context` can be flexible/redirectable or not. Some questions I have:
* http or https
* www or not
* with year or not: *3.2 Alternative Namespaces* says that the namespace will redirect but is it OK to have a TD without the year that redirects to the latest version upon JSON-LD processing.

mkovatsc:
> In the comments of PR #603 @mkovatsc said that we have only the value with www which contradicts the previous point.

It rather says the JSON Schema needs to be made more restrictive

* the URI is used as an identifier of the format version, hence no flexibility
* with www
* with year (Alternative Namespaces are for the Vocabularies, not the context file)
* http vs https is still open for the Namespace IRIs, as we saw both practices (SSN with http, verifiable credentials with https)
* for the context file, it should probably be `https` -- other comments?

sebastiankb:
Regarding https vs http and www & without www I just tested the existing `w3.org/ns/td` with postman and all variations returned the same document:

- http://w3.org/ns/td
- https://w3.org/ns/td
- https://www.w3.org/ns/td
- http://www.w3.org/ns/td

egekorkan:
> We need firstly the unique identifier aspect. 

I agree with this, whether they return the same document can depend on the configuration of the Web server/DNS

sebastiankb:
I would simple go for https://www.w3.org/2019/td/v1 as announced in the TD spec and define this in the JSON Schema.

sebastiankb:
I fixed the namespaces errors (https://www.w3.org/2019/wot/td/v1 --> https://www.w3.org/2019/td/v1) in the spec

egekorkan:
[...]

PR #632, opened 2019-05-07
Update validation schema

egekorkan:
* I have constrained the schema to accept only the `@context` (or namespace) of the discussion in #628 
* We now use iri instead of uri (this was a bug noticed by @jmcanterafonseca )
* I have used the proposal in the PR #603 by @jmcanterafonseca to reference the context url instead of repeating it

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -1056,8 +1056,7 @@
         },
         "anyUri": {
             "type": "string",
-            "$comment": "Until iri-reference is supported by ajv, it will stay as uri-reference in playground",
-            "format": "uri-reference"
+            "format": "iri-reference"
         }
     },
     "type": "object",
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -1,14 +1,22 @@
 {
     "title": "WoT TD Schema - 29 March 2019",
     "description": "JSON Schema for validating TD instances against the TD model. TD instances can be with or without terms that have default values",
-    "$schema ": "http://json-schema.org/draft-06/schema#",
+    "$schema ": "http://json-schema.org/draft-07/schema#",
     "definitions": {
+        "thing-context-w3c-uri": {
+            "type": "string",
+            "enum": [
+                "https://www.w3.org/2019/td/v1"
+            ]
+        },
         "thing-context": {
-            "oneOf": [{
+            "oneOf": [
+                {
                     "type": "array",
                     "items": {
-                        "anyOf": [{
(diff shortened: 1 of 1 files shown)


mkovatsc:
@jmcanterafonseca , could you please review if your points are covered and the schema is correct?

mlagally:
@egekorkan I tried using https://jsonschemalint.com/#/version/draft-07/markup/json
to validate a TD.
Got an error message that iri-reference is unknown.
This may be a problem in the JSON-schema core file, but please take another look.

egekorkan:
So iri is a new thing in draft 7 and even though some validators claim to support draft they do not have support for iri. The website you are referring to uses ajv which doesn't support iri yet: https://ajv.js.org/#formats
You can try it with https://www.jsonschemavalidator.net/ which supports it.

Playground still uses uri-reference since it is also based on ajv

sebastiankb:
hmm... If the well known Lint Validator does not support this, then it is not a good sign. If I read the [comment ](https://ajv.js.org/#formats) correctly there will be also no updates on that in the future.

jmcanterafonseca:
@egekorkan LGTM, thanks. Wrt "iri" vs "uri-reference",  is a pity that ajv does not support it yet. However JSON Schema validator does, so no strong opinion here

mlagally:
@egekorkan One way of solving it would be to define something like an explicit iri-reference-td type in the schema by yourself, but IM''hO the underlying problem is, that some tools do not support unicode for (U/I) resource identifiers. 

This may even be more true for resource constrained devices and raises the following question:

Is it really useful to have unicode based resource identifiers or is it potentially damaging,
since it requires unicode support from a device.

Are there real world use cases?

egekorkan:
Well I can use URI and I haven't yet seen TDs that use IRI strings. In the end every URI is also a IRI. The problem is that the spec says anyUri which is basically IRI.

mlagally:
Every URI is an IRI, but not vice versa.
What is the definition of anyURI?

egekorkan:
>  What is the definition of anyURI? 

Clicking on any `anyURI` in the spec redirects you to http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#anyURI

This link has the following text: 
> [Definition:]   anyURI represents an Internationalized Resource Identifier Reference (IRI).  An anyURI value can be absolute or relative, and may have an optional fragment identifier (i.e., it may be an IRI Reference).  This type should be used when the value fulfills the role of an IRI, as defined in [RFC 3987] or its successor(s) in the IETF Standards Track.

PR #646, opened 2019-05-08
Align structure, fix #599, and address i18n comments

### LINKED ISSUE #599: Example 2 title could be improved

mkovatsc:
This PR aligns the structure with the WoT Architecture specification.
It also improves the order of some sections and gives consistent section headings.

#599 is fixed (example title)

This PR also includes fixes for the transition from Thing name to title.

It improves the language of Section 6 based in the I18N comments.


index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
-}
 </pre>
 
   <p>
-  The <code>contentType</code> member is used to assign a media types [[!IANA-MEDIA-TYPES]]
-  including media type parameters as attribute-value pairs separated by a <code>;</code> character.
-  Example:
+  Here is a more complex example: a TD snippet showing digest authentication
+  on a proxy combined with bearer token authentication on the <a>Thing</a>.
+  Here the <a>Default Value</a> of <code>in</code> in the <code>digest</code> scheme,
```
mlagally: This sentence is not very clear, please reword.
What you want to say, is that it is not required to explicitly include default values.

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
     MUST be serialized as JSON members of the root object.
   </span></p>
 
-  <pre class="example">
+  <p>
+    A snippet for a serialized root object with all defined members is given below:
```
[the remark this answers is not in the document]
mkovatsc: It includes also the optional ones.

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
+    "titles": {...},
     "description": "Human readable information.",
     "descriptions": {...},
     "support": "mailto:support@example.com",
     "version" : {...},
     "created" : "2018-11-14T19:10:23.824Z",
     "modified" : "2019-06-01T09:12:43.124Z",
     "securityDefinitions": {...},
-    "security": [...],
+    "security": ...,
```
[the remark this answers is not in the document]
mkovatsc: Because it may be a string or an array.

index.html
mkovatsc: Same text as ever since, only marked because moved.

mkovatsc:
Actual text changes are limited to assertions and sentences directly around assertions. The rest git diff logic when moving sections...

index.html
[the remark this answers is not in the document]
mmccool: Yes, that is fine.   The ... could be hiding an array, or a string.

index.html
```diff
@@ -837,14 +837,12 @@ <h1>Introduction</h1>
 
 
  
- <aside class="example" id="thing-description-full-serialization" title="Thing Description with context extension">
+ <aside class="example" id="thing-description-full-serialization" title="Thing Description with Context Extension for semantic annotations">
```
mmccool: Capitalization should be consistent: Thing Description with Context Extension for Semantic Annotations.   Or was Context Extension capitalized because it was is a special term, now?

index.html
```diff
@@ -907,15 +937,23 @@ <h2>Terminology</h2>
...
   <dfn>WoT Interface</dfn>,
   <dfn>WoT Runtime</dfn>,
-  etc. is defined in <a href="https://w3c.github.io/wot-architecture/#terminology">section 3</a>
-  of the WoT Architecture document [[WOT-ARCHITECTURE]].
+  etc. is defined in <a href="https://w3c.github.io/wot-architecture/#terminology">Section 3</a>
+  of the WoT Architecture specification [[WOT-ARCHITECTURE]].
   </p>
 
   <p>
       In addition, this document introduces the following definitions:
```
mmccool: I see the change from "document" to "specification".   Should this be changed too?
Note: Editorial.  This can change can be deferred, and personally I think either is fine in this case, but it would be good to be consistent.

index.html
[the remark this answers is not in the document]
mmccool: Maybe say "with all defined members, including all optional members, is given below:"

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
-    <code>PropertyAffordance</code>, <code>InteractionAffordance</code>, or <code>DataSchema</code>,
-    MUST be serialized as members of the JSON object that results from
-    serializing the <code>PropertyAffordance</code> instance, with the
-    <a>Vocabulary Term</a> as name.
-  </span>
-  See <a href="#data-schema-serialization-json" class="sec-ref"></a> for
-  details on serializing <a href="#dataschema" class="sec-ref"><code>DataSchema</code></a> instances.
+    JSON members named <code>title</code> and <code>description</code> are 
+    used within a TD document to provide human-readable metadata.
+    It can be used as comments for developers inspecting a TD document
```
mmccool: "It" -> "They" since there are two.   Also "or for display texts for a user interface"

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
-    <code>ActionAffordance</code>
-    MUST be serialized as JSON objects.
+  The JSON members named <code>titles</code> and <code>descriptions</code> are 
+  used within the TD document to provide human-readable metadata
+  in multiple languages within a single TD document.
+  <span class="rfc2119-assertion" id="td-multi-languages">
+    All name-value pairs of a <code>MultiLanguage</code> <a>Map</a>
+    MUST be serialized as members of a JSON object,
+    where the name is a well-formed language tag as defined by [[!BCP47]]
+    and the value the human-readable string.
```
mmccool: "the value is a human-readable string in the specified language".   The "is a" is an editorial grammar fix, the second part is a suggested addition.  It is obvious but because this is spec we should probably state it.   It should not be legal to say something is in Japanese and then given an English string...

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
   </p>
 
-<aside class="example" id="action-serialization-sample" title="Sample of an Action serialization">
-<pre>
+  <p>A TD snippet using <code>titles</code> and <code>descriptions</code> at different levels is given below:</p>
+    
+<pre class="example">
+{
+    "@context": "https://www.w3.org/2019/td/v1",
+    "title": "MyThing",
```
mmccool: I find it odd that this example has a title but no titles and descriptions but no description at the Thing level.   It's not wrong according to the spec, just odd.   On the other hand, such an example would probably break content negotiation.   I see that later this is expanded upon, but I kind of wonder at the value of this particular example.   Maybe should just skip it, and use a simple title+description one and another with paired title+titles and description+descriptions.

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
-  A TD snippet of an Event object is given below:
+  TD instances may also combine the use of <code>title</code> and <code>description</code>
+  with <code>titles</code> and <code>descriptions</code>.
+  <span class="rfc2119-assertion" id="td-title-description">
+  When <code>title</code> and <code>titles</code> or
+  <code>description</code> and <code>descriptions</code>
+  are present within the same JSON object,
+  the values of <code>title</code> and <code>description</code> MAY be seen as the default text.
+  </span>
+  These values underly the default language, which is usually set by the creator of the Thing Description instance.
```
mmccool: "underly" is not a word.   If you meant "underlay", that does not mean the right thing here.   I think you want to say "These values are expressed using the default language"

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
-Event affordances have been defined in a flexible manner, 
-in order to adopt existing (e.g., WebSub) or customer-oriented event mechanisms (e.g., Webhooks).
-For this reason, <code>subscription</code> and <code>cancellation</code> can be defined according to the desired mechanism. Please find further details in [[WoT-Binding-Templates]].
-Example <a href="#webhook-example-serialization"></a> illustrates how Events can use <code>subscription</code> and <code>cancellation</code> to describe Webhooks.
-</p>
+  Another possibility to set the default language is through a language negotiation mechanism,
+  such as the <code>Accept-Language</code> header field of HTTP.
+  <span class="rfc2119-assertion" id="td-ns-multilanguage-content-negotiation">
+    In cases where the default language has been negotiated,
+    an <code>@language</code> member MUST be present to indicate the result of the negotiation and the corresponding default language of the returned content.
```
mmccool: Yes, I agree this should be mandatory.

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
+  </span>
+  <span class="rfc2119-assertion" id="td-ns-multilanguage-content-negotiation-no-multi">
+    When the default language has been negotiated successfully,
+    TD documents SHOULD include the appropriate matching values for the members <code>title</code> and <code>description</code>
```
mmccool: Yes, better.

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
+    in preference to <code>MultiLanguage</code> objects in <code>titles</code> and <code>descriptions</code> members.
+  </span>
+  <span class="rfc2119-assertion" id="td-ns-multilanguage-content-negotiation-optional">
```
mmccool: Yes, good to make this an assertion.

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
   </span></p>
 
-  <p><span class="rfc2119-assertion" id="td-form-protocolbindings">
-  If required, form objects
-  MAY be supplemented with protocol-specific <a>Vocabulary Terms</a>
-  identified with a prefix.
-  </span>
-  See also <a href="#protocol-bindings"></a>.
-  </p>
+  <p>The recommended version identification pattern value is to rely on the semantic versioning policy [[?SemVer]].</p>
```
mmccool: The grammar of this sentence is a little broken.   I think it should be "The recommended practice is to use a version identification pattern based on semantic versioning [[...]]".    It might be good to make this a SHOULD assertion, too: "The version SHOULD be identified with a value based on the semantic versioning pattern [[...]]"

index.html
```diff
@@ -2120,588 +2138,378 @@ <h3>Thing Root Object</h3>
...
 
-  <p>A TD snippet using a URI Template and <code>uriVariables</code> is given below:</p>
+  <p>
+  The following TD snippet shows a simple security configuration specifying
+  basic username/password authentication in the header.
+  The value given for <code>in</code> is actually the <a>Default Value</a> (<code>header</code>)
+  and could be omitted.
+  First, a named security configuration must be given
+  in the <code>securityDefinitions</code> map.
+  Second, that definition must be activated by inclduing its JSON name in the
```
mmccool: "inclduing" -> "including" (typo).   Note: First, Second, implies they need to be in order.   This is not actually the case: it would be acceptable to reverse the order.   Not nice, but a library might do so.   I would just remove "First, " and "Second, " from these sentences.

index.html
mmccool: Maybe say "given in" to avoid having two "in"s in a row...

index.html
```diff
@@ -3036,272 +3245,114 @@ <h3>Data Schema Serialization</h3>
...
 }
 </pre>
 
-<section>
+<section class="informative">
 <h3>Semantic Annotations</h3>
 
     <p>
-    The context extension allows to add additional vocabulary to a Thing Description instance.
+    <a>TD Context Extensions</a> allow to add additional vocabulary to a Thing Description instance.
```
mmccool: Grammar fix: "allow the addition of".   Also, should it be "vocabulary terms"?   Not sure about the latter, it depends on what the convention is (now).

index.html
```diff
@@ -3412,24 +3476,22 @@ <h3>Adding Protocol Bindings</h3>
...
-  for the version defined by this specification)
-  can be defined using the vocabulary extension mechanism.
-  For this example, we will use a fictional ACE security scheme
-  based on [[?draft-ietf-ace-oauth-authz]] and that is, for this example,
+  Finally, new security schemes that are not included in <a href="#sec-security-vocabulary-definition"></a>
+  can be imported using the <a>TD Context Extension</a> mechanism.
+  This example uses a fictional ACE security scheme
+  based on [[?draft-ietf-ace-oauth-authz]] that is, for this example,
   defined by the namespace at <code>http://www.example.org/ace-security#</code>.
   Note that such addiational security schemes must be <a>Subclasses</a> of the
```
mmccool: "addiational" -> "additional".   Nice improved example, BTW!

mmccool:
Since this PR changes the structure of the document it should be merged as soon as possible so that other PRs can be completed and merged.   Since most of the above requested changes are "local", I recommend going ahead and merging this PR, then addressing the above problems with a second PR (which, being a collection of small local changes, will be less likely to result in conflicts).

sebastiankb:
There are conflicts

index.html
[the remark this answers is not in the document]
mkovatsc: REsolved saying "including all optional"

index.html
mkovatsc: In the <code>digest</code> scheme, the <a>Default Value</a> of <code>in</code>
  (i.e., <code>header</code>) is omitted, but still applies.

index.html
mkovatsc: Defined terms capitalized

index.html
mkovatsc: CHanged to specification

index.html
mkovatsc: Fixed

index.html
mkovatsc: Clarified

index.html
mkovatsc: Fixed.

index.html
mkovatsc: The the language of the default text is indicated by the default language,

index.html
mkovatsc: Added assertion

index.html
mkovatsc: FIxed.

index.html
mkovatsc: allow for additional <a>Vocabulary Terms</a> to a Thing Description instance.

index.html
mkovatsc: Fixed

mkovatsc:
Resolved conflicts and addressed both review comments.

PR #651, opened 2019-05-08
Limit at-risk items to those necessary for CR transition

mmccool:
Previously the at-risk highlighting identified all features for which we did not have an automated test, but this is unnecessarily restrictive for the CR transition document.  This updates the at-risk highlighting to identify features for which it is possible we will not have an implementation by the time of the PR transition.  In summary, these are basically a set of security schemes for which we do not have implementations yet, one feature in the base security scheme that we do not have two implementations of yet (proxy), the qop feature in Digest and some other schemes where we do not yet have an implementation, and one feature in Form (scopes) which is in support of the OAuth2 security scheme (but if it is omitted, then scopes in Form should also be omitted).

Every other assertion is therefore considered NOT to be at risk.

Also, within sections marked at risk, I have supressed the table highlighting since technically EVERY element in those tables would be marked as at risk.   That just adds a lot of clutter and is redundant.


mmccool:
Note: it should be possible to merge this PR without conflicts even if the base document changes significantly, since most changes were to testing/inputs/atrisk.csv. However, there were a few small changes to index.template.html to clean up the at-risk highlighting explanation at the very start of the document. Diffs in index.html and in other places are "render noise" and can be ignored.

PR #655, opened 2019-05-08
Add cwt to list of supported formats for SecurityScheme

mmccool:
This is a VERY simple PR that just adds "cwt" to the list of allowed values for "format" in SecuritySchemes.    

If I have time tomorrow I will expand upon it (eg add normative references defining what CWT means, updating the ontology to use an actual type, etc) but this is the worst-case fallback plan.


ontology/hyperm.html
```diff
@@ -88,18 +88,18 @@ <h2>Introduction</h2>
...
-<tr><td>type</td><td><a href="#hyperm-hintsatmediatype">hyperm:hintsAtMediaType</a></td></tr>
-<tr><td>Form</td><td><a href="#hyperm-form">hyperm:Form</a></td></tr>
+<section class="specterm"><h4><a href="#hyperm-hasanchor">hyperm:hasAnchor</a></h4><p>IRI: https://www.w3.org/2019/wot/hypermedia#hasAnchor</p><p><strong>a OWL Class</strong></p><em>hasAnchor</em> - <span>By default, the context, or anchor, of a link conveyed in the Link header field is the URL of the representation it is associated with, as defined in RFC7231, Section 3.1.4.1, and is serialized as a URI.</span><table><tbody></tbody></table></section></section></section><section><h2>Term Mapping</h2><table><thead><tr><th>JSON String</th><th>Ontology Term</th></tr></thead><tbody><tr><td>op</td><td><a href="#hyperm-foroperationtype">hyperm:forOperationType</a></td></tr>
```
mmccool: This is just "render noise": as far as I can tell, the definitions in this file are generated in a random order each time... but the same definitions are there as before.  Probably this file should not be checked in...

ontology/wot-security.ttl
```diff
@@ -74,7 +74,7 @@
                                            
                                            rdfs:label "format" ;
                                            
-                                           rdfs:comment "Specifies format of security authentication information (one of jwt, jwe, or jws)."@en .
+                                           rdfs:comment "Specifies format of security authentication information (one of jwt, cwt, jwe, or jws)."@en .
```
mmccool: This is the ACTUAL change.

mmccool:
Ready for review.

mmccool:
Sorry, few small changes: made alg and format explicitly extensible.

mmccool:
I am going to resist making any more changes at this point and will wait for review feedback; then we can discuss in the upcoming TD call.

ontology/wot-security.ttl
```diff
@@ -218,7 +218,7 @@
                                                       
                                                       rdfs:subClassOf <https://www.w3.org/2019/wot/security#SecurityScheme> ;
                                                       
-                                                      rdfs:comment "Certificate-base asymmetric key security configuration identified by the vocabulary term <code>cert</code> (i.e., \"scheme\": \"cert\")." .
+                                                      rdfs:comment "Certificate-base asymmetric key security configuration conformant with [[!X509V3]] identified by the <a>Vocabulary Term</a> <code>cert</code> (i.e., <code>\"scheme\": \"cert\"</code>)." .
```
mkovatsc: "-base" --> "-based"
(d missing)

ontology/wot-security.ttl
```diff
@@ -84,7 +84,7 @@
                                              
                                              rdfs:label "identity" ;
                                              
-                                             rdfs:comment "Identifier validating access by providing information unique to the resource." .
+                                             rdfs:comment "Identifier providing information unique to the resource which can be used for selection or confirmation." .
```
mkovatsc: Not unique to the resource!
> Unique name within the security configuration referring to a particular secret for gaining access.

mmccool: Ok, I will delete the phrase "unique to the resource".   So it will be "Identifier providing information which..."

mmccool:
BTW, here is a diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fcdn.staticaly.com%2Fgh%2Fw3c%2Fwot-thing-description%2Fmaster%2Findex.html&doc2=https%3A%2F%2Fcdn.staticaly.com%2Fgh%2Fmmccool%2Fwot-thing-description%2Fadd-cwt-format%2Findex.html

ontology/wot-security.ttl
mmccool: Thx, will fix.

mmccool:
Review issues have been addressed as suggested.

mkovatsc:
Missing statement

The value assigned to the name `scheme` MUST be defined within a Vocabulary included in the Thing Description, either in the standard Vocabulary defined in §5 or in a TD Context Extension.

mmccool:
Got it.   I was wondering.

mkovatsc:
Can go into SecurityScheme class definition (the text after its definition table)

mmccool:
Will do so.

mmccool:
In a moment... need to call back into Arch call.

mmccool:
Done.   SecurityScheme section now starts with (with all the markup, links, assertion ids, etc):
Metadata describing the configuration of a security mechanism. The value assigned to the name scheme MUST be defined within a Vocabulary included in the Thing Description, either in the standard Vocabulary defined in § 5. TD Information Model or in a TD Context Extension.

PR #658, opened 2019-05-08
Update Link and Form

egekorkan:
I have changed their order according to what @mkovatsc mentioned in the call
I have also added highlighting to specific terms in form and link to increase readability


egekorkan:
I am not able to resolve conflict since it says that the conflict is too complex. It is only a couple of lines, any ideas why it seems complicated to GitHub?

vcharpenay:
It is because the file was deleted in master while you edited it in this branch. I can resolve it locally.

mkovatsc:
#659 is about to resolve this by copying over your edits to the new file environment.

egekorkan:
# 659 doesn't contain the changes of this PR. In particular, the order of Link and Form isn't changed and the highlighting is also not there. I will wait for the updates of #659

egekorkan:
Can be closed, thank you @sebastiankb

PR #687, opened 2019-05-14
Edited UML diagrams in Section 5.1 Overview

vcharpenay:
see `images/uml-overview.pptx` for the source file.


sebastiankb:
thanks. we should also remove the Editor's note:

>The following diagrams are automatically generated from the ontology files. The layout will be improved upon publication of Working Drafts.

vcharpenay:
It's included in commit 78fb658f69db34cc0142101570e9cc01ae77f161 (second one).

mkovatsc:
I marked this with the red "by CR transition", as we should include this figure fix for the CR publication.
We should merge it only after merging the "ralph-swick" branch into master.

PR #688, opened 2019-05-14
Refactored RDF examples for simplicity

vcharpenay:
- introductory uses an example namespace, to remove unnecessary details
- section 7 was edited, there are now 2 examples for data schema annotations (29, 30).
- appendix was modified accordingly (text is based on examples of section 7)


vcharpenay:
Example 29 was removed, as it introduces some feature for the first time (annotation with an embedded JSON-LD context) and not enough explanation is given in the text.

vcharpenay:
the changes of the PR should be taken into account before CR is published: TD examples with annotations are not fully consistent with the TD information model (in particular, `@type` is defined for instances of `Form` instead of `InteractionAffordance`).

index.html
```diff
@@ -3168,8 +3167,11 @@ <h3>Semantic Annotations</h3>
 
     <p>
       As an example, the TD snippet given below extends the version information container
-      by addind version numbers for the hardware and firmware of the Thing,
-      semantically annotates a Property, and uses a well-defined value for the data schema unit.
+      by adding version numbers for the hardware and firmware of the <a>Thing</a>, and uses
+      values from external <a>Vocabularies</a> for the <a>Thing</a> and for the
+      data schema unit. SAREF [[smartM2M]] and OM, the Ontology of Units of Measure
```
mlagally: Suggest to make the links to <a>SAREF</a> and <a>OM</a> and change the other occurrences in the text.
Are SAREF and OM the *only* ontologies to be used for the home domain, or just examples?

index.html
```diff
@@ -3205,43 +3206,51 @@ <h3>Semantic Annotations</h3>
...
+        
     <p>
-        In many cases, context extension may be used to annotate pieces of
-        a data schema, as in the TD snippet below. The example references
-        SSN/SOSA [[vocab-ssn]] and OM 1.8, an ontology for units of measure
-        [[Rijgersberg-et-al-2013]]. Note
-        that object assigned to <code>temperature</code> is a data schema and not
-        the actual temperature property of the physical world object. More details
-        on modeling the physical world can be found in Appendix
+        The TD snippet below uses the well-known SAREF external <a>Vocabulary</a> to
```
mlagally: Please add a link to SAREF.

index.html
```diff
@@ -3205,43 +3206,51 @@ <h3>Semantic Annotations</h3>
...
+        describe the state of a lamp. The external <a>Vocabulary Term</a> <code>ssn:forProperty</code>,
+        taken from the SSN <a>Vocabulary</a> [[vocab-ssn]], is being used to link the data schema of the
```
mlagally: Please add a link to <a>SSN Vocabulary</a>

index.html
```diff
@@ -5529,16 +5537,20 @@ <h3>Thing Description JSON-LD Context</h3>
     "id": "urn:dev:ops:32473-WoTLamp-1234",
     "title": "MyLampThing",
     "@type": "saref:LightSwitch",
+    "saref:hasState": {
+        "@id": "urn:dev:ops:32473-WoTLamp-1234/state",
+        "@type": "saref:OnOffState"
```
mlagally: The type OnOffState in SAREF has blanks. 'saref:On off state'
If we have a name without blanks, who is doing the mapping?

index.html
```diff
@@ -5529,16 +5537,20 @@ <h3>Thing Description JSON-LD Context</h3>
...
+    },
     "securityDefinitions": {"basic_sc": {
         "scheme": "basic",
         "in": "header"
     }},
     "security": ["basic_sc"],
     "properties": {
         "status": {
+            "ssn:forProperty": "urn:dev:ops:32473-WoTLamp-1234/state",
```
mlagally: Where is the ssn namespace defined?

index.html
```diff
@@ -5733,51 +5744,73 @@ <h3>Thing Description Ontology</h3>
...
-                <code>writeproperty</code> on the <code>saref:GetCommand</code> or an
-                <code>invokeaction</code> on the <code>saref:ToggleCommand</code>. Note that the
-                result of the operation cannot be known for certain, as the TD does not expose
-                its internal state. However, by selecting certain models of the physical world,
+                Note that, in the case the internal state is hidden, the result of the operation cannot
+                be known for certain. However, by selecting certain models of the physical world,
                 either by statistical means or by further axiomatization, clients can still
-                reach a certain level of autonomy with an ontological approach.
+                reach a certain level of autonomy with an ontological approach and e.g. learn from their
+                own behavior via a trial and error procedure.
```
mlagally: Is it the client's behavior or the behavior of the thing?
I'm not sure a "trial and error" procedure is practical in the real world.

index.html
vcharpenay: I assume you refer to the HTML documentation of SAREF showing 'saref:On off state'. This UI is not the best but the full IRI is shown right above the title as `https://w3id.org/saref#OnOffState`:

image

The title, with spaces, is in fact made of the human-readable label of the term, appended to the vocabulary prefix. It is however just a rendered view and not the canonical IRI for the term. I understand it is confusing.

index.html
vcharpenay: was a mistake, thanks! Fixed by 957294e.

index.html
vcharpenay: There is indeed an earlier occurrence of SAREF/OM at the end of Section 6.3.10. My suggestion is to remove these occurrences and simply point to Section 7.1 (as is already the case).

These vocabularies are examples (by design, everyone is free to expose their own vocabulary, including in the home automation domain). The words "among others" can be added at the end of the sentence.

index.html
vcharpenay: I don't know about the typographic conventions for W3C recs but I assumed a reference should appear only once: the reference for SAREF appears one section above.

index.html
vcharpenay: The reference that follows points directly to the (human-readable documentation of the) SSN vocabulary. Are you expecting something else?

index.html
vcharpenay: I am indeed referring to the behavior of the client, that learns by interacting with Things. I realize that according to the WoT Architecture terminology, it should be called a [consumer](https://www.w3.org/TR/wot-architecture/#dfn-consumer) instead of a client. I can update the text.

vcharpenay: See 21221db2a992ed592b615f57baf5e697892183b3.

index.html
vcharpenay: See cf53ebf4600df3119cdabc5acf1df2b5219199ed.

index.html
vcharpenay: See cf53ebf4600df3119cdabc5acf1df2b5219199ed.

index.html
vcharpenay: See a7ebc649e1430822a4d501ec1cde1874afa232a9 and cf53ebf4600df3119cdabc5acf1df2b5219199ed.

index.html
```diff
@@ -5741,27 +5751,28 @@ <h3>Thing Description Ontology</h3>
...
                 that is, conceptualizations of the physical world. To this end, an ontology does not cover
                 one possible model of the physical world by constructing abstract data structures. Instead,
                 it provides necessary constraints on all possible models of the world for these models to
                 be sound. This principle, at the basis of <em>model theory</em>, is what the semantics of RDF
                 (and OWL) is based on [[rdf11-mt]]. In the following diagram, because the only fact
-                we know about <code>urn:dev:ops:32473-WoTLamp-1234</code> is that it is a light
-                switch with generic affordances, it could be any light switch, be it mechanical,
-                digital or even virtual.
+                we know about <code>urn:dev:ops:32473-WoTLamp-1234</code> is that it is a <a>Thing</a>
+                with generic affordances, it could be <em>any</em> thing, including intangible entities.
```
sebastiankb: <em>any</em> thing --><em>any</em> <a>Thing</a>

index.html
```diff
@@ -575,29 +575,29 @@ <h1>Introduction</h1>
     {
         "@context": [
             "https://www.w3.org/2019/wot/td/v1",
-            { "saref": "https://w3id.org/saref#" }
+            { "eg": "http://example.org/iot#" }
```
sebastiankb: Example 27 uses "iot" as prefix. However, example 21 uses "eg". We should use always the same prefix for  "http://www.example.org/iot" for any example in the spec.

vcharpenay: In the latest committed version, I don't see the `iot` prefix anymore?

takuki:
Can @vcharpenay also describe this changes in the Recent Changes section?

vcharpenay:
thanks for the review, @takuki. Section 7 is informative, indeed: what I did is that I removed the "This section is non-normative." statement from all sub-sections and put it only once at the beginning of the whole section, instead.

index.template.html
```diff
@@ -2871,8 +2867,13 @@ <h3>Semantic Annotations</h3>
...
       As an example, the TD snippet given below extends the version information container
-      by adding version numbers for the hardware and firmware of the Thing,
-      semantically annotates a Property, and uses a well-defined value for the data schema unit.
+      by adding version numbers for the hardware and firmware of the <a>Thing</a>, and uses
+      values from external <a>Vocabularies</a> for the <a>Thing</a> and for the
+      data schema unit: <a href="https://w3id.org/saref">SAREF</a>,
+      also used in <a href="#thing-description-full-serialization"></a>, and 
+      <a href="http://www.ontology-of-units-of-measure.org/resource/om-2/">OM</a>, the
+      Ontology of Units of Measure [[Rijgersberg-et-al-2013]]. These <a>Vocabularies</a>
+      are used as example: other ones may exist, in particular in the home automation domain.
```
mlagally: Suggestion: "... used as examples - others may exist ..."

mlagally:
This looks very good now, please consider the suggestion for a small language improvement.

There's a respec error being reported (probably not caused by your changes), but ideally you would also fix it.

index.template.html
vcharpenay: see 39fd7b0.

vcharpenay:
Commit  202e788 fixes the ReSpec issue.

vcharpenay:
@takuki @sebastiankb @mlagally I'll merge this PR at the end of the day (~5pm CET) if you have no further comment.

PR #690, opened 2019-05-15
Annotate at-risk features with searchable text

mmccool:
- As requested, annotate at-risk features with searchable text.  "This feature is at risk" has been added to all table entries that were marked as at risk.
- Default values associated with other features at risk are also at risk but are in a different section.  I added a reference to the default value section to the SOTD section but did not list the specific default values (this would be redundant, as they are associated with the other features at risk listed immediately above).
- Cleanup in the SOTD section to mention the above annotation and some grammar fixes.  Changes to talk about "features" and "sections" where appropriate rather than "Text or table entries".

We actually don't have any assertions at risk so this COULD be simplified further to just talk about table entries.   But I decided to leave it alone.  For the time being I also left the "possible" wording alone for the producers/consumers statement although I think "appropriate" would be a better word here.

There were also some ReSpec issues having to do with metadata mandatory for CR.   This PR also adds the following to the preamble:
```
            , crEnd: "2019-06-18"
            , previousMaturity: "WD"
            , previousPublishDate: "2018-10-21"
            , implementationReportURI:      "https://w3c.github.io/wot-thing-description/testing/report.html"
```
Of these, `crEnd` is particularly important: it is the day just before the date (2019-06-19) by which we will have to make a resolution to proceed to PR.   Implementors have until this crEnd to complete their implementation and file their implementation report data.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,7 +7,7 @@
     <script class='remove'>
           var respecConfig = {
               specStatus:     "CR"
-            , crEnd: "2019-06-18"
+            , crEnd: "2019-06-13"
             , previousMaturity: "WD"
             , previousPublishDate: "2018-10-21"
             , implementationReportURI:      "https://w3c.github.io/wot-thing-description/testing/report.html"
@@ -420,10 +420,10 @@ a[href].internalDFN {
   The Working Group seeks implementation feedback, having set the
   requirement of at least two implementations of each feature as the
   exit criteria for the Candidate Recommendation phase. 
-  <font color="red">The group aims
-  to obtain reports from one producer and one consumer for each
-  feature if possible.</font> 
-  For details, see the draft 
+  The group aims
+  to obtain reports from one TD producer and one TD consumer for each
+  feature if applicable.
+  For details, including definitions of implementation, TD producer, and TD consumer, see the draft 
   <a href="https://w3c.github.io/wot-thing-description/testing/report.html">implementation
   report</a>.
   </p>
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -7,6 +7,10 @@
     <script class='remove'>
           var respecConfig = {
               specStatus:     "CR"
+            , crEnd: "2019-06-18"
```
mkovatsc: Where does the June 18th come from?
We have:
* June 13: Deadline for CR Comments
* June 20: PR Transition Request
* June 27: Publication of PR
*

index.html
```diff
@@ -390,29 +394,36 @@
   insufficient implementation experience (reports).
   </p>
   <ul>
-  <li>Whole features described in <a href="#apikeysecurityscheme"></a>, <a href="#certsecurityscheme"></a>, 
+  <li>Whole sections related to the security schemes described in <a href="#apikeysecurityscheme"></a>, <a href="#certsecurityscheme"></a>,
```
mkovatsc: Wouldn't it be better to call these security schemes features?
I think the usual terminology is "features at risk", no?

mkovatsc: `<li>Security schemes <a href="...">APIKeySecurityScheme</a>, CertSecurityScheme, PSKSecurityScheme, PublicSecurityScheme, PoPSecurityScheme, and OAuth2SecurityScheme</li>`
?

mkovatsc:
I marked this with the red CR label, as this PR is meant to be merged before the CR publication.
We also still need a PR to merge the ralph-swick branch into master -- and the figure PR #687 after that into master as well.

PR #691, opened 2019-05-15
Edited UML diagrams in Section 5.1 Overview (bis)

vcharpenay:
(see #687: same commits without unnecessary files)


sebastiankb:
Looks good. I do not see any errors

Issue #702, opened 2019-05-19
[Testing] Wrong default value in assertion descriptions of td-default-X


egekorkan:
The issue applies to the following assertions so I think it is a code bug that introduces this bug:

td-default-alg
td-default-alg-pop
td-default-contentType
td-default-flow
td-default-format
td-default-format-pop
td-default-in-apikey
td-default-in-basic
td-default-in-bearer
td-default-in-digest
td-default-in-pop
td-default-op-actions
td-default-op-events
td-default-op-properties
td-default-qop
td-default-readOnly
td-default-safe

In all of them the description is like:
> The value associated with member " BasicSecurityScheme " if not given MUST be assumed to have the default value " in ".

It should be:
> The value associated with member " BasicSecurityScheme " if not given MUST be assumed to have the default value "header".

mmccool:
OK, let me take a look; looks like a bug in assertions.js.

mmccool:
I put a fix into PR https://github.com/w3c/wot-thing-description/pull/707.   I was pulling the columns out of the table in the wrong order...

mkovatsc:
Was merged and implementation report looks fine now.
Closing.

Issue #701, opened 2019-05-19
[Testing] td-default-idempotent is used for writeOnly


egekorkan:
The assertion with id `td-default-idempotent` that is `td-default-idempotent-1` in the implementation report should not be called this but `td-default-writeOnly` since it's description is 
> The value associated with member "DataSchema" if not given MUST be assumed to have the default value " writeOnly ".

Applying this will remove the duplicate id as well

mmccool:
Fixed in PR https://github.com/w3c/wot-thing-description/pull/707

mmccool:
Looks like this was fixed already in the CR version, too, by @ashimura

mkovatsc:
Implementation report looks correct.
Closing.

PR #707, opened 2019-05-20
Update Test results post-CR (using only TDs provided by implementers)

mmccool:
Note: still waiting to merge with duplicate id fixes from @ashimura; however, I will deal with those in a separate PR as I need to update these test results now.

- Recent test run with new AssertionTester results and using CR draft
- Using only validated test TDs provided by implementors

Also addresses issues
- https://github.com/w3c/wot-thing-description/issues/701 
- https://github.com/w3c/wot-thing-description/issues/702
- https://github.com/w3c/wot-thing-description/issues/710

Rendered Implementation Report Draft: https://cdn.statically.io/gh/mmccool/wot-thing-description/master/testing/report.html?env=dev


mmccool:
So, there are a LOT of errors now, i.e. lines with 0-0-0-0 results.  What happened?  Was there a mass renaming of ids?   Are partial test results not being generated if TDs don't validate, e.g. because of the "title" change?

mmccool:
Note that a list of TDs that are failing validation can be generated by doing a "grep" on the output of the update.sh script, which I have checked in as "wot/testing/tests/2019-05/update.log":
```
git clone https://github.com/w3c/wot.git
cd wot/testing/tests/2019-05
grep -B 1 "ERROR" update.log
```

egekorkan:
An invalid td (such as not having title) will cause a lot of problems in the beginning and stop the validation. I am guessing it is because of that since the playground examples pass. The examples are [here](https://github.com/thingweb/thingweb-playground/tree/master/WebContent/Examples/Valid)

mmccool:
Yes, I figured the invalid TDs (and they were basically ALL invalid) were causing the problem.   To test, I updated all Intel TDs to pass validation and the results are much more reasonable.   But everyone needs to finish their updates.   For the record, here is what I had to fix:
- "name" -> "title" in Thing
- "lastModified" -> "modified" in Thing (note also that DateTime needs to fully satisfy the XML DateTime schema.  In particular notice that seconds in the time are mandatory...)
- Update the mandatory context reference

mmccool:
So the Intel TDs can now also be used as "valid" examples.

mmccool:
I will still leave this as WIP until everyone updates their TDs so they validate.   Hopefully people can do this before the Wed call.   One of use should send an email to the group reminding people to do this.   Ege, can you do that, and point to the playground?   I will take care of merging any PRs updating inputs and will rerun the tests as I see new results...

egekorkan:
An email is sent to the WG with necessary informations

sebastiankb:
Siemens just updated their TDs

mmccool:
Thanks.  Panasonic also updated their TDs.   I manually updated the Oracle and Fujitsu TDs to get a more complete report but as these are not implementor-provided updates I created a separate PR(s) for that(those).

PR #716, opened 2019-05-30
reflected feedback from CR except rfc2119-assertion

ashimura:
talked with @mmccool and decided to split [the changes between the Editor's draft and the CR version](https://github.com/w3c/wot-thing-description/blob/master/publication/cr/the_diff) into:
1. minor edits
1. rfc2119-assertion class which is used to generate the implementation report

This pullrequest is for 1 above, and McCool is working on 2 above.

Note that we need to convert index.template.html to index.html using the render.sh script.
Also need to generate the implementation report from index.html using the assertion.js script.


mmccool:
Actually, Kaz, that was not what I meant.   I understand your changes to ids, it is better if you merge them and I update.   So please do so, I will watch for the changes (and fix things if necessary).

mmccool:
OK, I will update the IDs (fixing duplicates) as part of a testing update...

Issue #721, opened 2019-05-31
[Testing] Additional assertions needed for JSON open ecosystem requirements


mmccool:
Some assertions are needed to confirm that implementations conform with the "open ecosystem" requirements for JSON:

- [x] 2 additional assertions for td-json-open

egekorkan:
This is already tested but the online results show null. Are you sure that you have taken the latest version of manual assertions?

mmccool:
There are some sub-assertions for td-json-open having to do with byte order.  However, acceptance is a "MAY".   We only need one implementation to do a MAY.   It turns out node-wot uses node.js JSON.parse that throws an error if it sees this... but there is an option to let it accept these.  Node-wot will be updated to use this option (@sebastiankb and @danielpeintner is assigned)

takuki:
> TD Processors MAY ignore the presence of a byte order mark rather than treating it as an error.

I do not think this assertion belongs to TD specification. TD is in JSON format which is defined by RFC 8259. This assertion belongs to RFC 8259.

We are not testing RFC 8259. Therefore, I suggest to demote this assertion to a statement without an asserting qualification.

danielpeintner:
I agree with Taki that this does not really belong to the TD. 

Anyhow, I added a quick check to node-wot that removes a potential Byte Order Mark (BOM), see https://github.com/eclipse/thingweb.node-wot/commit/c2278c7fabd80dd0eb3c54d072020814781513e6

mkovatsc:
I agree that this is redundant, as we clearly build on top of RFC 8259. So even if we remove this, it still applies...

mkovatsc:
Anyhow, we cannot just remove assertions. And we have enough implementation results now.
Closing.

PR #766, opened 2019-06-12
A requirement (Can ignore BOM) is not an assertion of TD (Issue #721)

takuki:
The assertion does not belong to TD specification.
This is an assertion in RFC 8259.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1807,9 +1807,10 @@ In summary, this requires the following:
 <li><span class="rfc2119-assertion" id="td-json-open_utf-8">TDs MUST be encoded using UTF-8 [[!RFC3629]].</span</li>
 <li><span class="rfc2119-assertion" id="td-json-open_no-byte-order">Implementations MUST NOT 
   add a byte order mark (U+FEFF) to the beginning of a TD document.</span></li>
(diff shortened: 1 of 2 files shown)


mkovatsc:
I agree that this is redundant, as we clearly build on top of RFC 8259. So even if we remove this, it still applies...

mmccool:
Yes, I added these mostly as an aid to testing.   In a future version we can replace these with compliance with RFC8259, as long as we are clear that TDs are an "open ecosystem".   For now though, as Matthais says, removing them (which, BTW, we can't do in CR status) won't change our testing requirements, and leaving them in makes our test requirements clearer.

mmccool:
To be clearer: this PR must NOT be merged until after the REC has been released.   We can't take assertions out at this point.   I have added a label to make this clearer.

takuki:
As long as we can successfully test this JSON feature in time, everything is fine.
Please remember we are testing interoperability of TD specification.
If we are expected to test assertions in underlying specification, we would be overwhelmed.

takuki:
As I wrote in the last comment, this PR is no longer necessary though the technical point is still valid.

Issue #738, opened 2019-06-01
[Testing] Additional Manual Validation Needed for Text Direction Assertions


mmccool:
Text-direction assertions are constraints on the rendering of strings found in TDs and so must be validated in Consumers with multilanguage text display capabilities.

Additional manual validation is needed for the following assertions, which have to do with selection of the base text direction:
- [x] td-context-default-language-direction-heuristic	
- [x] td-context-default-language-direction-independence
- [x] td-context-default-language-direction-inference	
- [x] td-context-default-language-direction-script
- [x] td-context-ns-multilanguage-text-direction-infer

The "direction-script" assertion requires some peculiar test cases as there are only a few languages that can be written in two different scripts with different base directions.  One is Azeri, which can be written in Latin script (az-Latn) or Arabic script (az-Arab).   So we also need a TD with Azeri descriptions to test against as well as two Consumers that can render Azeri... or at least can output a base text direction inference.

mmccool:
So it's going to be harder to come up with test cases for direction-script than I thought.   My original idea was to generate some simple test cases (say, for a Light) using online translators but apparently az-Arab is only used in Iran, and all the online dictionaries, wikipedia, and google translate seem to only support az-Latn.  If anyone knows of a translator that can output az-Arab that would be great (even from az-Latn, which I can get; az-Arab is also not really a Latin script, BTW, it is just based on it but has a bunch of special characters, like the schwa).   I haven't seen any other mention of a language written today using both an RTL and LTR script.   There are some other languages like Javanese that historically used Arabic script but most have converted to Latin script.

mmccool:
So... I found the following OSS tool to convert Azeri scripts: http://azconvert.sourceforge.net.  It only runs on Windows though, and I'm not entirely confident about installing this on my work Windows machine, for security reasons.   I do have a sandboxed/sacrificial Windows machine I could use for this purpose, though; once I generate a few text examples I will share them here.

mmccool:
See new intel-camera TD.   Test results updated, which addressed some of the multilanguage testing gaps and provides a test case for the consumers needed for the above assertions.

k-toumura:
@mmccool @mkovatsc This is screenshot of bidirectional support in Node Generator.
<img width="968" alt="Screen Shot 2019-06-07 at 1 11 15" src="https://user-images.githubusercontent.com/30313213/59072001-345bae00-88c1-11e9-9cd3-69a56734f1af.png">
[...]

PR #768, opened 2019-06-12
String-meta requirement is not an assertion in TD (Issue #738)

takuki:
td-context-default-language-direction-independence assertion belongs to string-meta.


mmccool:
Same issue as with the BOM requirement.  It's a requirement of string-meta, so we could replace this with a more general assertion about string-meta.   But at this point I would not merge this PR as we can't remove assertions in the CR phase.   At the worst this is simply redundant/more detail than is necessary.   However, taking it out would not remove the requirement to test it.

takuki:
This PR is no longer necessary though the technical point is still valid.

PR #771, opened 2019-06-14
Update oracle.html

mlagally:
update description to reflect the Munich demo scenarios.


mkovatsc:
BTW, it turned out https://github.com/w3c/wot/tree/master/testing/tests/2019-05/descriptions is the origin for the HTML descriptions.

@mmccool with this redundancy, I am not sure what is actually the intended version by people. We might need to check with everyone and then kill one of the duplicates...

PR #777, opened 2019-06-18
Add RFC references for security schemes

mkovatsc:
The security schemes Basic, Digest, and Bearer lack the basic reference to their specification.

This PR adds references to the corresponding RFCs.


mkovatsc:
Merging as discussed in the call on 19 June 2019.

Issue #700, opened 2019-05-18
Figure 2 title "JSON schema vocabulary" should be "Data Schema Vocabulary"


takuki:
as pointed out in May 15 WoT Main teleconference.

takuki:
Fixed. See https://github.com/w3c/wot-thing-description/pull/780 .

mkovatsc:
Thanks, @takuki !

PR #780, opened 2019-06-19
Figure 2 title was changed to "Data Schema Vocabulary" (Issue #700)

takuki:
It was originally "JSON Schema Vocabulary".


mkovatsc:
Fixes the issue.
Capitalization is not Chicago style, but consistent with other captions.
Merging.

PR #782, opened 2019-06-20
Remove Appendix D (see resolution)

mkovatsc:
Removed Appendix D about RDF processing.
Pointing to ontology documentation files under the namespace IRIs instead now.
Removed text about SHACL.

### DIFF
--- a/index.html
+++ b/index.html
@@ -704,10 +704,11 @@ a[href].internalDFN {
 </aside>
 
 <p>
-        The declaration mechanism inside some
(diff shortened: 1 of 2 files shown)


mkovatsc:
Please at one review to check spelling etc.

mkovatsc:
I talked with @vcharpenay directly and he agreed. He is traveling at the moment, hence cannot comment here.

Will merge then. Thanks!

PR #785, opened 2019-06-20
Remove heading 7.2

mkovatsc:
This is an undo of a change introduced by @vcharpenay 

He added "7.2 Annotating Data Schemas" but then does not show how data schemas are annotated (which would be an `@type` alongside the `type` of JSON Schema).

This PR simply removes the wrong heading, but keeps the section text (basically reverts to the structure of the CR).

### DIFF
--- a/index.html
+++ b/index.html
@@ -3335,12 +3335,6 @@ Example <a href="#webhook-example-serialization"></a> illustrates how Events can
 }
 </pre>
 
-</section>
-
-<section>
-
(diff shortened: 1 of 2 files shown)


takuki:
I think this makes sense.

It appears Example 29 uses "ssn:forProperty" instead of @type annotation.

It may be similar to @type annotation, but to avoid confusion, I agree to remove the potentially misleading header.

Issue #792, opened 2019-07-12
Remove or update github links in the spec


sebastiankb:
The specification contains many references to documents located on github. E.g.,

>All participating entities in the W3C Web of Things, that is, Things, Consumers, and Intermediaries as defined in the Web of Things (WoT) Architecture. 

Web of Things (WoT) Architecture is linked to the architecture document hosted on github. The w3c domain should be used instead.

I will setup a PR about this.

sebastiankb:
PR was merged

PR #802, opened 2019-08-13
Update few links + add chapter link 

sebastiankb:
address issue #792 + add chapter link in TD template section


takuki:
We will need to use dated URI before becoming a Recommendation.
We can do it after PR publication.

Issue #808, opened 2019-08-30
Proposals are referenced by Google and have wrong date


egekorkan:
Similar to the need to create have an ARCHIVE repo on w3c/wot we need it here. E.g. searching for `simplified thing description w3c` on google returns the following document: https://w3c.github.io/wot-thing-description/proposals/simplified-td/

This document has the date of the official TD

egekorkan:
From the respec documentation, we can set a static publication date for such documents. See https://github.com/w3c/respec/wiki/publishDate

danielpeintner:
Note: in general there are ways to *tell* search engines that certain folders should not be indexed/removed.

e.g., robots.txt file in root with
```
User-agent: *
Disallow: /proposals/
```

sebastiankb:
issue solved by PR #816

k-toumura:
Unfortunately, robot.txt files in subdirectories are not valid.

in [Robots.txt Specifications](https://developers.google.com/search/reference/robots_txt):
> Crawlers don't check for robots.txt files in subdirectories.

PR #814, opened 2019-09-12
remove simplified TD proposal

sebastiankb:
see #808


danielpeintner:
I tend to agree with @egekorkan: keeping the document seems fine.
We just need to tell search engines "do not index" it

see https://github.com/w3c/wot-thing-description/issues/808#issuecomment-531144866

sebastiankb:
ok. I will follow your proposal + change the date. I will provide a new PR

Issue #730, opened 2019-05-31
[At-Risk] Implementations of OAuth2SecurityScheme needed


mmccool:
The OAuth2SecurityScheme is currently at risk.
- 2 full implementation (with all sub-features, including the various flows) are needed

mmccool:
@mmccool to create a PR that limits support to the code flow.   We have a code flow implementation from Panasonic, both a producer and a consumer, and also a simulator and real device.  @ToruKawaguchi  to work on this.

yamagile:
@mmccool , I created the [PR](w3c/wot#819) on this issue. It makes some assertions passed like below:

---

pass: 0->2

* td-arrays_scopes
* td-security-oauth2-code-flow
* td-vocab-authorization--OAuth2SecurityScheme
* td-vocab-flow--OAuth2SecurityScheme
* td-vocab-in--APIKeySecurityScheme
* td-vocab-name--APIKeySecurityScheme
* td-vocab-refresh--OAuth2SecurityScheme
* td-vocab-scheme--SecurityScheme_apikey
* td-vocab-scheme--SecurityScheme_oauth2
* td-vocab-scopes--Form
* td-vocab-scopes--OAuth2SecurityScheme
* td-vocab-token--OAuth2SecurityScheme
* td-vocab-authorization--PoPSecurityScheme (This may be a bug of the assertion tester because I have no change for PoPSecurityScheme.)

---

pass: 1->2

* td-default-flow

---

pass: 0->0

* td-security-oauth2-implicit-flow
* td-security-oauth2-password-client-flow
    * **note**: There is only code flow implementation from Panasonic.

mkovatsc:
We have implementations for OAuth 2 code flow.
Ideally, we only drop the unimplemented sub-features, but keep the security scheme `oauth2` for code flow.

sebastiankb:
@mkovatsc 
> Ideally, we only drop the unimplemented sub-features

What does it mean? Do we have to remove some terms from the https://w3c.github.io/wot-thing-description/#oauth2securityscheme ?

mkovatsc:
We have no positive implementation report for the implicit, password, and client flows. Hence, to my understanding, the corresponding assertions and Vocabulary Terms need to be removed:

* For the implicit flow authorization MUST be included.
* For the password and client flows token MUST be included.
* `string (one of implicit, password, client, or code)` -- in the Type row for the `flow` field of OAuth2SecurityScheme

takuki:
[...]

PR #822, opened 2019-10-15
Make oauth2/flow mandatory, remove default

mmccool:
This PR addresses a problem noted in Issue https://github.com/w3c/wot-thing-description/issues/730 that the default value for flow ("implicit") has no implementations.  This PR therefore makes "flow" mandatory and removes the default value.

This does not technically resolve the issue since we are still missing some implementations for some OAuth2 flows.  In fact right now we only have two implementations, as far as I know, for the "code" flow.  However, by removing the default we make it possible to accept a modified Oauth2 spec supporting only the "code" flow and then adding back in the other flows once we have implementations, i.e. for the updated TD spec next year.

Summary of changes:
- Make "flow" mandatory in table under oauth2 SecurityScheme
- Remove "flow" default value entry in "Defaults" table
- In JSON Schema, mark "flow" as "required".

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/822.html" title="Last updated on Oct 15, 2019, 5:40 AM UTC (6cfb00e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/822/1d99000...mmccool:6cfb00e.html" title="Last updated on Oct 15, 2019, 5:40 AM UTC (6cfb00e)">Diff</a>


sebastiankb:
looks good

sebastiankb:
in today's TD meeting we decided to merge this PR

PR #827, opened 2019-10-16
WIP: Define Terminology Section 3 as non-normative

sebastiankb:
This PR defines the terminology section as non-normative.

However, it must be decided whether this should really be the case. Depends on the decision of the Arch document whether the terminology section is also informative.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/827.html" title="Last updated on Oct 16, 2019, 2:53 PM UTC (32679b0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/827/72188d1...32679b0.html" title="Last updated on Oct 16, 2019, 2:53 PM UTC (32679b0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -693,7 +693,7 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
 
 </section>
 
-<section id="terminology" class="normative"> 
+<section id="terminology" class="informative"> 
 <h2>Terminology</h2>
 
   <p>The fundamental WoT terminology such as
--- a/index.template.html
+++ b/index.template.html
@@ -585,7 +585,7 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
 
 </section>
 
-<section id="terminology" class="normative"> 
+<section id="terminology" class="informative"> 
 <h2>Terminology</h2>
 
   <p>The fundamental WoT terminology such as


mlagally:
The Architecture spec preserves the terminology section as being normative.

sebastiankb:
good news. Have a short discussion on this tomorrow's TD call and let's close this PR without PR

ashimura:
Well, I'm a bit confused...

As I mentioned within my message sent to the group list and during the main call, etc., terminology sections don't have to be normative and can be informative. So I checked with Philippe and confirmed our usage of terminology doesn't have to be normative, so the terminology sections of Thing Description and  WoT Architecture can be informative. That's why I suggested we make them informative for the 2nd CR publication.

Please imagine we get a possible comment for the 2nd CRs to ask us change/add some terminology definition. If the terminology sections are normative, we need to publish yet another CR to address that comment.

Let's talk about that during the TD call today (Oct. 18). Maybe I should file an issue for WoT Architecture as well.

mmccool:
As I mentioned in the Arch call, I think it would be odd for TD to be informative and the Arch to be normative.   That said, we have to make the dependency of the TD on Arch informative.  That means the terminology dependency is informative... and it would be weird to have an informative dependency in a normative section (the Terminology section of the TD) because then some terminology would be informative and some normative.

It would be a LOT simpler to just make both Terminology sections in both specifications informative.   AFAIK we don't actually have any vocabulary specifications that are so precise (eg date formats) that we NEED a normative section for defining them.

ashimura:
I'd agree with @mmccool :)

Anyway, let's talk about this during the TD call later.

sebastiankb:
group decided to merge this PR since we want to reduce the risk of any concerns about this section during review

PR #849, opened 2019-11-19
fix indent

crossjs:


dontcallmedom:
Marked as non substantive for IPR from ash-nazg.

sebastiankb:
thanks for the fix

Issue #788, opened 2019-07-11
Explainer: Mention alternative rejected solutions


kenchris:
In response to https://github.com/w3ctag/design-reviews/issues/357

It is common that explainers have a section listing some of the other solutions that were considered and rejected. This really helps us understand why the chosen solution is better and how you have learned from existing and prior solutions.

takuki:
In [7/12 TD telecon](https://www.w3.org/2019/07/12-wot-td-minutes.html), @sebastiankb volunteered to update [TD Explainer](https://github.com/w3c/wot-thing-description/blob/TD-TAG-review/explainer/Explainer.md) add some description about the difference from the design of existing technologies such as OpenAPI and RAML.

sebastiankb:
the PR is merged. it provides a related work sections with alternative approaches. I will close this issue

PR #855, opened 2019-12-05
provide a related work in explainer

sebastiankb:
provide a related work for explainer

also see #788


sebastiankb:
the group decided to merge this PR

PR #856, opened 2019-12-06
Remove at-risk features and update implementation report for PR transition

mmccool:
The intention of this pull request is to remove all at-risk items from the specification and from the implementation report.   At-risk items have generally been commented out rather than deleted so they can be added back in later (eg for 1.1).   The exception is the JSON-schema, since we do not want the commented-out items to appear in the spec.  In this case I copied the old file to one starting with ext- so it can be easily reverted.

Not quite finished, there are a few small issues:
1. Figures till mention at-risk items such as the cert, public, and pop security schemes.  The autogenerated SVG files were updated correctly, but the spec actually uses PNG files which seem to have been manually generated at some point (but there was no SVG source checked in?).   Anyway, these need to be updated.
2. td-writeall-consumer is still showing up as at-risk in the implementation report and in section 6.3.9 of the spec.   Is this a testing mistake?   Should this be removed?  EDIT: I removed this (commented it out) so the current PR omits it, but we can discuss.
3. the JSON-schema has been updated to remove the at-risk items but has not been tested
4. the wording of the implementation report needs to be reviewed to make sure it reflects PR status rather than CR status
5. we need to confirm if only the code flow is supported by oauth2 implementations.  There was some discussion that client may also have been supported.  Unfortunately one of the at-risk assertions mentioned both client (which MIGHT be supported) and password (which is NOT), so the testing was failing for this assertion... but it might be due to password.   We really need to break this up into two separately tested assertions.   For now the conservative option is to remove it (and support for the client flow).   Also, since only the code flow is supported in the flow enum for OAuth, there is a strange "one of (code)" defintion of allowable values for "flow".   This is strange but should be left as it is as it allows us to easily add back in the other flows in the future.  I updated an OAuth2 example that was using the implicit flow to use the code flow as well.  This required adding a (mandatory) token field.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/856.html" title="Last updated on Dec 6, 2019, 8:07 AM UTC (514d505)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/856/93bc859...mmccool:514d505.html" title="Last updated on Dec 6, 2019, 8:07 AM UTC (514d505)">Diff</a>


mmccool:
I removed the writeall-consumer assertion, but we should discuss; I can revert if there are two implementations willing to add manual assertions.

sebastiankb:
we should change the the flow type definition from "string (one of code)" to "string (e.g., code)"

sebastiankb:
the infromation "In the case of operation type writeallproperties , it is expected that the Consumer provides ALL writable (non readOnly ) properties and the (new) assigned values (e.g., within payload). Otherwise the Thing MAY refuse this call to avoid inconsistency. " should be not removed.

It should be transformed to a 'non-assertation' statement.

sebastiankb:
group decided to merge this PR. A second PR will be provided by @mmccool that addresses the comments above

PR #867, opened 2019-12-11
WIP: Fix #839: Clarify DataSchema vs contentType

### LINKED ISSUE #839: Clarify InteractionAffordance::contentType vs DataSchema::type
Coming from [Scripting issue 198](https://github.com/w3c/wot-scripting-api/issues/198), should we clarify the following:

1. The TD spec should contain normative prose on how to represent binary stream data.
What is the `DataSchema::type` of a Property when it contains a form with  `contentType` `"image/jpeg"` for instance? 
We need normative prose on how  `DataSchema::type` and `InteractionAffordance::contentType` should be used together, or separately.

2. The TD doc can be serialized only as UTF-8 (without BOM). But what does "string" type mean then? How do you represent string payloads from sensors that are UTF16BE or UTF16LE? 

3.  In [section 5.4](https://w3c.github.io/wot-thing-description/#sec-default-values), `contentType` is a `Form`. That looks like a bug (or else needs explaining). BTW there should be links in that table.

4. In the example in [Appendix B](https://w3c.github.io/wot-thing-description/#json-schema-for-validation), some `"response"` elements contain `contentType` as an object (a `DataSchema` as it looks). But that contradicts the definition of [`ExpectedResponse`](https://w3c.github.io/wot-thing-description/#expectedresponse) which expects a string, so this is a bug.

Overall I have the impression the handling of `contentType` has changed with time in the TD spec and now we have inconsistencies because of that.

zolkis:
This is a change to [DataSchema Vocabulary](https://w3c.github.io/wot-thing-description/#sec-data-schema-vocabulary-definition) section from line 1268, but it could be also moved to the [Representation Formats / Data Schemas](https://w3c.github.io/wot-thing-description/#data-schema-serialization-json) section.

Rendered version can also be checked here:
https://zolkis.github.io/wot-td/index.html#sec-data-schema-vocabulary-definition 

In addition, my editor removed end-of-line spaces (which are considered errors usually), that's a lot of fake change. BTW the spec source would need some formatting alignments.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/wot-thing-description/pull/867.html" title="Last updated on Dec 11, 2019, 9:28 AM UTC (a110a59)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/867/111c078...zolkis:a110a59.html" title="Last updated on Dec 11, 2019, 9:28 AM UTC (a110a59)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -379,24 +379,24 @@ a[href].internalDFN {
 </head>
 <body>
 
-  
+
 
   <section id="abstract">
   <p>
-  This document describes a formal model and a common representation 
-  for a Web of Things (WoT) Thing Description. 
+  This document describes a formal model and a common representation
+  for a Web of Things (WoT) Thing Description.
   A Thing Description describes the metadata and interfaces of <a>Things</a>,
   where a <a>Thing</a> is an abstraction of a physical or virtual entity that
-  provides interactions to and participates in the Web of Things. 
-  Thing Descriptions provide a set of interactions based on a small vocabulary 
-  that makes it possible both to integrate diverse devices and 
-  to allow diverse applications to interoperate. 
+  provides interactions to and participates in the Web of Things.
+  Thing Descriptions provide a set of interactions based on a small vocabulary
+  that makes it possible both to integrate diverse devices and
+  to allow diverse applications to interoperate.
   Thing Descriptions, by default, are encoded in a JSON format that also allows
   JSON-LD processing. The latter provides a powerful foundation to represent
   knowledge about <a>Things</a> in a machine-understandable way.
   A Thing Description instance can be hosted by the <a>Thing</a> itself or hosted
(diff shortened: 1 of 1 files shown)


zolkis:
Feel free to make commits to this PR
@danielpeintner @mmccool

mmccool:
Upon review it seems that there is content here, in the contents of the table, that could be interpreted as normative.  So I don't think we CAN merge it at this point.  There is also a MAY in there and we can't add RFC2119 assertions, so at the very least it would have to be reverted to informative text.   So despite the signicant improvements to the text, I think we have to defer this to the next version, and not include it in the PR release.

zolkis:
I think we should discard this PR in favor of #869. That can be handled at nearest convenience.

PR #869, opened 2019-12-11
[WIP] feat: add content type tables with data schema

danielpeintner:
***
<a href="https://pr-preview.s3.amazonaws.com/danielpeintner/wot-thing-description/pull/869.html" title="Last updated on Dec 11, 2019, 12:55 PM UTC (37fba04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/869/0fef606...danielpeintner:37fba04.html" title="Last updated on Dec 11, 2019, 12:55 PM UTC (37fba04)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1264,6 +1264,7 @@ where a sequence of three numbers separated by a dot indicates the major version
     
       <section id="sec-data-schema-vocabulary-definition">
       <h2>Data Schema Vocabulary Definitions</h2>
+      <p>A data schema is an abstract notation for data contained in data formats.</p>
       <p>
         The data schema vocabulary definition is reflecting a very common subset of the terms defined by JSON Schema [[?JSON-SCHEMA]]. It is noted
         that data schema definitions within Thing Description instances are not limited to this defined subset and may use additional terms
@@ -1273,8 +1274,7 @@ where a sequence of three numbers separated by a dot indicates the major version
         <a href="#json-ld-ctx-usage"></a> and the documentation under the namespace IRIs, e.g.,
         <a href="https://www.w3.org/2019/wot/td">https://www.w3.org/2019/wot/td</a>).
       </p>
-      <p>A data schema is an abstract notation for data contained in data formats. 
-         In a TD, concrete data formats are specified in Forms (see <a href="#form"></a>) using content types.
+      <p>In a TD, concrete data formats are specified in Forms (see <a href="#form"></a>) using content types.
          When the value of a content type in an instance of the Form is <code>application/json</code>, 
          the data schema can be processed directly by JSON Schema processors.
          Otherwise, Web of Things (WoT) Binding Templates [[?WOT-BINDING-TEMPLATES]] defines data schema's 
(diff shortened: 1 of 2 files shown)


danielpeintner:
meant to replace https://github.com/w3c/wot-thing-description/pull/867 with no whitepace diffs and modifying `index.template.html` instead of `index.html`

egekorkan:
I have found some new information regarding this.

Actually, JSON Schema allows the description of non-JSON data by using the keywords `contentEncoding` and `contentMediaType` (see [here](https://json-schema.org/understanding-json-schema/reference/non_json_data.html)). This could already cause a problem if someone uses these keywords with some values in the data schema level but uses other values in the form level by using TD specific keywords `contentType` and `contentCoding`. 

On the other hand, openAPI describes such payloads with their own schema where a string can have a `format` called `binary`, see [here](https://swagger.io/docs/specification/describing-request-body/file-upload/) or end of the page [here](https://swagger.io/docs/specification/media-types/). However, the format keyword does not have a value called `binary` according to JSON Schema. @handrews had mentioned that openAPI will be fully JSON Schema compatible so maybe there will be some alignment in the future?

relu91:
Great job finding this out. I think stating `contentEncoding` and `contentMediaType` at DataSchema, it is a great solution. It avoids a lot of confusion and it makes easier to understand the data returned from Action/Property/Event. Plus, if it is already in JSON-schema and, therefore, it may be a well-known concept in the community.

handrews:
@egekorkan @relu91 OpenAPI will definitely adopt the `content*` keywords along with the rest of JSON Schema Core and Validation in OAS 3.1.  They will probably deprecate the use of the custom `format`.  So their `binary` and `base64` formats would be discouraged starting in OAS 3.1 (and maybe removed in OAS 4.0?  I'm a little unclear on the process).

TBH, I am hoping that as people come to understand extension vocabularies, we (the JSON Schema project proper) can get rid of `format`.  It's a horrible mess of a keyword that attempts to do too much and does so unreliably- it's never had very good support in implementations.

For example, a small vocabulary of date-time related keywords would be quite easy to implement- you could easily use regular expressions or any number of commonly available libraries to implement such checks as reliable validation assertions.

On the other hand, determining whether an email address is "valid" is (I've found) an incredibly contentious subject, to the point that some implementations attempt to do regex validation, some pass it off to a 3rd-party libraries, and others just check to see if there's an "@" somewhere in the string.

Since both of those are stuffed into `format`, you end up with a keyword with wildly varying support depending on its value and the choices made by the implementation.  It's a mess.

zolkis:
> I think stating contentEncoding and contentMediaType at DataSchema, it is a great solution.

I agree. Of course it needs to be aligned with the Forms contentType, but that should be easy, since DataSchema is supposed to specify Form/contentType further.

Then we could use a generic value type like mentioned [here](https://github.com/w3c/wot-scripting-api/issues/201#issuecomment-573702999).

relu91:
> I agree. Of course it needs to be aligned with the Forms contentType, but that should be easy, since DataSchema is supposed to specify Form/contentType further

While working on a list of Scripting API use cases and examples, I was wondering which is the most correct way to represent strings that encode content types. At  first, I tried this solution: 
```javascript
// image property affordance
"image": { 
	"description" : "image", 
	"type": "string"
	"forms": [
		{ 
			"op": "readproperty", 
			"href": "coaps://mylamp.example.com/lastPicture", 
			"cov:methodName" : "GET",
			"contentType": "image/png",
	                "contentCoding": "base64"
		}
	] 
}}
```

Then I thought about your comment and I came up with this solution (which I prefer):
```javascript
// image property affordance
"image": { 
	"description" : "image", 
	"type": "string",
        "contentType": "image/png",
	"contentCoding": "base64"
	"forms": [
		{ 
			"op": "readproperty", 
			"href": "coaps://mylamp.example.com/lastPicture", 
			"cov:methodName" : "GET",
                        "contentType": "application/json" 
		}
	] 
}}
```
This solution has the benifit that it clearly states which is the "basic" contentType of the response and it further specify it inside the data schema. It can even be exploited on other content types since  data schema is not limeted only on `application/json` as suggested in https://github.com/w3c/wot-thing-description/issues/839#issuecomment-552438034 . 

I guess this example (or another similar example) deserves a place inside the spec so that it easier to understand the relation between DataSchema and form content type. Should we open a new issue for that?  

p.s. 
> On the other hand, determining whether an email address is "valid" is (I've found) an incredibly contentious subject, to the point that some implementations attempt to do regex validation, some pass it off to a 3rd-party libraries, and others just check to see if there's an "@" somewhere in the string.

I know this problem quite well too. In the end, I implemented a regex method plus a DNS call to understand if the domain after `@` was a "real" domain. Quite tricky but it was fun. I described the process and the final result [here](https://www.instagram.com/p/BkaAnVFhh5s/)

egekorkan:
So from the Consumer point of view, the first method would rely on the Scripting API implementation (i.e. node-wot) to handle the payload and represent to the scripting level an image  (or whatever the payload is). In the second case, the script would have to call some functions to convert the obtained JSON string into an image. 
Moreover, in the first case, there is some out-of-band information that the contentType is a JSON string. I would strongly prefer the second approach since it accurately describes what is going on: packing an image into a JSON string.

sebastiankb:
in today's TD call we were thinking about the text/plain case which maybe also reflected by (simple) data schema types. The TD spec has also an example: https://www.w3.org/TR/wot-thing-description/#example-34-myilluminancesensor-with-mqtt-protocol-binding

sebastiankb:
from today's TD call:
- no merge today
- need also decision if this content is useful
- PR needs to be applied with the new render script

sebastiankb:
from today's TD call:
- decided to merge this PR
- however, we should mark the list "at risk"
- @danielpeintner will resolve conflict

danielpeintner:
Use https://github.com/w3c/wot-thing-description/pull/976 instead

PR #874, opened 2020-01-10
Address feedback from IANA

sebastiankb:
Today we received the feedback from IANA about the mediaType registration. There was the request to provide the concrete section for the RFC 8259 reference (-->section 12) and we should provide the absolute path (and not the relative path) to the section 7 in the Thing Description specification.

This minor changes are addressed in this PR.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/874.html" title="Last updated on Jan 10, 2020, 10:01 AM UTC (47b88ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/874/5725d82...47b88ab.html" title="Last updated on Jan 10, 2020, 10:01 AM UTC (47b88ab)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3989,7 +3989,7 @@ in a WoT Thing Description MUST accurately describe the <a>WoT Interface</a> of 
       <dt>Encoding considerations:</dt>
       <dd>See <a href="https://tools.ietf.org/html/rfc6839#section-3.1">RFC&nbsp;6839, section 3.1</a>.</dd>
       <dt>Security considerations:</dt>
-      <dd>See <a href="https://tools.ietf.org/html/rfc8259">RFC&nbsp;8259</a>.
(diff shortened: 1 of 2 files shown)


ektrah:
Background: IANA publishes the media type registration as a stand-alone document, so the whole section needs to be self-contained. For example, the media type registration doesn't have a section 7 itself, so instead of saying "see Section 7" it needs to say something like "see Section 7 of the published specification".

See [text/html](https://www.iana.org/assignments/media-types/text/html) and [application/senml+json](https://www.iana.org/assignments/media-types/application/senml+json) for examples.

sebastiankb:
There seem to be no objections. I'm going to merge this PR.

Issue #882, opened 2020-02-28
Giving credit to editors and contributors of the implementation report


egekorkan:
From the TD meeting of 28.02.2020, we have reached the following resolution regarding the implementation report:
> McCool and Ege can be editors of this document, and people contributed to testing can be described in acknowledgement.

Thus the implementation reports first contribution list should be changed and the suggested editors should be added. At the end of the document, acknowledgements should be changed to mention everyone who contributed with implementations and their descriptions.

takuki:
In my opinion, McCool and Ege **are** the editors of the [implementation report](https://github.com/w3c/wot-thing-description/blob/master/testing/report.html).

@egekorkan , do you have concrete suggestions who we can mention in the acknowledgements?

egekorkan:
I would acknowledge the contributed implementations, it could be the name of the people or the organization behind it. The list of implementations is available [here](https://github.com/w3c/wot-thing-description/tree/master/testing/inputs/implementations). 
The list of organizations would be:
- SmartThings
- Ercim
- Fujitsu
- Hitachi
- Huawei
- Intel (redundancy with editors)
- Oracle
- Panasonic
- Siemens
- TUM (redundancy with editors)

List of contributors would be:
- Michael Koster
- Dave Raggett
- Ryuichi Matsukura, couldn't track others from Fujitsu in the implementation report but Takuki Kamiya?
- Kunihiko Toumura
- Matthias Kovatsch
- Michael Lagally
- Takeshi Yamada, Toru Kawaguchi
- Sebastian Käbisch, Daniel Peintner, Victor Charpenay

mryuichi:
@egekorkan, could you change Fujitsu's contributors to 
Takahisa Suzuki, Takeshi Sano, Ryuichi Matsukura?

takuki:
Resolved. It was suggested to close the issue in 3/6 TD call.

PR #884, opened 2020-03-06
Giving credit to editors and contributors of the implementation report.

takuki:
This is for issue #882 .


takuki:
I fixed @egekorkan  affiliation. (TUM → SIemens / TUM) as suggested in TD telecon.

PR #914, opened 2020-06-17
fix: remove wrong "," in example 17 TD snippet

danielpeintner:
...knowing that many TD examples are not valid JSON files as a "whole" this PR removes a wrong "," in the *middle* of a TD snippet

### DIFF
--- a/index.html
+++ b/index.html
@@ -2663,7 +2663,7 @@ In summary, this requires the following:
                     },
                     "duration": {"type": "number"}
                 },
-                "required": ["to","duration"],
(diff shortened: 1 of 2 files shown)


sebastiankb:
we can also fix this in the REC document. Doing this we should collect more bugs first and then do a fix once. It is recommended to create another issue for collecting all kind of editorial bugs for the TD REC. @danielpeintner will create this issue

sebastiankb:
In today's TD call we decided to merge this PR.

Issue #915, opened 2020-06-18
Vocabulary term "scheme" not defined when used inside securityDefinitions


nikosft:
When I try to parse a TD that includes `securityDefinitions` (using for example jsonld.js) I receive an error for every term included in the `securityDefinitions` collection. For example even this small snippet will fail with `Error: The property "scheme" in the input was not defined in the context`:

```
{
  "@context": "https://www.w3.org/2019/wot/td/v1",
  "id": "urn:dev:ops:32473-WoTLamp-1234",
  "title": "MyLampThing",
  "securityDefinitions": {
    "basic_sc": {
      "scheme": "basic",
      "in": "header"
    }
  },
  "security": [
    "basic_sc"
  ]
}
```

My suspicion is that this is happening because in https://www.w3.org/2019/wot/td/v1 all terms related to  `securityDefinitions` are defined in the `security` collection. As a result, the following (odd) snippet is parsed without errors:

```
{
  "@context": "https://www.w3.org/2019/wot/td/v1",
  "id": "urn:dev:ops:32473-WoTLamp-1234",
  "title": "MyLampThing",
  "securityDefinitions": {
    "basic_sc": {}
  },
  "security": [
    "basic_sc",
    {
      "scheme": "basic",
      "in": "header"
    }
  ]
}
```

[...]

PR #937, opened 2020-07-29
WIP: swap securtiy and securityDefinition in context file 

sebastiankb:
This PR will fix the bug #915.

There some open questions:

- [ ] Why do we have two context files, td-context-1.1.jsonld and td-context.jsonld?
- [ ] There seems no definition for `td:securityDefinitions`, nor in validation or security ttl file:

image

@vcharpenay may I ask you to check this.

### DIFF
--- a/context/td-context-1.1.jsonld
+++ b/context/td-context-1.1.jsonld
@@ -104,7 +104,7 @@
       "@type": "@id",
       "@container": "@index"
     },
-    "security": {
+    "securityDefinitions": {
       "@id": "td:hasSecurityConfiguration",
       "@type": "@id",
       "@container": "@set",
@@ -168,7 +168,7 @@
         "oauth2": "wotsec:OAuth2SecurityScheme"
       }
     },
-    "securityDefinitions": {
+    "security": {
       "@id": "td:securityDefinitions",
       "@type": "@id",
       "@container": "@index"


sebastiankb:
From today's TD telko we should check it this is just a type: td:securityDefinitions --> td:security ? 

@sebastiankb will check this

danielpeintner:
I believe that the file `context/td-context-1.1.jsonld` is a merge from several other files like `context/td-context.jsonld`

sebastiankb:
from today's TD call:
- it turned out that the context deinition is correct, however, it is uncompleted
- @vcharpenay will have look and will complete the security / securityDefinition in the contex
- there is a related PR: #945

mmccool:
I see there were a lot of changes to these definitions as part of the render script (PR https://github.com/w3c/wot-thing-description/pull/951) as well.  Do we have to rebase this now, and/or has the issue been dealt with already?

sebastiankb:
from today's TD call:
- not ready to merge
- @vcharpenay how much time do you need to fix this issue (if possible)?

mmccool:
So I'm trying to understand what's going on here.  It looks like the definitions for "security" and "securityDefinitions" got munged together somehow.  Right now, in the rendered spec, there is only "security", and its description is correct, but the type is wrong, it should be array of strings (or one string).  And "securityDefinitions" is missing, but should be "Map of SecurityScheme".  The thing is... the *published* spec at https://www.w3.org/TR/wot-thing-description/#thing is correct.  So something broke it along the way.

BTW I wanted to based my "additionalSchemas" proposal for "additionalResults" on the setup for "securityDefinitions" but...

mmccool:
BTW I don't think the PR fixes the problem, the definitions under "ontology" and elsewhere seem to have gone AWOL also.

sebastiankb:
From today's TD call:
- we need a round trip for security definitions
- @vcharpenay will provide a PR

sebastiankb:
form today's TD call: @vcharpenay is this PR still relevant or can we close it?

sebastiankb:
from today's TD call, PR is not relevant anymore.

PR #938, opened 2020-07-29
move Thing Model (former TDT) to a main section + formalize Thing Model + introduce placeholder

sebastiankb:
Start to provide formal definition of Thing Model (former TDT) and its opportunists in TD spec.  

- [x] move Thing Model to a main section

- [x] formalize Thing Model

- [x] introduce placeholder

- [x] optional / manditory

- [x] illustration for relation Thing Model and Thing Description instance  

- [ ] extend / import feature   --> will be addressed in the next draft


sebastiankb:
added TDT definition from https://github.com/w3c/wot-architecture/issues/454#issuecomment-604272883

index.template.html
```diff
@@ -3916,6 +3925,115 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+    <li> develop common applications across devices from different manufacturers that share a common <a>Thing</a> model.</li>
+    <li> combining multiple models into a <a>Thing</a>.
+    </ul>
+    </p>
+    <p>
+    The Thing Model is a logical description of the interface and possible interaction with <a>Thing's</a>
+    <a>Properties</a>, <a>Actions</a>, and <a>Events</a>, however it does not contain <a>Thing</a> instance-specific
+    information, such as security information and concrete protocol usage, or even a serial number and GPS location, .
+    </p><p>
+    Since a Thing Model MAY not contain complete <a>Protocol Binding</a> information such as endpoint address and/or
```
mlagally, line 3951: Please check RFC 2119 for the usage of "MAY". Perhaps "MUST" would be appropriate here?

index.template.html
```diff
@@ -3916,6 +3925,115 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+    
+    
+    
+    <p>
+     The same Thing Model can be implemented by <a>Things</a> from multiple vendors,
+    a <a>Thing</a> can implement multiple Thing Model, define additional metadata
+    (vendor, location, security) and define bindings to concrete protocols.
+    To avoid conflicts between properties, actions and events from different Thing Model
+    that are combined into a common <a>Thing</a>, all these identifiers must be unique within a <a>Thing</a>.
+    </p><p>
```
mlagally, line 3995: Uniqueness across different providers of thing models is difficult to achieve. IMHO we need a namespace concept.

index.template.html
```diff
@@ -3916,6 +3925,115 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+    A common Thing Description Template for a class of devices enables writing applications across vendors and
+    creates a more attractive market for application developers.
+    A concrete Thing Description can implement multiple Thing Model and
+    thus can aggregate function blocks into a combined device.
+    </p><p>
+    The business models of cloud vendors are typically built on managing thousands of identical devices.
+    All devices with the same Thing Description Template can be managed in the same way by cloud applications.
```
mlagally, line 4002: ... cloud services and applications.

index.template.html
sebastiankb: from RFC 2119: MAY   This word, or the adjective "OPTIONAL", mean that an item is
   truly optional. 

I think this is the right way, because the implementer should be aware that a thing model gets along without or partly without this vocabulary. 

Maybe what would be also possible "SHOULD NOT". What do you think?

index.template.html
sebastiankb: good point. The idea would be to use the "$ref" mechanism from JSON Schema. I will check how namespace conflicts are handled.

index.template.html
sebastiankb: will be updated in my next commit.

index.template.html
sebastiankb: In my new commit I used the term "SHOULD NOT".

index.template.html
```diff
@@ -3916,6 +3925,191 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+      
+  <p class="note">
+    The following section has its origin in [[wot-thing-description]], Annex C. Here Thing Description 
+    Template is renamed to <a>Thing Model</a>, but keeps the same intention.
+    For this version of the specification, <a>Thing Model</a> and its model features (e.g., extendions, obligations, placeholder) are formal 
+    introduced.
+  </p>
+
+    <p>
+    A <em>Thing Model</em> is a special kind of a <a>Thing Description</a> with less restriction as
```
mmccool, line 3939: I think calling it a "special kind" of Thing Description is not a good way to put this.   I would say it is "related to" a Thing Description, but it is NOT a Thing Description and so can't be a special kind of one.  Note as well this also implies we should not use the same content type (td+jsonld) but should define a new one (tm+jsonld would be good).

index.template.html
```diff
@@ -3916,6 +3925,191 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+    Template is renamed to <a>Thing Model</a>, but keeps the same intention.
+    For this version of the specification, <a>Thing Model</a> and its model features (e.g., extendions, obligations, placeholder) are formal 
+    introduced.
+  </p>
+
+    <p>
+    A <em>Thing Model</em> is a special kind of a <a>Thing Description</a> with less restriction as 
+    driven in <a href="#sec-vocabulary-definition"></a> and <a href="#sec-td-serialization"></a>. Thing Model samples can 
+    also be seen as a subset of the <a>Thing Descriptions</a> information model, which may not contain, 
+    for example, concrete protocol and security information.
```
mmccool, line 3942: Probably we want to say "It may also contain URL templates rather than concrete URLs" although I guess we have to discuss this.

index.template.html
```diff
@@ -3916,6 +3925,191 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+    <li> simulate of devices/<a>Things</a> that have not yet been developed.</li>
+    <li> develop common applications across devices from different manufacturers that share a common <a>Thing</a> model.</li>
+    <li> combining multiple models into a <a>Thing</a>.
+    <li> implementation support of a concrete <a>Thing</a>.
+    </ul>
+    </p>
+    <p>
+    The Thing Model is a logical description of the interface and possible interaction with <a>Thing's</a>
+    <a>Properties</a>, <a>Actions</a>, and <a>Events</a>, however it does not contain <a>Thing</a> instance-specific
+    information, such as security information and concrete protocol usage, or even a serial number and GPS location.
```
mmccool, line 3957: I think TMs *should* be *allowed* to include security schemes.   Note that security schemes don't declare instance-specific information anyhow.   They *might*, however, have URLs (like token servers) that might need to be omitted or parameterized (with templates) although in a lot of cases these might also be given (a TM might easily define a set of instances all of whom share a specific token server).   Simplest thing here is just to remove the "for instance" phrase.  Actually, I also often want to specify the protocol so the best you can say here is "such as concrete fully-specified URLs"

index.template.html
```diff
@@ -3916,6 +3925,191 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+    <span class="rfc2119-assertion" id="td-model-identification">
+        <a>Thing Model</a> definitions MUST use the keyword <em>@type</em> at top level 
+        and a value of type string or array that equals or respectively contains <code>ThingModel</code>.
+    </span>
+    </p>
+
+    <p>
+        <span class="rfc2119-assertion" id="td-model-protocol-security-restriction">
+    Since a <a>Thing Model</a> SHOULD NOT contain complete <a>Protocol Binding</a> information such as endpoint address and/or
+    specific security mechanism.
```
mmccool, line 3971: Again, I DO want to specify the security schemes if they apply to the entire class of instances the model describes.

index.template.html
```diff
@@ -3916,6 +3925,191 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+    creates a more attractive market for application developers.
+    A concrete Thing Description can implement multiple Thing Model and
+    thus can aggregate function blocks into a combined device.
+    </p><p>
+    The business models of cloud vendors are typically built on managing thousands of identical devices.
+    All devices with the same Thing Description Template can be managed in the same way by cloud services and applications.
+    It is easy to create multiple simulated devices, if the interface and the instance are treated separately.
+    </p>
+    <p>
+        Since a Thing Description Template is a subset of the Thing Description in which some optional and mandatory <a>Vocabulary Terms</a>
```
mmccool, line 4082: Thing Description Template -> Thing Model.   Would be a good idea to do a search for "Template" to see if you've caught them all.

index.template.html
```diff
@@ -3916,6 +3925,191 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+    A concrete Thing Description can implement multiple Thing Model and
+    thus can aggregate function blocks into a combined device.
+    </p><p>
+    The business models of cloud vendors are typically built on managing thousands of identical devices.
+    All devices with the same Thing Description Template can be managed in the same way by cloud services and applications.
+    It is easy to create multiple simulated devices, if the interface and the instance are treated separately.
+    </p>
+    <p>
+        Since a Thing Description Template is a subset of the Thing Description in which some optional and mandatory <a>Vocabulary Terms</a> 
+        do not exist, however, it can be serialized in the same way and in the same formats as a Thing Description. Note that Thing
```
mmccool, line 4083: If items mandatory in a TD can be omitted in a TM, then a TM is not a TD and we should be careful not to call it a subset.   It will not pass validation as a TD.

mmccool (approved):
I think more work is needed on this section, in particular I think we need to allow TMs to include security sections if the same security applies to all instances.  However, we should make "security" optional instead of mandatory.   Also, since a TM is not in fact a TD (it will not validate as one) I think we should create a new contentType for it (maybe tm-jsonld).  HOWEVER I'm ok with merging this PR after creating issues to discuss those two points.

index.template.html
```diff
@@ -3916,6 +3925,191 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
 
   </section>
 
+  <section id="thing-templates" class="normative">
```
sebastiankb, line 3928: should be the id="thing-model"

index.template.html
sebastiankb: ...is a template (or class) for Thing Description...

index.template.html
```diff
@@ -3916,6 +3925,191 @@ <h5>Inferencing of Personally Identifiable Information Privacy Risk</h5>
...
+  <h1>Thing Model</h1>
+      
+  <p class="note">
+    The following section has its origin in [[wot-thing-description]], Annex C. Here Thing Description 
+    Template is renamed to <a>Thing Model</a>, but keeps the same intention.
+    For this version of the specification, <a>Thing Model</a> and its model features (e.g., extendions, obligations, placeholder) are formal 
+    introduced.
```
sebastiankb, line 3935: Mention that also a new content type is under discussion for Thing Model.

index.template.html
sebastiankb: we can add this information after this sentence

index.template.html
sebastiankb: I can also mention this.

index.template.html
sebastiankb: This depends on the use cases (digital twin use case vs. developer use case). Then let's go back to MAY NOT to cover different use cases. @mlagally is it also ok for you?

index.template.html
sebastiankb: ok

index.template.html
sebastiankb: we can add this information to make the different to the TD clear, especially in the context of TM processing.

sebastiankb:
from today's TD call:
- not ready to merge
- @sebastiankb will take the comments and improve this PR for the next meeting

index.template.html
sebastiankb: done

index.template.html
sebastiankb: done

index.template.html
sebastiankb: done

index.template.html
sebastiankb: done

index.template.html
sebastiankb: done

index.template.html
sebastiankb: done

index.template.html
sebastiankb: done

index.template.html
sebastiankb: done

sebastiankb:
from today's TD call:
- decided to going to be merged

Issue #940, opened 2020-08-05
Add optional proof section to TDs using ld-proofs.


mmccool:
Add an optional "proof" section following https://w3c-ccg.github.io/ld-proofs/ to ensure the integrity of TDs, as discussed in the security session in the last F2F.  This is the same mechanism used in DID documents.

mmccool:
We will also be including the "proofChain" section, as specified in ld-proofs.   This is important for directories and proxies, since if they modify the TD for some reason they have to update the signature, or if another entity also wants to sign it (eg the user or redistributor).   So actually this PR will add two new sections, an optional "proof" section and an optional "proofChain" section.   To check: it seems that only one of "proof" or "proofChain" should be used.   The first (most recent) element in the proofChain is the "current" proof.   Also note that "proof" can either be a single element or an array, and if more than one element it is considered a "proof set" where there are multiple proofs from different entities, and order is not relevant.   The ld-proofs vocabulary should be included in the base vocabulary for TD 1.1.  For TD 1.0, an extension context (https://www.w3.org/2018/credentials/v1, suggested prefix "cr") must be used.

mmccool:
A discussion of JWS and canonicalization requirements was added to wot-profiles recently: https://github.com/w3c/wot-profile/issues/55
However, I'm still thinking the discussion should be under the TD spec.
Summary (see Arch minutes [here](https://www.w3.org/2020/11/19-wot-arch-minutes.html)) people want to proceed with JWS but a TD canonicalization mechanism is needed.   We did find a JSON canonicalization reference and the idea is to add a few other requirements for TDs (default values, etc) but we probably also need to deal with some JSON-LD issues (e.g. use of prefixes as opposed to URL expansions, etc) which may affect systems that process RDF internally (eg SPARQL-based directories).

farshidtz:
Would it be possible to sign parts of the object and list the signed fields to allow partial verification?

e.g. proof.fields = ["id", "properties", "links[0]"]

This is going to be very useful, as other entities can annotate a TD while keeping the original TD verifiable. Examples:
- A proxy may add additional security or forms
- Directory may enrich with geospatial or registration attributes

mmccool:
Rebooting.  Want to define a signing mechanism.  Following are some notes on requirements arising from security TF call on May 3, 2020.
1. Directories returning signed TDs are basically using object security, and we should set up the TDs that way.  Figuring this out will be useful for other Things that want to use object security.
[...]

PR #943, opened 2020-08-11
WIP: Add proof and proofChain sections

mmccool:
For signing TDs, add proof and proofChain sections consistent with Linked Data Signatures (https://w3c-ccg.github.io/ld-proofs/#linked-data-signatures).   WIP, need to discuss:
1. How to declare objects defined in a different spec.  we don't want to redundantly repeat the definition.
2. How to handle context; context file not yet updated
3. How to achieve backward compatibility... explicit extension context needed in some cases?
4. Need to definite normalized serialization order for signatures.  Need to be careful about items where order is defined as not mattering, but does, eg actions, properties, etc.  Also need to revisit whether order matters for forms.  Propose introducing a "Set" abstract data type to make this clearer (syntactically it would still be an array, though).

Satisfies https://github.com/w3c/wot-thing-description/issues/940

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/943.html" title="Last updated on Feb 24, 2021, 3:23 PM UTC (8b587cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/943/58f543f...mmccool:8b587cb.html" title="Last updated on Feb 24, 2021, 3:23 PM UTC (8b587cb)">Diff</a>


sebastiankb:
lets discuss this in tomorrow's TD meeting

mmccool:
Note the LD-Proofs spec allows some fields to be open-ended, and I think for interoperability we want to support only a finite set of options.

sebastiankb:
from today's TD call:
- not ready to merge
- topic to discuss with the JSON-LD group during TPAC

sebastiankb:
from today's TD call:
- going to merge this PR
- mark this as "at risk"
- @mmccool will resolve the conflicts and will merge this PR

mmccool:
Ugh, this is going to be difficult to update to use the new render script since the original PR just modified index.template.html directly and did not have any ontology definitions... since one of the things I wanted to resolve was grabbing those definitions from the DID/VC WG.   That will have to wait until we meet with them to sort out so for now I think we have to defer this until after the FPWD.

iherman:
Note: we must be careful that the LD Signature document is not (yet) normative. There are initial discussions to, eventually, set up a Working Group whose goal would be to provide a standard in that direction (including a canonicalization algorithm for linked data, which is the hard nut to crack for such a WG), but there is no specific plan, charter proposal, etc, yet for this.

Here is a place where the evolution can be followed: https://github.com/w3c/strategy/issues/116

sebastiankb:
From today's TD call:
- this PR is still valid, however, the PR is not fixable
- best way is to create a new PR and close this one
- @mmccool will take care on this

mmccool:
Looks like there has been no further progress on a canonical form for JSON-LD so I think we should proceed assuming such will not be done in time.  Also, as pointed out above, LD-PROOFS is not normative so I can't just cite it as a definition like in my original PR.  Rather I will have to include all the necessary definitions in the TD document itself so it is self-contained.

sebastiankb:
from today's TD call:
- close this PR since the base line is outdated
- @mmccool creates a new PR (will be based on the issue #940)

Issue #901, opened 2020-05-14
Clarifying use of multiple security schemes in the security term


egekorkan:
The specification allows for the `security` to be an array of strings and example 11 shows this. However, there is no evidence of this feature being used in the testing inputs and also the explanation in the specification is not very clear. I am guessing that this pattern was taken from OpenAPI but we should explain this better and experiment in the Plugfests. 

Also, what happens if there are conflicting security schemes described in the array?

egekorkan:
@mmccool any inputs on this?

farshidtz:
In OpenAPI, the security schemes array is applied in an OR fashion (see [step 2](https://swagger.io/docs/specification/authentication/)). But TD applies them in an AND fashion.:

> Here is a more complex example: a TD snippet showing digest authentication on a proxy combined with bearer token authentication on the Thing. 
https://www.w3.org/TR/wot-thing-description/#security-digest-example

> In other words, the use of different security configurations within multiple forms provides a way to combine security mechanisms in an "OR" fashion. In contrast, putting multiple security configurations in the same security member combines them in an "AND" fashion, since in that case they would all need to be satisfied to allow activation of the Interaction Affordance.
https://www.w3.org/TR/wot-thing-description/#example-14


Apart from conflicting schemes (e.g. basic and various types of bearer tokens passed as Authorization header), the specification leads to very large TDs when each endpoint supports few security schemes. For example, an API secured with OAuth2 may support few flows, as offered by the authorization server. The only way to specify those is to repeat each form instance for every supported flow:
```json
{
    ...
    "securityDefinitions": {
        "oauth2_code": {
            "scheme": "oauth2",
            "flow": "code",
            ...
        },
        "oauth2_client": {
            "scheme": "oauth2",
            "flow": "clientCredentials",
            ...
        },
        "oauth2_password": {
            "scheme": "oauth2",
            "flow": "password",
            ...
        }
    },
    "security": ["oauth2_code"],
    ...
    "properties": {
        "status": {
            ...
            "forms": [{
                "href": "https://scopes.example.com/status"
            },{
                "href": "https://scopes.example.com/status",
                "security": ["oauth2_client"]
            },{
                "href": "https://scopes.example.com/status",
                "security": ["oauth2_password"]
            }]
        }
    },
    ...
}
```

egekorkan:
[...]

PR #944, opened 2020-08-11
Combination security scheme

mmccool:
Addresses: https://github.com/w3c/wot-thing-description/issues/901

A security scheme that allows other security scheme definitions to be combined using either "any" or "all" combinations.   The "anyOf" and  "allOf" elements are meant to be similar to the corresponding data schema definitions (which is why I used these terms instead of the "and" and "or" terms discussed in the linked issue).

Things to discuss:
1. Maybe we should use "oneOf" instead of "anyOf"? (edit: decided "oneOf" is better)
2. We could alternatively define separate schemes for "anyOf" and "allOf", but then some other keyword is needed to list the arguments. (edit: can have one scheme, but if made default, can get a simple syntax in the end)
3. Ontology files and JSON Schema not yet updated, we should discuss first; once we have reached consensus on the design, these details can be filled in. (edit: now done)
4. The restriction that only one of "anyOf" or "allOf" must be used is not easy to describe in the table. Basically exactly one of these is mandatory. (edit: restriction added to JSON schema, but not sure how to describe in ontology or shapes; marked with FIXME, needs to be discussed)

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/944.html" title="Last updated on Sep 30, 2020, 11:20 PM UTC (fc67c88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/944/ac6b16e...mmccool:fc67c88.html" title="Last updated on Sep 30, 2020, 11:20 PM UTC (fc67c88)">Diff</a>


farshidtz:
I vote for "oneOf" ([exactly one](http://json-schema.org/understanding-json-schema/reference/combining.html#oneof)) instead of "anyOf" ([one or more](http://json-schema.org/understanding-json-schema/reference/combining.html#anyof)). A consumer program which can work with more than one, may choose to do so and override inputs (e.g. header) and possibly cause conflicts.

farshidtz:
As mentioned [here](https://github.com/w3c/wot-thing-description/issues/901#issuecomment-661347764), having the combination scheme, I suggest deprecating the Array type for `security` or at least making it NOT RECOMMENDED. 

Having different ways to define AND combination (or anything for that matter) reduces usability and makes it harder for applications to adopt Thing Descriptions.

mmccool:
The text really needs an example.  Farshid has one [here](https://github.com/w3c/wot-discovery/blob/71612e81f987ba43f6943f9fd542d15492bcefdb/directory.td.json#L38), which I extract below:
```
        "combination": {
            "scheme": "combination",
            "oneOf": [
                "oauth2_code",
                "oauth2_client",
                "oauth2_device"
            ]
        }
```
Note this uses the suggested "oneOf" syntax (which I agree with and intend to change). 
As an additional idea, if we allow the ["inline security definitions"](link) to also be used for "oneOf" and "allOf", then we could dispense with the names and securityDefinition and do things like:
```
       "security": {
            "scheme": "combination",
            "oneOf": [
                 {
                    "scheme": "oauth2",
                    "flow": "code",
                    ...
                 },
                 {
                     "scheme": "oauth2",
                     "flow": "client",
                     ...
                 },
                 {
                     "scheme": "oauth2",
                     "flow": "device",
                     ...
                 }
            ]
        }
```

mmccool:
So I had another thought about how to make this more compact, related to a point raised in the Security TF meeting today about "why don't we just have oneOf/allOf by themselves".  I realized this IS possible if we put them at the same level as the "scheme" element and get rid of the special pseudo-scheme "combination".   Basically, in a SecurityScheme object you would use one of "scheme", "oneOf", or "allOf".

Here are the examples above rewritten to use that approach:

mmccool:
@farshidtz also proposed deprecating the current syntax where an array in "security" means an "allOf" combination.  There are two reasons to do this: 1. it is redundant with the new "allOf" syntax 2. it conflicts with the usage in OpenAPI, which uses "OR" (oneOf) semantics.  To my mind 2 is the bigger problem.  Note that "deprecation" means in the 1.1 spec we would say "NOT RECOMMENDED" and would only remove it in TD 2.0 (since it would break backward compatibility).   However, with the compact proposed syntax, there is a simple equivalence:
```
     "security": ["a", "b", "c"]
```
would be exactly equivalent to
```
     "security": { "allOf": ["a", "b", "c"] }
``

mmccool:
Upon reflection, I realized the proposed "compact" syntax w/o "scheme" would cause a compatibility problem in the 1.x version of the spec since "scheme" is mandatory.   Omitting it would break TD 1.x processors that expect it.  So what I propose doing is introducing this in several steps:
1. Adopt a "combination" security scheme as in this PR.
2. In TD 1.1, make this the recommended way to do combinations and mark the array syntax as deprecated (but still legal)
3. In TD 2.0, make "scheme" optional, or more precisely, give it the default value of "combination".  Then the above syntax works: if "scheme" is omitted, then you need one of "allOf" or "oneOf" instead.
4. In TD 2.0, also remove the array syntax for "allOf" combinations; however, the compact syntax is available instead.

takuki:
In Sept 2 TD call, it was suggested to add certain constraints such as prohibiting nested CombinationSecurityScheme, for example.

It was also pointed out that we need to ask for some implementation experience first, especially on Consumer side.

In the coming PlugFest, we should ask implementers to experiment with Combination security scheme.

There is a mashup of microservices use case on producer (or proxy) side for Combination security scheme.

Consumer can use a security scheme from "oneOf" from the Combination security scheme.
This expectation should be described in a note in this PR.
(Note currently, there is a limitation in NodeWoT implementation.)

mmccool:
Marked as draft because I have to add one more assertion to prevent recursive use (to simplify testing).
Also this can't be merged at the moment due to render script updates - rebasing and possibly manual conflict resolution are still needed.

sebastiankb:
from today's TD call:
- no merge today, lets check this next week again
- same situation with #945, implementations needed

sebastiankb:
from today's TD call:
- decided to merge this PR
- however, this new feature should be marked "at risk"
- there are conflicts, have to be fixed. 
- add an editorial note about a specific use case
- @mmccool will address those points and will merge this PR

Issue #957, opened 2020-08-20
Unsubscribing and unobserving are not in default values


egekorkan:
The default values section do not say that `unsubscribeevent` is a default value in the `op` of an Event Affordance but `subscribeevent` is. If the defaults are filled, this results in a TD that cannot be used by a Scripting API implementation since when the `op` value is existing, its contents cannot be changed and there is no form to unsubscribe from the event.

We can also look at it from another perspective and add an assertion saying that if there is a form in the forms array with `subscribeevent` there should be a form (or the same form) with the `op` value of `unsubscribeevent`. We can say the same for observable properties, i.e. if there is `op` value of `observeproperty`, there should be also `unobserveproperty` somewhere.

**Warning:** Even though it is a small change, it is a breaking change from my point of view.

danielpeintner:
Nice finding!

> We can also look at it from another perspective and add an assertion saying that if there is a form in the forms array with `subscribeevent` there should be a form (or the same form) with the `op` value of `unsubscribeevent`. We can say the same for observable properties, i.e. if there is `op` value of `observeproperty`, there should be also `unobserveproperty` somewhere.

I think the same applies vice-versa also meaning that if  `unsubscribeevent` is present there should be `subscribeevent` also.

> **Warning:** Even though it is a small change, it is a breaking change from my point of view

Personally I do not see it as _dangerous_ and I assume we can add an explanation saying that if _one_ is present the counterpart is present also...

zolkis:
I think it's more like a "fixing" change than a "breaking" change.

A script could rightfully assume that if it can subscribe to something, it can also unsubscribe from it. This should be regarded as _one_ functionality with two aspects.

takuki:
> Another question is what to do with "orphan subscriptions", e.g. a subscription record at the server that contains an address that is offline ATM or has the address changed. I guess this is impl optimization since it's protocol-dependent.

If a protocol supports heartbeat to detect connection loss, that protocol may be able to consider the subscription dead, and no explicit unsubscription may be necessary. But I agree it is a good practice to have both subscription and unsubscription at the same time.

sebastiankb:
Regarding @takuki comment, I wonder about the scenarios like streaming data via websocket and clients can simply listen to the stream. No explicit unsubscription interaction is required to stop the listing process (the client simple stop the listening).

egekorkan:
Same with longpoll :)

sebastiankb:
From today's TD call:
[...]

PR #960, opened 2020-09-15
docs: add unsubscribeevent to default list and add notes

danielpeintner:
The note is about best practices for unsubscribing

fixes https://github.com/w3c/wot-thing-description/issues/957

***
<a href="https://pr-preview.s3.amazonaws.com/danielpeintner/wot-thing-description/pull/960.html" title="Last updated on Oct 1, 2020, 8:36 AM UTC (5b7438e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/960/32ba69e...danielpeintner:5b7438e.html" title="Last updated on Oct 1, 2020, 8:36 AM UTC (5b7438e)">Diff</a>


danielpeintner:
Note: If we agree on the way forward I will provide updates to index.template.html also.

index.html
```diff
@@ -1218,7 +1218,9 @@ <h2>Core Vocabulary Definitions</h2>
 <code>unobserveproperty</code> or an  <a>Array</a> containing a combination of these terms.
     </span>
 </p>
-
+<p class="note">
+    It is considered to be good practice that each <code>observeproperty</code> has a corresponding <code>unobserveproperty</code> unless the protocol supports implicit unsubscription mechanisms (e.g., heartbeat to detect connection loss).
+</p>
```
relu91, line 1223: I'd also add this sentence: *Implementations might/should (?) use the "observerproperty" form as the default for "unobserverproperty" WoT interaction*. I think we can spare a little bit of redundancy here. For example:
```json
{
  "op": "observeproperty",
  "href": "mqtt://wot.example.com/level"
}
```
An implementation might infer that to *unobserveproperty* it needs to *unsubscribe* from `mqtt://wot.example.com/levelevent` topic.

relu91 (approved):
It looks good. I have only a single comment to open a discussion.

zolkis:
Question: how can someone find the `unsubscribe` Form that matches its `subscribe` Form?

danielpeintner:
> Question: how can someone find the `unsubscribe` Form that matches its `subscribe` Form?

It needs to be part of the same event/property (i.e. `"illuminance"` event)

```
    "events": {
        "illuminance": {
            "data":{"type": "integer"},
            "forms": [
                {
                    "href": "mqtt://192.168.1.187:1883/illuminance",
                    "contentType" : "text/plain",
                    "op" : "subscribeevent"
                },
                {
                    "href": "mqtt://192.168.1.187:1883/illuminanceUnsub",
                    "contentType" : "text/plain",
                    "op" : "unsubscribeevent"
                }
            ]
        }
    } 
```

Of course it is also possible that *one* form/href deals with both ops: `"op" : ["subscribeevent", "unsubscribeevent"]`  but it gets a bit trickier and you may need to know the current state...

zolkis:
Of course they need to be part of the same interaction, but what if there are multiple protocols supported.
So we could take a look at the `subscribe` Form, and select an `unsubscribe` Form that has same scheme in `href`, or if multiple are found, then if there is one with also the same `host` and then `port`?

egekorkan:
In a realistic scenario, the same protocol interface would handle the sub and unsub but TD allows that the protocols do it. I would not try to add a protocol equality check since if we think in detail:
- `subprotocol` must be same/usable
- `security` must be same/usable
- `href` must be _similar_, i.e. protocol, host (domain, local network, ...), port same/usable
- `contentType` same/usable, what if the implementation cannot parse XML that is needed only for unsubscription

There can be other points I miss as well

zolkis:
We need a better algorithm for `unsubscribe` in Scripting than we have now, that is why I am asking.
I will factor out an algorithm for this, then it could be checked/tested/updated separately.

sebastiankb:
from today's TD call:
- we call for more feedback on this discussion
- decide to merge in the next TD call then

sebastiankb:
issue that is also related: #968

sebastiankb:
from today's TD call:
- we decided to spend one more week for commenting
- there is a new issue from @zolkis #968 that can be common be discussed

sebastiankb (approved):

sebastiankb:
From today's TD call:
- we will merge this PR
- however, the index.template.html has to be edited, @danielpeintner will work on this

danielpeintner:
Note: this PR contains the necessary fixes in index.template.html and validation/td-validation.ttl and no update (expect aligning with master) to the other files like index.html.
Hence *someone* needs to run render script ... which contains many more changes due to other updates

danielpeintner:
Close since a *fresh* PR https://github.com/w3c/wot-thing-description/pull/975 has been used instead

Issue #941, opened 2020-08-05
Evaluate geolocation terms in TD model (input from Profile TF)


sebastiankb:
There are new vocabulary proposed in the latest Profile draft (also see https://github.com/w3c/wot-profile/issues/17). Which should evaluate how those should be covered in the core model. Which should also clarify which existing ontologies can be re-used for this kind of vocabulary. 

![image](https://user-images.githubusercontent.com/13832739/89497708-2eab7200-d7bd-11ea-9776-0bced248d1cd.png)

[See](https://www.science20.com/news_articles/what_happens_bridge_when_one_side_uses_mediterranean_sea_level_and_another_north_sea-121600) for a justification why loc_altitude is problematic.

sebastiankb:
for the geo based terms we should check GeoSPARQL https://www.ogc.org/standards/geosparql which is also used by SSN.

sebastiankb:
here is another source https://www.w3.org/community/geosemweb/

sebastiankb:
another (prominent) ontology https://lov.linkeddata.es/dataset/lov/vocabs/geo

sebastiankb:
There is a discussion in the Use Case TF about different geo location standards: https://github.com/w3c/wot-thing-description/pull/927

A promising one seems to be https://www.w3.org/TR/geolocation-API/

sebastiankb:
In today's TD call we discuss the option to define a geo container like `geoLocation` that embeds all geo-location based terms.

benfrancis:
> * we should also check https://schema.org/GeoCoordinates

E.g.

```
{
  "@context": [
        "https://www.w3.org/2019/wot/td/v1",
        { "s": "http://schema.org" }
    ],
    "id": "urn:dev:ops:32473-WoTLamp-1234",
    "s:longitude": "40.75",
    "s:latitude": "73.98",
    "elevation": "15",
    "s:manufacturer": {"name": "Acme"},
    "title": "MyLampThing",
    ...
}
```

egekorkan:
I would also be for using the already existing vocabularies in schema.org

takuki:
In [Spatial Data on the Web Best Practices](https://www.w3.org/TR/sdw-bp/) (2017), there is a Best Practice provided for "State how coordinate values are encoded" ([Best Practice 8](https://www.w3.org/TR/sdw-bp/#bp-crs)).

[...]

PR #961, opened 2020-09-15
geo location metadata

sebastiankb:
request from #941

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -3173,7 +3173,14 @@ On the <a>Thing</a> side, <a>Thing</a> is expected to return readable (non <code
     </p>
 
     <p>
-      As an example, the TD snippet given below extends the version information container
+    Next subsections shows some sample usage of different kind of ontologies in Thing Descriptions.
+    </p>
+
+    <section id="semantic-annotations-example-version-units">
+    <h3>Example I: Version and Unit Annotations</h3>
+
+    <p>
+      The TD snippet given below extends the version information container
       by adding version numbers for the hardware and firmware of the <a>Thing</a>, and uses
       values from external <a>Vocabularies</a> for the <a>Thing</a> and for the
       data schema unit: <a href="https://w3id.org/saref">SAREF</a>,
@@ -3214,6 +3221,10 @@ On the <a>Thing</a> side, <a>Thing</a> is expected to return readable (non <code
     ...
 }
 </pre>
+
+</section>
+<section id="semantic-annotations-example-state">
+    <h3>Example II: State Annotations</h3>
 
     <p>
         In many cases, <a>TD Context Extensions</a> may be used to annotate pieces of a data schema,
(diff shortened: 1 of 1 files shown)


sebastiankb:
from today's TD call:
- @mmccool samples shall relevant to concrete use case
- we shall discover the findings from the next PlugFest
- proposal is to merge the PR for now and make them available for the next draft, however, examples will be modified based on the findings from PlugFest
- there will be editorial note that points out the WIP status --> @sebastiankb 
- the second update shall be updated to a property that return location based data --> @sebastiankb

index.template.html
```diff
@@ -3282,6 +3293,73 @@ <h3>Semantic Annotations</h3>
...
+        The TD snippet below uses <code>lat</code> and <code>long</code> from the [[WGS84]] ontology to 
+        provide latitude and longitude metadata at Thing's top level. 
+    </p>
+
+    <pre class="example" id="saref-state-annotation-example">
+{
+    "@context": [
+        "https://www.w3.org/2019/wot/td/v1",
+        {
+            "wgs84": "http://www.w3.org/2003/01/geo/wgs84_pos#"
```
vcharpenay, line 3317: the usual prefix for WGS84 is `geo` (see [prefix.cc](https://prefix.cc/geo))

index.template.html
```diff
@@ -3282,6 +3293,73 @@ <h3>Semantic Annotations</h3>
...
+{
+    "@context": [
+        "https://www.w3.org/2019/wot/td/v1",
+        {
+            "iot": "http://schema.org#"
+        }
+    ],
+    ...
+    "properties": {
+        "status": {
```
vcharpenay, line 3347: I guess the example intends to show how e.g. to interpret

```
{
  "long": 40.75,
  "lat": 73.98,
  "height": 15
}
```

as

```
{
  "@context": "http://schema.org/",
  "schema:longitude": 40.75,
  "schema:latitude": 73.98,
  "schema:elevation": 15
}
```
.

I provide an example for this in the [JSON Schema in RDF](https://www.w3.org/2019/wot/json-schema#defining-a-json-ld-context-for-data-instances) vocabulary, which simply consists in linking to a JSON-LD context from the TD, as follows

```
{
  "properties": {
    "position": {
      "jsonld:context": {
        "schema": "http://schema.org/"
        "long": "schema:longitude",
        "lat": "schema:latitude",
        "height": "schema:elevation"
      },
      "type": "object",
      "properties": {
        "long": { "type": "number" },
        "lat": { "type": "number" },
        "height": { "type": "number" }
      }
    }
  }
}
```

The value for `jsonld:context` is an _inline_ JSON-LD context that provides a mapping to the keys defined in the JSON schema.

index.template.html
vcharpenay: but it might be confusing for readers to use two different vocabularies (WGS84 and schema.org) for the same kind of information (long/lat). WGS84 is very small and entirely included in schema.org. It might be better to use schema.org only.

index.template.html
sebastiankb: Would it be also possible to use this example?

{
  "properties": {
    "position": {
      "type": "object",
      "properties": {
        "@type" : "schema:GeoCoordinates",
        "longitude": { "type": "number" },
        "latitude": { "type": "number" },
        "elevation": { "type": "number" }
      }
    }
  }
}

index.template.html
sebastiankb: I will change this to geo.

>It might be better to use schema.org only.

Thats right. The intention was to show different examples that uses different ontologies.

index.template.html
vcharpenay: as a further example, yes. I would keep the one with `jsonld:context`, though, because it corresponds to another requirement.

sebastiankb: Lets keep both approaches to show the different kind of options how semantic based annotations can be applied.

sebastiankb:
I updated the samples based on @vcharpenay feedback.

index.template.html
relu91: Shouldn't it be:
```js
{
  "properties": {
    "position": {
      "type": "object",
       "@type" : "schema:GeoCoordinates",
      "properties": {
        "longitude": { "type": "number" },
        "latitude": { "type": "number" },
        "elevation": { "type": "number" }
      }
    }
  }
}
```
?

vcharpenay: yes, my bad.

sebastiankb: thanks for the hint. I updated the example.

sebastiankb:
merge this PR and apply the new render script as next

PR #982, opened 2020-10-05
Added Change Log section.

takuki:
Describe changes made since W3C Recommendation 9 April 2020 version.


sebastiankb:
many thanks. I will generate the index.html

Issue #987, opened 2020-10-07
Review - Thing Description 1.1 version


danielpeintner:
Besides some obvious typos (see PR https://github.com/w3c/wot-thing-description/pull/986) please find below some more comments/questions...

### 4. Namespaces

section speaks that "associated JSON-LD context file is versioned such that every version has its own URI (v1, v1.1, v2, ...) "

Doesn't that mean we should have

... this specification is identified by the following IRI: 
`https://www.w3.org/2019/wot/td/v1.1`

Instead we use everywhere 
`https://www.w3.org/2019/wot/td/v1`

### 5.3.2.4 NumberSchema
and
### 5.3.2.5 IntegerSchema

both have the statement for multipleOf
"Only applicable for associated number or integer types"
I think this is not necessary since we are in a given type number or integer context already.

Moreover in NumberSchema I think the following statement is wrong
"The result of division of the numeric instance on the multipleOf value should be integer"
There is no need for being of type integer. It can be number also right?

### 6.2 Omitting Default Values

Some minor consistency issues in Example 3
* some terms are like this
`"security": "basic_sc",`
while others contains spaces after :
`"readOnly" : false`
`"writeOnly" : false`

Same applies to many more example 6, 7, 8, 9, ...
Maybe an overall _pretty-print_ should be done..

### 6.3.2 Human-Readable Metadata

"default language is defined by a value assigned to @language in the @context, and this, along with a script subtag if necessary, can be used to determine a base text direction"

I am not sure what _script subtag_ refers to. The same text is present twice in the document.
Do we have an example?

### 7.1 Semantic Annotations

"Next subsections shows" --> "Next subsections show" ?

### 7.1.3 Example III: Geolocation Annotations

"and can be relyed on different" --> "and can be relied on different" ?

sebastiankb:
>Doesn't that mean we should have ... this specification is identified by the following IRI:
> https://www.w3.org/2019/wot/td/v1.1

Good point. Alternative would be, that we go back to the https://www.w3.org/ns/td namespace which we used during the draft phase of TD 1.0

sebastiankb:
>both have the statement for multipleOf
>"Only applicable for associated number or integer types"
>I think this is not necessary since we are in a given type number or integer context already.

This text is coming from the ontology definition which has not specific context at that point:
https://w3c.github.io/wot-thing-description/ontology/jsonschema.html#multipleof

sebastiankb:
[...]

PR #989, opened 2020-10-14
update abstract + fix bug + #987

sebastiankb:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/989.html" title="Last updated on Oct 19, 2020, 8:11 PM UTC (c7c8cfb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/989/a37f060...c7c8cfb.html" title="Last updated on Oct 19, 2020, 8:11 PM UTC (c7c8cfb)">Diff</a>


sebastiankb:
update abstract about TD 1.1 and relation to TD 1.0. + fix bug in example 8

index.html
```diff
@@ -336,7 +336,7 @@
   <section id="abstract">
   <p>
   This document describes a formal model and a common representation 
-  for a Web of Things (WoT) Thing Description. 
+  for a Web of Things (WoT) Thing Description 1.1.
```
danielpeintner, line 339: Not sure if this change is actually necessary since the version information come later

index.html
```diff
@@ -2408,7 +2410,7 @@ <h3>Human-Readable Metadata</h3>
 <pre class="example">
 {
     "@context": "https://www.w3.org/2019/wot/td/v1",
-    "title": "MyThing",
+    "title": "MeinDing",
```
danielpeintner, line 2413: Do not quite understand why it is changed to German. Having said that, I think both is fine.

index.html
```diff
@@ -349,7 +349,9 @@
   A Thing Description instance can be hosted by the <a>Thing</a> itself or hosted
   externally when a <a>Thing</a> has resource restrictions (e.g., limited memory space)
   or when a Web of Things-compatible legacy device is retrofitted 
-  with a Thing Description.
+  with a Thing Description. This specification is a continuation and further development 
+  of the Thing Description 1.0 [[WOT-THING-DESCRIPTION]]. 
+  It has the claim that the Thing Description 1.1 is backward compatible to Thing Description 1.0.
```
danielpeintner, line 354: Maybe we can simplify it by saying:

Web of Things Thing Description 1.1 is backwards compatible with Thing Description 1.0 [[WOT-THING-DESCRIPTION]].

danielpeintner (approved):
I am generally OK with the PR (I did add some minor comments but I do not see issues with the current status).

Having said that, I think we still need to update the namespace from
https://www.w3.org/2019/wot/td/v1 
 to
https://www.w3.org/2019/wot/td/v1.1

Sorry, I have missed the call yesterday. Not sure if there was a decisions w.r.t. that.

index.html
sebastiankb: I think, it is important to point out that this document is about TD 1.1. This is also provided by JSON-LD 1.1 https://www.w3.org/TR/json-ld11/#abstract

index.html
sebastiankb: Keeping this in English would confuse reader since the setting is "@language" : "de"

index.html
sebastiankb: The text is replaced by: "This specification describes a superset of the features defined in Thing Description 1.0. Unless otherwise specified, documents created with version 1.0 of this specification remain compatible with Thing Description 1.1."

index.html
danielpeintner: I think you changed the **wrong** example.

You changed [Example 7](https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/989.html#example-7) and wanted to change [Example 8](https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/989.html#example-8).
Is this the case?

sebastiankb: thanks. last commit should be correct now

PR #993, opened 2020-10-19
introduce working namespaces in TD 1.1

sebastiankb:
Introduce working namespaces for drafts of TD 1.1

In TD 1.1 we are no longer able to use the namespaces from TD 1.0. Namespaces are used to identify the different TD versions. I suggest to use temporary work namespaces during the draft status of the document. We did this also at the time when the TD 1.0 was still a draft. For simplicity I would use the same temporary namespace we used for TD 1.0.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/993.html" title="Last updated on Oct 20, 2020, 12:26 PM UTC (8d2ea5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/993/456be07...8d2ea5b.html" title="Last updated on Oct 20, 2020, 12:26 PM UTC (8d2ea5b)">Diff</a>


sebastiankb:
In TD 1.1 we are no longer able to use the namespaces from TD 1.0. Namespaces are used to identify the different TD versions. I suggest to use temporary work namespaces during the draft status of the document. We did this also at the time when the TD 1.0 was still a draft. For simplicity I would use the same temporary namespace we used for TD 1.0.

index.html
```diff
@@ -694,17 +694,22 @@ <h2>Terminology</h2>
 <section class="normative">
   <h1>Namespaces</h1>
 
+  <p class="note">
+    The following introduced namespaces are working namespaces for the time the specification has the Draft status.
```
mlagally, line 698: Better call them out as "placeholders" and talk about "final namespaces" that will be used after CR.

We could call them "2020-tentative/wot/td/v11"

index.template.html
```diff
@@ -694,17 +694,22 @@ <h2>Terminology</h2>
 <section class="normative">
   <h1>Namespaces</h1>
 
+  <p class="note">
+    The following introduced namespaces are working namespaces for the time the specification has the Draft status.
```
mlagally, line 698: See previous comment.

danielpeintner (approved):
I am perfectly fine with as it is but I wonder whether it is even better to make it more explicit that these namespaces are temporary.

Using "http://www.w3.org/ns/td" does really convey this statement.

What about 
"https://www.w3.org/2019/wot/td/v1.1-snapshot" or
"https://www.w3.org/2019/wot/td/v1.1-draft" or

https://www.w3.org/2019/wot/td# -> https://www.w3.org/2019/wot-draft/td#
https://www.w3.org/2019/wot/json-schema# -> https://www.w3.org/2019/wot-draft/json-schema#

**The advantage of the current approach is that we do not need to decide *NOW* about the actual namespace year and so which makes it very flexible.**

sebastiankb:
@danielpeintner 

>What about
>"https://www.w3.org/2019/wot/td/v1.1-snapshot" or
>"https://www.w3.org/2019/wot/td/v1.1-draft" or

I think this pattern does not fulfill the W3C URI guidelines https://www.w3.org/2005/05/tr-versions

index.html
sebastiankb: good idea. Just updated the note

danielpeintner:
> I think this pattern does not fulfill the W3C URI guidelines https://www.w3.org/2005/05/tr-versions

Agree. The document you cited would use the following for editor drafts 
`/TR/<shortname>/ed/`

Having said, I am fine with the current form we used in the past also.

PR #1012, opened 2020-11-25
fix #1003

### LINKED ISSUE #1003: Default value for op in Property Affordance should be linked to read/writeOnly
Currently the spec says in https://w3c.github.io/wot-thing-description/#x5-4-default-value-definitions that the `op` gets the default value of `["readproperty","writeproperty"]`. However, this is true only in the case of a readOnly or writeOnly property. Thus, I propose that the text should be changed from:
> Array of string with the elements readproperty and writeproperty

to:

> Array of string with the elements readproperty and writeproperty when readOnly and writeOnly are set to false OR Array of string with the element readproperty if readOnly is set to true OR Array of string with the element writeproperty if writeOnly is set to true.

This does make it more complicated because it actually is. Once one tries to make a defaults generation algorithm, this becomes important. It would be relevant for scripting as well. 

This is coming from this PR in the Playground: https://github.com/thingweb/thingweb-playground/pull/156#issuecomment-727911019

egekorkan:
***
<a href="https://pr-preview.s3.amazonaws.com/egekorkan/wot-thing-description/pull/1012.html" title="Last updated on Nov 25, 2020, 10:50 PM UTC (6440709)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1012/b8e1f9e...egekorkan:6440709.html" title="Last updated on Nov 25, 2020, 10:50 PM UTC (6440709)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1919,7 +1919,14 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
               <a href="#dfn-array" class="internalDFN"
               data-link-type="dfn">Array</a> of <code>string</code>
               with the elements <code>readproperty</code> and
-              <code>writeproperty</code><br>
+              <code>writeproperty</code> when <code>readOnly</code> and <code>writeOnly</code> are set to <code>false</code> or
+              <a href="#dfn-array" class="internalDFN"
+              data-link-type="dfn">Array</a> of <code>string</code>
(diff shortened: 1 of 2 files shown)


sebastiankb:
In today's TD call we decided to merge this PR?

Issue #999, opened 2020-11-09
[ThingModel] Example of Oracle's device model (Festo)


mlagally:
As discussed during the vF2F, here's an example of the device model of the Festo plant as it is used in Oracle's IoT Cloud applications.

Device models can be exported and imported from/to the Oracle IoT Cloud. 
A thing description can be generated from a device model using a converter (dm2td). 
A device model can be generated from a thing description using another converter (td2dm).

Other examples can be provided on request.

```{
  "urn": "urn:com:siemens:wot:festo",
  "name": "Festo",
  "description": "Common Device Model for Festo live plant and simulator",
  "system": false,
  "draft": false,
  "created": 1591711126087,
  "createdAsString": "2020-06-09T13:58:46.087Z",
  "lastModified": 1591711126087,
  "lastModifiedAsString": "2020-06-09T13:58:46.087Z",
  "userLastModified": "IoTAppsDev_w3cpoc_FRA_INTERNAL_APPID",
  "attributes": [
    {
      "name": "PumpStatus",
      "type": "BOOLEAN",
      "writable": false
    },
    {
      "name": "Tank101MaximumLevelStatus",
      "type": "BOOLEAN",
      "writable": false
    },
    {
      "name": "Tank101MinimumLevelStatus",
      "type": "BOOLEAN",
      "writable": false
    },
    {
      "name": "Tank101OverflowStatus",
      "type": "BOOLEAN",
      "writable": false
    },
    {
      "name": "Tank102LevelValue",
      "type": "NUMBER",
      "writable": false
    },
    {
      "name": "Tank102OverflowStatus",
      "type": "BOOLEAN",
      "writable": false
    },
    {
      "name": "ValveStatus",
      "type": "BOOLEAN",
      "writable": false
    }
  ],
  "actions": [
    {
      "argType": "BOOLEAN",
      "name": "CloseValve"
    },
    {
      "alias": "OpenValve",
      "argType": "BOOLEAN",
      "description": "Opens the valve",
      "name": "OpenValve"
    },
    {
      "argType": "BOOLEAN",
      "name": "StartPump"
    },
    {
      "argType": "BOOLEAN",
      "name": "StopPump"
    }
  ],
  "formats": [
    {
      "urn": "urn:com:siemens:wot:festo:tank101overflow",
      "name": "Tank101Overflow",
      "description": "Tank101 Overflow",
      "type": "ALERT",
      "deviceModel": "urn:com:siemens:wot:festo",
      "value": {
[...]

Issue #168, opened 2018-07-19
import/extend functionality in TD


sebastiankb:
As original initiated by @Kazuo-Kajimoto and reactivated again in Bundang we should bring in an import / extend functionality in the TD:

Lets assume there is a master TD template providing all basic interactions and a huge set of (identical) Things implements these interactions. It would be beneficial to maintain only the master TD and all the Things would inherit the interactions. Thing’s TD only extends its individual information (e.g., id, base URI, semantics like locations, etc) or overwrite master TD information. Example:

![image](https://user-images.githubusercontent.com/13832739/42951321-3b1805b6-8b76-11e8-8436-b8e5dafea6fd.png)

Open questions:

- Security aspect: Can we trust a master TD?
- Who construct the "full" TD? Lets assume when a client request a TD. Does it get the full resolved TD or the client has to download the master TD separately to get the full TD?

vcharpenay:
There are two analogies I think of around this idea:
 - when a Node.js module imports another one, the corresponding source file is dynamically loaded in memory and evaluated, so that the program can transparently use it.
 - HTML does not have an explicit import tag. Instead, it is often the case that a template engine renders the full document before it is exposed to the client.

Neither requires a standard import mechanism (in Node.js, import is implemented in JavaScript, it is not part of the interpreter). I would therefore argue that having TDs extending a master TD should be up to WoT implementers. Only fully interpretable TDs should be exchanged between servients. The exchange overhead that comes along is probably not significant and even if it is, it could be solved at the protocol or serialization level (as with Protocol Buffers or schema-informed EXI).

However, for maintenance purposes, a TD could refer by a link to the master TD (or 'template') that was used to generate it, like:
```js
{
  "links": [
    { "href": "...", "rel": "td:extends" }
  ]
}
```
and the relation type `td:extends` could be part of the TD model. There also exists several well-defined terms that one may reuse for that purpose, like [`prov:wasDerivedFrom`](https://www.w3.org/TR/2013/REC-prov-o-20130430/#wasDerivedFrom) or [`dc:source`](http://dublincore.org/documents/dcmi-terms/#terms-source).

benfrancis:
@sebastiankb wrote:
> Who construct the "full" TD? Lets assume when a client request a TD. Does it get the full resolved TD or the client has to download the master TD separately to get the full TD?

[...]

PR #1024, opened 2020-12-23
Topics around Thing Model 

sebastiankb:
- PR addresses https://github.com/w3c/wot-thing-description/issues/999#issuecomment-733805049
- introduce new subsection in TM chapter

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1024.html" title="Last updated on Feb 17, 2021, 2:44 PM UTC (6a2562d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1024/e07c8fa...6a2562d.html" title="Last updated on Feb 17, 2021, 2:44 PM UTC (6a2562d)">Diff</a>


JKRhb:
Thanks for this additional information regarding the Thing Model concept! I hope it is okay if I ask a question regarding Thing Models under this PR.

One aspect I find particularly interesting about Thing Models is the possibility to let a Thing implement mulitple models and therefore support a modular system when it comes to creating devices and applications. With the current state of the draft and this PR, however, I wondered how to deal with Thing Models that define affordances with the same name. 

This could occur in (at least) two scenarios: Either when a device offers multiple instances of the same Thing Model (for example, when two temperature sensors for inner and outer temperature are connected to a controller) or when two similar Thing Models are implemented (for example, when using both a temperature and a humidity sensor which both provide a value, a value range, and information regarding the unit of measurement).

Are there any ideas yet on how to deal with such scenarios? Thanks in advance!

sebastiankb:
Very good question which we already discussed in one of our meetings. 

The current approach (rel=type) allows to point to an existing TM which a TD is an instance of. So, it is expected that the TM causes not any name clashes since a TD can only instance one TM in the same time (I just provided the definition into the PR). 

Your concerns goes more to a TM definition that may extend multiple other TMs which maybe have, e.g., the same properties name, right?  I think, there are multiple solutions: programming languages like C++ uses namespaces for context differentiation. Java allows only to extend a single class / model to avoid the mess. Most likely the cleanest way is the introduction of a model import / reference mechanism with prefix assignment.

Other ideas are welcome.

JKRhb:
Thank you for your quick answer!

> Your concerns goes more to a TM definition that may extend multiple other TMs which maybe have, e.g., the same properties name, right? I think, there are multiple solutions: programming languages like C++ uses namespaces for context differentiation. Java allows only to extend a single class / model to avoid the mess. Most likely the cleanest way is the introduction of a model import / reference mechanism with prefix assignment.

I think the possibility to import other TMs sounds like a very good idea :) Maybe the imported TMs could be added to an `imports` field at the `thing` level (similar to the `@context` extensions) and then be referenced by affordances with an `extends` field? This could work similar to SDF's `sdfRef` (see https://tools.ietf.org/html/draft-ietf-asdf-sdf-01#section-4.4) which allows for inheriting and overriding definitions.

Inspired by SDF, I was also thinking about allowing for nested TMs/TDs where, for example, a sensor is its own `thing` or `object` which is encapsulated by another thing (in this case, for instance, a complex device with multiple sensors). But I guess this probably adds too much unnecessary complexity?

sebastiankb:
I think, two kind of imports are necessary: 

1. an  "extend" import to have the full definition of an other existing TM. For that the link container can be used with 'rel=extend' 
2. an import mechanism that allows to import sub-definitions of other existing TMs. This is the direction that is explained [here](https://github.com/w3c/wot-thing-description/pull/1024#issuecomment-767144430).

This kind of imports will be similar that is known from Java.

relu91:
Just some random thoughts about Thing Model extension mechanisms. If we stick with the current definition of a TM single inheritance is the only good option.  In fact, there are some features of a TM (and a TD) that can not "multiple inherited". For example, consider the following:
```js
//TM1
{
   properties : {
        temperature : {
            type: "number"
        }
   }
}
//TM2
{
   properties : {
        temperature : {
            type: "string"
        }
   }
}
//TD
{
   links : [{
        "rel": "extends",
        "href": "url/to/TM1",
        "type": "application/td+json"
     },{
        "rel": "extends",
        "href": "url/to/TM2",
        "type": "application/td+json"
     }
   ],
    properties : {
       temperature : {
           type : "???" // Which value should I choose?
       }
   }
}
```
*Notice*: using a combination of the two types breaks the client assumptions (?), see below.

As we all know this problem could be addressed using namespacing but how the TD will look like? See the next example:
```js
{
   links : [{
        "prefix": "tm1",
        "rel": "extends",
        "href": "url/to/TM1",
        "type": "application/td+json",
     },{
        "prefix": "tm2",
        "rel": "extends",
        "href": "url/to/TM2",
        "type": "application/td+json"
     }
   ],
    properties : {
       "tm1:temperature" : {
           type : "number"
       },
      "tm2:temperature" : {
           type : "string"
       },
   }
}
```

This solution is not preferred from a client's point of view. Taking for example a client that only knows TM1; it expected to interact with the web thing using a property called `temperature` not `tm1:temperature`. Therefore, he has to preprocess the TD retrieve the prefix of the TM is interested in, and then contact the prefix with the property name. This preprocessing step could even be more complex with TM1 already extended other TMs and had some prefixes. `type` is not the only feature that cannot be inherited; others are: `name`, `description`, `security` (maybe combo scheme can help here), etc. 

However, if we limit the TM definition to only a subset of the current feature they could multiple-inherited:
```js
//TM1
{
   properties : {
        temperature : {
            forms:[{
               "op": "readproperty",
               "href" : "http://mytemp.example.com:8080/temperature",
                "contentType": "application/xml",
               "htv:methodName": "GET"
           }]
        }
   }
}
//TM2
{
   properties : {
        temperature : {
           forms:[{
               "op": "readproperty",
               "href" : "http://mytemp.example.com:5683/temp",
                "contentType": "application/json",
               "htv:methodName": "GET"
           }]
        }
   }
},

//TM3
{
   properties : {
        temperature : {
        }
   }
},

//TD
{
   links : [{
        "rel": "extends",
        "href": "url/to/TM1",
        "type": "application/td+json"
     },{
        "rel": "extends",
        "href": "url/to/TM2",
        "type": "application/td+json"
     },
     {
        "rel": "extends",
        "href": "url/to/TM3",
        "type": "application/td+json"
     }
   ],
    properties : {
       //no problem here
       temperature : {
           forms: [{
               "op": "readproperty",
               "href" : "http://mytemp.example.com:8080/temperature",
                "contentType": "application/xml",
               "htv:methodName": "GET"
           },
           {
               "op": "readproperty",
               "href" : "http://mytemp.example.com:5683/temp",
                "contentType": "application/json",
               "htv:methodName": "GET"
           }
       }
   }
}
```
In my opinion, there are no problems in this case. There are no ambiguities between the three TMs since forms can be concatenated in the forms array. Even if the form specified collides in some sense, the `forms` property in the TD list alternatives therefore a client could choose one of the two interchangeably. Clients' assumptions are not broken either; it gets a TD with a temperature property as it is expecting. Another property that could be extended by multiple TMs is `@type`, my conjecture is that every array/object value in a TD can be multiple inherited. 

Are we gonna miss the ability to specify `name`, `description` etc. in TMs? I don't know but they all feel pretty implementation-specific whereas TM should describe abstract features of WebThings. 

Multiple inheritance might be a powerful feature to have since it gives the opportunity to combine classes in more complex ways.  However:
> With great power comes great responsibility

relu91:
Another use case that we have to take in mind is that TMs might be used also for discovery. Having single inheritance could also limit the discovery flexibility.

relu91:
The import mechanism looks nice because it can be used to use TM are components or modules. This would increase the flexibility of how to compose TD but we have to make it clear that composition is not inheritance ([Composition over ineritanche](https://en.wikipedia.org/wiki/Composition_over_inheritance)).  Mixing the two options in the TD might create some confusion. 

Another option that we should evaluate is to restrict inheritance only between TMs and use only composition in TDs.

JKRhb:
I think you are making a lot of good points, @relu91. However, with the example you gave here:

```
 properties : {
       //no problem here
       temperature : {
           forms: [{
               "op": "readproperty",
               "href" : "http://mytemp.example.com:8080/temperature",
                "contentType": "application/xml",
               "htv:methodName": "GET"
           },
           {
               "op": "readproperty",
               "href" : "http://mytemp.example.com:5683/temp",
                "contentType": "application/json",
               "htv:methodName": "GET"
           }
       }
}
```

I see the problem that you could have two distinct kinds of temperature (maybe ambient and water temperature) which would make it harder to distinguish. I also the the problem from my second scenario where you may have two TMs

```json
{
   "title": "Temperature Sensor",
   "description": "First Thing Model",
   "properties" : {
             "value": { "..." },
             "range": { "..." },
             "unit": { "..." },
        }
   }
}

{
   "title": "Humidity Sensor",
   "description": "Second Thing Model",
   "properties" : {
             "value": { "..." },
             "range": { "..." },
             "unit": { "..." },
        }
   }
}
```

which would become indistinguishable or at least harder to distinguish. 

For me, all of these points are actually arguments for allowing for nested TMs/TDs. This could be realized by defining an extra field `things` or `components` at the thing level where you could include other TMs/TDs as a map. This could look like this:

```json
{
    "title": "MyMultipleTemperatureThing",
    "components": {
        "ambientTemperature": {
            "title": "Ambient temperature",
            "properties": {
                "temperature": { "..."},
            }
        },
        "waterTemperature": {
            "title": "Water temperature",
            "properties": {
                "temperature": { "..."},
            }
        },
    }
}
```

This would allow both for using only single inheritance without losing any potential information and would satisfy the "composition over inheritance" principle. Although could lead to much more complex TDs and make it more to difficult to parse them I think this would solve a problem that would be difficult to avoid otherwise – I already encountered it when I wanted to reuse existing definitions for temperature and humidity sensors in a thing that provides both capabilities.

The suggestions I made above are once again inspired by IETF's SDF Working Group (see [here](https://tools.ietf.org/html/draft-ietf-asdf-sdf-01#section-6) for more information on their solution for composition).

relu91:
> I see the problem that you could have two distinct kinds of temperature (maybe ambient and water temperature) which would make it harder to distinguish. I also the the problem from my second scenario where you may have two TMs

You're right. My example was syntactically and operationally correct, but with semantics things get messy. However, even in OOP semantic checks of inheritance are not performed. For example, you could define a class Person with one method getName and say that the class Pet is a sub-class of Person. As humans we still something to say about design choices. 

Still, the component approach is good and I am not against it. Maybe we should evaluate again why not using links for this mechanism. You could just define a link type " `composedBy`" and link other web things.  Following your example it would be something like this:
```js
{
   name : "myComposedThing"
   links : [{
        "rel": "composedBy",
        "href": "url/to/TemperatureSensor",
        "type": "application/td+json"
     },{
        "rel": "extends",
        "href": "url/to/HumiditySensor",
        "type": "application/td+json"
     }
   ]
}
```
I think that this solution might be less readable, but it does not introduce any new concept.

sebastiankb:
@relu91 @JKRhb thanks for this very interesting discussion.

Regarding the first example [here](https://github.com/w3c/wot-thing-description/pull/1024#issuecomment-768966909) I would allow only the rel=extend for TMs since this is a kind of a modeling operation. TDs should only use the rel=type to provide the hint which TM is followed.  In my view this model concept with `extend` should be used in the following way:

Let's assume I have a basic model like a on/off model. 

```js
// Basic On/Off Model
{
   properties : {
        onOff : {
            type: "boolean"
        }
   }
}
```

Now, I like to design a new device class model 'Smart Lamp Control' that should be used as template for creating TD instances. This model will extend the existing basic on/off model with a dimmer:

```js
//Smart Lamp Control Model
{
   links : [{
        "rel": "extends",
        "href": "url/to/BasicOnOffTM",
        "type": "application/td+json"
     }],
   properties : {
        dimmer : {
            type: "integer",
            minimum: 0,
            maximum: 100
        }
   }
}
```

Creating a instance of this TM should result to TD instances like this:

```js
//TD instance from SmartLammpControlTM
{
   links : [{
        "rel": "type",
        "href": "url/to/SmartLammpControlTM",
        "type": "application/td+json"
     }
   ],
   properties : {
        onOff : {
            type: "boolean",
            forms: [...]
        },
        dimmer : {
            type: "integer",
            minimum: 0,
            maximum: 100,
            forms: [...]
        }
   }
```

TDs should always be self-contained. Theoretically, the link entry in the TD can also be omitted, since the full details of the Basic and Smart TM are reflected by the TD instance. It has more of a hint character.

JKRhb:
Thank you for your summary, @sebastiankb, I think you are providing a very good basis for the extend mechanism and the next steps of the discussion.

Should the `extend` concept also be used for overriding existing models? For example, if we have another model that inherits the Smart Lamp Control Model could this then override the dimmer's maximum like so:

```
//Smart Lamp Control Model With Higher Maximum
{
   links : [{
        "rel": "extends",
        "href": "url/to/SmartLampControlTM",
        "type": "application/td+json"
     }],
   properties : {
        dimmer : {
            maximum: 200
        }
   }
}
```

or should this rather be part of an import/reference mechanism?

sebastiankb:
This is a good question. I think there are different views on this. Overwriting would increased individualization of models. However, extensive overwriting loses the semantic meaning of the extended model. @kolotu provide a [comment](https://github.com/w3c/wot-thing-description/issues/168#issuecomment-768094780) that overwriting is not allowed in Vorto.

Maybe we can think about a hybrid: Overwriting is only allowed for metadata like `title(s)`, `description(s)`, `maximum`, `minimum`. JSON Schema types and the names of the properties, actions, and events are not allowed to overwrite.  How about this?

JKRhb:
> 
> 
> Maybe we can think about a hybrid: Overwriting is only allowed for metadata like `title(s)`, `description(s)`, `maximum`, `minimum`. JSON Schema types and the names of the properties, actions, and events are not allowed to overwrite. How about this?

I think this would a good compromise that also makes sense considering the class structure: As the dimmer is of subclass `IntegerSchema` you can only replace it with a definition of the same subclass but not with an `ArraySchema` for instance. 

This analogy might not be entirely applicable to the `ObjectSchema` class (as it can have vastly different shapes) but given the constraint that names and types of `properties` are not allowed to change this is probably already addressed. Should it be allowed, however, to add more fields to an `ObjectSchema`? If all existing fields and requirements are still present I guess this shouldn't do any harm, should it?

I was also thinking about `enums` and `oneOf` definitions – should it be allowed to extend these when all existing values/schemas are still present?

sebastiankb:
>As the dimmer is of subclass IntegerSchema you can only replace it with a definition of the same subclass but not with an ArraySchema for instance.

Right. I would support the exception only for the number - integer case. This is also motivated by the embedded world, where you can typically choose to take a compiled firmware with float-based arithmetic or not (e.g. NodeMCU firmware). 

> Should it be allowed, however, to add more fields to the properties of an ObjectSchema?

I think so, as the core definition of the extended TM will still be available and can be further specified by terms such as `minimum`, `maximum`, `enums`, etc.

sebastiankb:
I think the PR is ready to merge and a good base for the next PlugFest begin of March. 

TODOs:
- Import mechanism with $ref or similar (will be discussed during the next vF2F)
- more examples for the TM example section that are related to Device Model (@mlagally), SDF / OneDM (@mjkoster ) and Vorto (@kolotu)

sebastiankb:
From today's TD call:
- before merge small changes should be applied
- the required approach should be marked as wip, maybe another approach (from SDF) will be evaluated (#1046)
- The assertion about the placeholder should be also allow any characters
- model have to be defined in the TD information model (#1045)
- there should be more explanation how TDs can be generated from a TM (#1047)

sebastiankb:
all topics are addressed from last TD meeting.

index.html
```diff
@@ -4649,6 +4661,10 @@ <h2>Basic Concept</h2>
          Due to the definition of <a>Thing Model</a> the term <code>instance</code> can be omitted within the <code>version</code> container. 
         </p>
 
+    </section>
+    <section id="thing-model-etension-import" class="normative">
```
danielpeintner, line 5058: typo "thing-model-etension-import" -->"thing-model-extension-import"

index.html
```diff
@@ -4722,11 +4738,163 @@ <h2>Basic Concept</h2>
...
+
+<section id="thing-model-td-placeholder" class="normative">
+    <h3>Placeholder</h3>
+
+<p>
+    <a>Thing Model</a> can specify which terms should be taken in a TD instance, but their values are unspecific and are first known during TD instantiation. 
+    In such a case the placeholder labeling MAY be used in Thing Model that MUST be substituted with a concrete value when TD instance is created 
+    from the Thing Model. 
+    <span class="rfc2119-assertion" id="td-model-placeholder">
+        The string-based pattern of the placeholder MUST follow <code>{&ZeroWidthSpace;{PLACEHOLDER_IDENTIFIER}}</code>,
```
danielpeintner, line 5145: I don't think "&ZeroWidthSpace;" is needed

danielpeintner (commented):
I did not review all. I just stumbled over some parts

index.html
sebastiankb: ok, thanks

index.html
sebastiankb: strange I did not add this

sebastiankb:
From today's TD call:
- merge this PR
- we have to be careful about the overwriting number with integer (see https://github.com/w3c/wot-thing-description/issues/1054) --> @sebastiankb will provide a note about this after the merge

Issue #998, opened 2020-11-06
[security] API key and PSK security schemes are not referenced or explained


egekorkan:
In contrary to the other security schemes, API key and PSK are not referenced to other specifications that specify them. I thought a W3C recommendation cannot use a non standardized "standard" without properly explaining the way it works? The security guidelines document (https://w3c.github.io/wot-security/) provides no explanation either. In that regard, this has slipped through the review process before publication.

In another regard, if I am reading the TD recommendation for the first time, I have no idea what these security schemes are, when they are used, how they are used or simply where to find more information about them

mmccool:
@ereshetova (TODO: add Oliver Pfaff here also) would be good to comment.  API key was meant to cover proprietary schemes such as used in Amazon Greengrass, and simply means a key is provided in the URL (we might want to add an example, perhaps using an API template in a TM?).   PSK was meant for proprietary key-distribution schemes such as used in zero-trust systems, e.g. by Google, and also in OT contexts like smart factories.  This basically means a certificate exists that is used by TLS etc. but is set up out-of-band.  We should at least explain the purpose of these better; I agree that they are underspecified, and we should include the above examples at a minimum.  If anyone can suggest any references it would be helpful.

mmccool:
Here are some possible references (but web pages; not sure if there are "normative" references in the usual sense, which may or may not be an issue):
* API keys
    - Amazon: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-setup-api-key-with-console.html
    - Google: https://developers.google.com/maps/documentation/javascript/get-api-key
* PSK - Preshared Keys 
    - https://en.wikipedia.org/wiki/Pre-shared_key
    - https://en.wikipedia.org/wiki/TLS-PSK

"API key" is kind of a marketing term,  since it depends on other technologies.   In a sense these are "security patterns" widely used in industry, rather than "standards".   However, we do know that Amazon, etc. will use a combination of technologies in a certain way.  Google and Amazon, however, will use different combinations of technologies.  So this is still useful to identify expectations in a consumer... To make this more complete we might also want to identify the "vendor ecosystem" - we probably need some additional information to make this complete.   Right now it does specify where the API key can be found, though.

For PSK, there are also some variants (see TLS-PSK, there are multiple ciphersuites... which identifies the flavor, e.g. symmetric keys, or authenticated key exchange, etc).   IANA is also a good source:

[...]

Issue #923, opened 2020-06-29
[question] How to describe Philips Hue security scheme


egekorkan:
I have a Philips Hue system where the Hue bridge creates a key/token for the users that should be appended to URLs while sending requests. Below is an example request (from [their documentation](https://developers.meethue.com/develop/get-started-2/) with the token `1028d66426293e821ecfd9ef1a0731df` to read the state of the light number 1:

`GET https://<bridge ip address>/api/1028d66426293e821ecfd9ef1a0731df/lights/1`

This can be modeled as a property in a TD like:
```json
{
	"title": "Hue Light 1",
	"securityDefinitions": {
		"nosec_sc": {
			"scheme": "nosec"
		}
	},
	"security": "nosec_sc",
	"properties": {
		"lightState": {
			"forms": {
				"href": "https://192.168.0.1/api/1028d66426293e821ecfd9ef1a0731df/lights/1"
			}
		}
	}
}
```
I would imagine that I should be able to define a security scheme and not put the token in the href, which is visible to anyone who gets the TD. Following is using `myScheme` security scheme.

```json
{
	"title": "Hue Light 1",
	"securityDefinitions": {
		"mysec_sc": {
			"scheme": "myScheme",
                        "in":"uri",
                        "name":"mytoken"
		}
	},
	"security": "mysec_sc",
	"properties": {
		"lightState": {
			"forms": {
				"href": "https://192.168.0.1/api/{{mytoken}}/lights/1"
			}
		}
	}
}
```

@mjkoster @mmccool any ideas?

This is also related to https://github.com/w3c/wot-thing-description/issues/897 but this information needs to be displayed somehow in an actual TD and not a TDT.

relu91:
I want to add that also oAuth2.0 has some variability on how to construct `code` and `token` requests. For example OAuth2.0 simplified explains that the protocols parameters can be passed using a `POST` request ([here](https://www.oauth.com/oauth2-servers/access-tokens/authorization-code-request/)) while the rfc makes the same example using a `GET` request with URI query parameters ([here](https://tools.ietf.org/html/rfc6749#section-4.2.1)).

For oAuth we might even introduce Form elements inside the SecuritySchemaDefinition like the following: 
```js
"securityDefinitions": {
        "oauth2_sc": {
            "scheme": "oauth2",
            ...
            "flow": "code",
            "authorization": {
[...]

Issue #922, opened 2020-06-29
[question] How to describe apikey in query


egekorkan:
## Issue

The default location for the API key credential in `apikey` security scheme is stated to be in the `query` by the specification. There are no examples of this in the specification nor anyone has provided any inputs on the implementation report, i.e. searching for `apikey` in the wot-testing repository returns only TDs that have it in the header like the following:
```jsonc
{
...
"api_key_sc": {
      "scheme": "apikey",
      "in": "header",
      "name": "X-API-KEY"
    },
...
}
```

## Questions

1. Why are there no examples of `apikey` security scheme in the specification? There are also no mentions of it in the security guidelines or architecture. 
2. Has this been tested in any way? I mean, has any Consumer tried to consume a TD and interact with its Thing that has API key in the query. If this is the default way to do it, it should be tested.
3. Should this be the default assumption? Since no one in the group implemented it, the OpenAPI 3.0 does not mention it as common practice, [OpenAPI 2.0](https://swagger.io/docs/specification/2-0/authentication/api-keys/) says `The key is usually sent as a request header` and [this Stack Overflow post](https://stackoverflow.com/questions/5517281/place-api-key-in-headers-or-url) also says that it should be in the header.
4. Why is there no explanation at all for the meaning of words `header`, `query`, `cookie` and `body` in any of the W3C WoT Specifications? They are somewhat intuitive but not really. `query` can mean a lot of things but I guess here it is meant as a query parameter.
5. Looking at [OpenAPI 3.0](https://swagger.io/docs/specification/authentication/api-keys/), I find that the following request would correspond to using `apikey` security scheme with credentials in `query`:
```GET /something?api_key=abcdef12345```
Does the following `securityDefinitions` of a TD correctly represent it?
```jsonc
{
...
"securityDefinitions":{
  "api_key_sc": {
      "scheme": "apikey",
      "in": "query",
      "name": "api_key"
    },
}
...
}
```
5.1 If it is, should this query parameter always be added to the `href`s or should it be invisible in the `href`s and what happens once normal `uriVariables` are also used?. E.g: Below are forms of a property in a TD that has the security scheme shown above. Which one is correct?
```json
{
	"properties": {
		"temperature": {
			"uriVariables": {
				"unit": {"type": "string","enum": ["F", "C"]}
			},
			"forms": {
				"href": "http://example.org/thing/property/temp{?unit}"
			}
		}
	}
}
```
OR
```json
{
	"properties": {
		"temperature": {
[...]

PR #1031, opened 2021-01-17
Improve descriptions of apikey and psk security schemes

mmccool:
To resolve issue https://github.com/w3c/wot-thing-description/issues/998

Notes: 
1. The wotsec.ttl file has been updated, and rendering of index.html updated.
2. Will treat the problem of non-query URI templates separately and orthogonally (discussed in issue https://github.com/w3c/wot-thing-description/issues/923)
3. An example would be useful, as requested in issue https://github.com/w3c/wot-thing-description/issues/922.  Not handled in this PR, however.
4. td-validation.ttl also updated, although it seems redundant.  It would be helpful to clarify which is the master.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1031.html" title="Last updated on Jan 25, 2021, 11:57 PM UTC (20f91c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1031/091d6a0...mmccool:20f91c3.html" title="Last updated on Jan 25, 2021, 11:57 PM UTC (20f91c3)">Diff</a>


mmccool:
- Added copies of descriptions of apikey and psk to td-validation.ttl.  Which is the master?   The render script does not make clear.   It would be nice to not have to enter things twice.
- Render script seems to be broken, getting a syntax error in the included file.  Waiting for a fix.  Until this is fixed the changes will not get reflected in the index.html file. In the meantime, reviewers should look at the text in ontology/wotsec.ttl for the apikey and psk schemes.

mmccool:
Marking as "ready for review" but "not ready for merging".

mmccool:
@oliverpfaff notes that there is an IETF RFC for security terminology we should cite, but unfortunately it does not define either "API Key" (which is a "marketing" term...) but more surprisingly (since it is used in other RFCs like TLS-PSK) does not define "PSK" or "pre-shared key".

mmccool:
Rendering error seems to have been fixed now but new text still not showing up, so I'm probably not making the edit in the right place... will try to fix in/for the TD call, in the meantime please do try to provide feedback on the proposed text.  I plan to reword the last sentence in the PSK spec to make is sound less like an assertion (it really should be, but testing would be difficult, so...)

egekorkan:
This looks good and has the content that I would look for when I am seeing these terms for the first time

sebastiankb:
From today's TD call:
- the content looks good
- however, PR not ready yet, needs some render script check (@vcharpenay can you help here?)
- PR maybe ready to merge next week

danielpeintner:
>     * however, PR not ready yet, needs some render script check (@vcharpenay can you help here?)

I did look into the the "issue" and my local rendering does work just fine. I think the updated content is in index.html.
Maybe you want trying to re-run the render script.


[index.zip](https://github.com/w3c/wot-thing-description/files/5847792/index.zip)

mmccool:
* Thanks, I'll try the rendering again.  I assume a bug got fixed...
* In the discussion in the TD call and in the security call, we agreed that an assertion prohibiting putting secret keys in the TD would be useful.  It can't easily be tested, but can be manually asserted, and gives us a way to say that people who put secret keys in a TD are not compliant with the specification.   I plan to structure this as a single blanket assertion rather than assertions in particular schemes (so it is prohibited, using the MUST NOT RFC2119 keyword, for *all* schemes, current and future).   Wording of the assertion, to be added to the header section for all security schemes, to be as follows: "For all security schemes, private keys, passwords, or other sensitive information directly providing access should be shared and stored out-of-band and MUST NOT be stored in the TD."  This would be followed by the informative statement: "The purpose of the TD is to describe how to access the system if a consumer has authorization, not to grant that authorization."

mmccool:
Rendering fixed, changes discussed above implemented.  Ready to merge.

egekorkan:
One thing I have realized is that the API Key says that it can be in a proprietary format. Thus, wouldn't we need a keyword called format (or similar) that can be used to convey this information. Otherwise, one would need documentation on that format via out-of-band-information. The string value can be left open or constrained to a set of values if such formats are explored.

sebastiankb:
from today's TD call we will merge this PR.

PR #1032, opened 2021-01-18
Add URI template location for security scheme parameters

mmccool:
WIP; do not merge.  Content may be reviewed, but rendering needs to be fixed.

Resolves https://github.com/w3c/wot-thing-description/issues/923

- Add an option for the "in" value for security parameters to be "uri", which indicates use of a URI template
- Add descriptions of each option (all of them, not just the new one...)
- Add "parameters" field to allow mapping between "id" and "key" generic categories to particular URI template parameter names

NOTE: see ontology/wotsec.ttl and validation/td-validation.ttl for changes.  Render script is currently broken so HTML has not been updated.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1032.html" title="Last updated on Feb 9, 2021, 2:45 PM UTC (84cb52c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1032/eb726c8...mmccool:84cb52c.html" title="Last updated on Feb 9, 2021, 2:45 PM UTC (84cb52c)">Diff</a>


mmccool:
Needs an example.   Will review in Security call on Jan 18.   Once rendering is fixed I will add an example (which can also serve as an example for APIKey, since it can use the new "uri" location e.g. for the Philips Hue scheme).

mmccool:
Thoughts:
- Really should define data schema for uritemplate variables in scheme itself so it does not have to be replicated in each interaction.  So "parameters" will be replaced with something that directly provides data schemas for the variables used in URLs
- Combination schemes may be needed here.  For example, to model the Microsoft Oauth scheme, you might have to indicate a template parameter as well for the client ID, and maybe a data schema for the key encoded in the body...
- The "body" mechanism is underspecified, and probably needs a data schema for HOW the data elements (which may be plural...) are encoded in the body.  The complication here is that the security elements may be mixed in other other data elements, e.g. for posting an updated value to a property.
- Yes, really really needs an example or three.  Working on it.

sebastiankb:
From today's TD call:
- still in wip
- maybe ready next week then
- also render problems here (@danielpeintner can have look where the changes has do made to make the render script working)

danielpeintner:
>     * also render problems here (@danielpeintner can have look where the changes has do made to make the render script working)

It looks like the  additional text for "in" after "Specifies the location of security authentication information." shows up and also adds uri type to one of.

What seems not to be working is parameter in the table for APIKeySecurityScheme.

I guess what you need to do there is add a local property in https://github.com/w3c/wot-thing-description/blob/091d6a093e997ff0d6836bfd72dd0ee824c23b9c/validation/td-validation.ttl#L840-L851 for parameters and remove the global one.. 
adding something like the following
```

    sh:property [
        sh:path wotsec:parameters ;
        skos:definition """TODO description."""^^rdf:HTML ;
        sh:datatype xsd:string ;
        sh:defaultValue "auth"^^xsd:string ;
        sh:order 1 ;
    ] ;
```


like it has been done for wotsec:qop in DigestSecuritySchemeShape

@vcharpenay is this correct?
The table in my case gets the additional line.. for some reason the column "vocabulary term" is not populated with "parameters" ?

mmccool:
Plan to update the uri schemes to that "parameters" is actually "uriVariables" and is the same class as the other "uriVariables" and includes a data schema for the variables.  For example:

```
"securityDefinitions": {
   "template": {
      "scheme": "uri",
      "uriVariables": {
                "ID" : { "type": "string", "@type": "SecurityID" },
                "KEY" : { "type": "string", "@type": "SecurityKey" }
            }
      }
}
```

Note the use of `@type` to identify the semantic meanings of various fields, which gives people the freedom to pick the actual names of the URI template variables to avoid conflicts.  I also propose the convention of using all-caps for template parameter names although this is not enforced.

relu91:
The example looks fine. I was wondering if the `uriVariable` `type` attribute should be always fixed to `string` (declaring also as optional since it is constant), or allow other DataSchema types as `number` or even `object`. As discussed today, serializing numbers and objects as security keys would not be straightforward.

egekorkan:
I think that the example is sort of fine but I would be more inclined to not use the same term as uriVariables. They are not the same thing here and query parameters are also something that can change based on the request whereas here we want to actually convey the meaning of a static value in the URI that is just hidden in the TD due to security reasons.

mmccool:
So, a couple of issues came up:
1. DataSchemas and uriVariables are overkill since the type is always "string"
2. Multiple variables may not be needed in many use cases, and really there is almost always only one "secret" variable
3. The design could be greatly simplified if we only have one variable, since then we can just use "name" for it
4. If we do have multiple secrets, then the "combo" scheme could be used to combine multiple uri schemes

Here's a proposal, by way of example:
```json
"securityDefinitions": {
   "uri_key": {
      "scheme": "uri",
      "name": "KEY" 
    }
}
```

There might be cases where we need multiple value (e.g. there is a "public" client key, but we want to keep it out of the TD to avoid tracking, so...):
```json
"securityDefinitions": {
   "uri_key": {
      "scheme": "uri",
      "name": "KEY" 
    },
   "uri_id": {
      "scheme": "uri",
      "name": "ID" 
    },
    "uri_combo": {
      "scheme": "combo",
      "allOf": ["uri_key" ,"uri_id"]
    }
}
```
Also, it would be easier to add the "uri" location to ALL security schemes (not just APIkey).    We can argue that we can separately test parameter embeddings and security schemes but not necessarily all possible combinations (in other cases, if we had to test all possible combinations of other features, testing would be impossible...)

mmccool:
Discussed "name" proposal in Security TF call on 1 Feb.  Consensus was to go with the "name" option.

mmccool:
Ready to merge, so removed WIP in title.  Adding an example (such as the one above) would be nice, however.

mmccool:
I added an example, and while I was at it I realized an assertion was needed to disallow name conflicts.  In theory such conflicts can only arise where a given security scheme is used (and so multiple schemes that don't overlap COULD in theory use the same names) but this is annoying to check.  Ditto for scoping declarations, etc.  So I went with a stronger-than necessary rule that all URI variables used for security schemes must be distinct from all other variables, which is easy to check.  In practice putting a prefix in front of security variables like sec* (as I intend to update the example to use in a moment) makes this easier.

sebastiankb:
From today's TD call:
- we should review the situation about the URL parameters that can be defined in the interaction (uriParameter) vs. the definition can be done within the security definition. It should be clear of the TD processor where to get the right information
- more examples would be helpful to have clear understanding about this approach

mmccool:
It would be helpful to know what kind of additional examples would be useful.  Perhaps one more showing both a security parameter and a uri parameter in an interaction and discussing how name conflicts are to be avoided?

BTW, I think using full data schemas for uriVariables is possibly a mistake.  In general you can only embed "primitive" types in URIs, i.e. numbers and strings.   Objects and arrays are not feasible, so maybe rather than a full DataSchema in the uriVariables it should be BasicDataSchema (or whatever the class is for primitive types in the TD spec, if there is one).

For security, though, the "number" type has only a limited number of bits that is not good enough for security keys, so limiting security parameters to strings is necessary as well.  Since there is only one type that makes sense, simplifying the syntax to omit the type is my/our preference.

sebastiankb:
From today's TD call we decided to merge this PR.

Issue #883, opened 2020-02-28
Automate running TD rendering script


danielpeintner:
Changes to the TD document require to run 

`npm run render`
or
`./render.sh`

locally to get the generated `index.html`.

I wonder whether it makes sense to automate this process based on each commit.
Do we want to investigate how to do that or are there any reasons why we did not do that in the first place?

egekorkan:
Also as you mentioned, `npm run render` does not work for windows

takuki:
I use [cygwin](https://www.cygwin.com/) to run `npm run render` on Windows, and it runs fine.

takuki:
During March 6th TD telecon, Daniel mentioned he can look into the feasibility. 
However, the participants noted this is not a pressing issue. See minutes [here](https://www.w3.org/2020/03/06-wot-td-minutes.html).

sebastiankb:
what is the status here?

danielpeintner:
I can try to setup a GitHub action as mentioned [here](https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-nodejs). Maybe we do that after FPWD.

I see 2 issue:
1. We need to test with the official repo and can not easily play around without (I will look into whether it can be tested with a fork too)
2. We need to add credentials since after successfully rendering we need to push the update to the repo (see [here](https://github.com/marketplace/actions/update-files-on-github))

@ashimura I wonder whether there are other W3C repos doing something similar?

egekorkan:
It would good to know what kind of GitHub payment plan W3C has since our builds will consume minutes that is shared by the entire W3C organization in GitHub. See https://github.com/pricing

sebastiankb:
is this something what we can simple do as [GitHub Actions](https://github.com/w3c/wot-thing-description/actions/new)?

danielpeintner:
I think both can work, Travis and GitHub actions.

I noticed that the render script causes some _noise_ after each run (e.g., always updating SVG images etc). Not sure what we could do about that. If I work on a PR without changing SVG details I simply don't commit those changes. A script can not do so but that might be a non-issue...

danielpeintner:
@JKRhb I recall you mentioned that you want to work on automating the render process. Is this correct? I do not find any issue / mentioning anywhere. Thanks for any feedback!

JKRhb:
[...]

PR #1038, opened 2021-01-21
Add draft TD context URI

### LINKED ISSUE #1036: Wrong context in JSON Schema for TD v1.1
Hi,

In [TD 1.1](https://www.w3.org/TR/wot-thing-description11/) examples, the main context has changed to: `http://www.w3.org/ns/td`

However, the JSON Schema still enforces `https://www.w3.org/2019/wot/td/v1` as the first context:

https://github.com/w3c/wot-thing-description/blob/48e6fa20c6d769f5594eef7675852ac85024d139/index.html#L5140-L5166

This causes JSON Schema based validations to fail.

farshidtz:
This adds the new context URI used in the latest draft. See the issue for details.

It closes #1036.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-thing-description/pull/1038.html" title="Last updated on Jan 21, 2021, 10:39 PM UTC (01eac5e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1038/091d6a0...farshidtz:01eac5e.html" title="Last updated on Jan 21, 2021, 10:39 PM UTC (01eac5e)">Diff</a>

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -60,7 +60,8 @@
         "thing-context-w3c-uri": {
             "type": "string",
             "enum": [
-                "https://www.w3.org/2019/wot/td/v1"
+                "https://www.w3.org/2019/wot/td/v1",
+                "http://www.w3.org/ns/td"
             ]
         },
         "thing-context": {
--- a/index.html
+++ b/index.html
@@ -5211,8 +5211,7 @@ in a WoT Thing Description MUST accurately describe the <a>WoT Interface</a> of 
         "thing-context-w3c-uri": {
             "type": "string",
             "enum": [
-                "https://www.w3.org/2019/wot/td/v1",
-                "http://www.w3.org/ns/td"
+                "https://www.w3.org/2019/wot/td/v1"
             ]
         },
         "thing-context": {
--- a/index.html
+++ b/index.html
@@ -5211,7 +5211,8 @@ in a WoT Thing Description MUST accurately describe the <a>WoT Interface</a> of 
         "thing-context-w3c-uri": {
             "type": "string",
             "enum": [
-                "https://www.w3.org/2019/wot/td/v1"
+                "https://www.w3.org/2019/wot/td/v1",
+                "http://www.w3.org/ns/td"
             ]
         },
         "thing-context": {


egekorkan:
The changes have to be done at https://github.com/w3c/wot-thing-description/blob/master/validation/td-json-schema-validation.json

More precisely at https://github.com/w3c/wot-thing-description/blob/091d6a093e997ff0d6836bfd72dd0ee824c23b9c/validation/td-json-schema-validation.json#L63

Thank you for the PR :)

farshidtz:
Thanks @egekorkan , I also added it to that. I did not know that the schema is maintained in two locations. I think it is important to have it in index.html too because that is what is shown in the editor's draft and used by many.

Why not embed that inside index.html using the [data-include](https://respec.org/docs/#data-include) attribute? Example from discovery:
https://github.com/w3c/wot-discovery/blob/52a3a5f652c0e87f8943f787d3796de4c791bda4/index.html#L498

egekorkan:
@farshidtz it is actually done like that and index.html is actually generated from index.template.html that includes the schema at https://github.com/w3c/wot-thing-description/blob/091d6a093e997ff0d6836bfd72dd0ee824c23b9c/index.template.html#L4466

The generation process of the TD document is a bit convoluted :)

farshidtz:
Thank you for your patience @egekorkan  😊  I cleaned up my mess.

Though it would be nice if PRs which change auto-generated documents get rejected automatically with some CI magic.

egekorkan:
> Though it would be nice if PRs which change auto-generated documents get rejected automatically with some CI magic.

This would be actually super nice. @ashimura is this possible?

danielpeintner:
We did discuss to call the render the script automatically, see https://github.com/w3c/wot-thing-description/issues/883

I never found the time to look into the issue...

validation/td-json-schema-validation.json
```diff
@@ -60,7 +60,8 @@
         "thing-context-w3c-uri": {
             "type": "string",
             "enum": [
-                "https://www.w3.org/2019/wot/td/v1"
+                "https://www.w3.org/2019/wot/td/v1",
```
sebastiankb, line 63: we should only have the "http://www.w3.org/ns/td" in the list since this schema should cover 1.1 version only

sebastiankb:
From today's TD call:
- we should only have the "http://www.w3.org/ns/td" in the list since this schema should cover 1.1 version only (see above)
- the 1.1  covers the definition from 1.0
- when we have the CR we should use the 1.1 namespace (e.g., https://www.w3.org/2021/wot/td/v1)
- we should also think about to have a fix location of the JSON Schema file such as for the version 1.0 and 1.1 in the future

sebastiankb (approved):

PR #1049, opened 2021-02-15
TM Schema generation

egekorkan:
I have currently started with the requirements and will be pushing my code to this PR

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -1,5 +1,5 @@
 {
-    "title": "WoT TD Schema - 20 November 2020",
+    "title": "WoT TD Schema - 16 February 2021",
     "description": "JSON Schema for validating TD instances against the TD model. TD instances can be with or without terms that have default values",
     "$schema ": "http://json-schema.org/draft-07/schema#",
     "definitions": {
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -533,6 +533,17 @@
                 "maximum": {
                     "type": "number"
                 },
+                "minLength":{
+                    "type": "integer",
+                    "minimum": 0
+                },
+                "maxLength":{
+                    "type": "integer",
+                    "minimum": 0
+                },
+                "multipleOf":{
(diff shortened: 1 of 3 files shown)


egekorkan:
I have added the removal of required. The comments are also added and some questions:
- Should the `required` term and thus its value removed or should its value be replaced by `[""]` as to make sure that nothing is required.

egekorkan:
One important question, is the following allowed in a TM? (let's say that this a DataSchema

```js
"type":"object"
"properties":"{{}}"
```

Similarly, is the following allowed in a TM?

```js
@context: "...",
"actions":"{{}}",
"events":"{{}}"
```

danielpeintner:
> One important question, is the following allowed in a TM? (let's say that this a DataSchema)

I would say both are *not* allowed. 
properties/actions/events require a named object....

egekorkan:
@danielpeintner I have also thought the same but the spec is currently very open about this. I will open an issue

sebastiankb:
From today's TD call:
- we were not sure if the PR is still in wip
- we will merge next week then

egekorkan:
> we were not sure if the PR is still in wip

It is marked as draft, so still WIP

relu91 (approved):
The algorithm sees fine to me, but I agree that a TM model that looks like this:
```js
{
  "@context": "...",
  "actions":"{{}}",
  "events":"{{}}"
}
```
does not really make sense, plus you could obtain the same by just omitting the keywords:
```js
{
  "@context": "...",
}
```
Right? Probably the same is true for `securityDefinitions`:

```js
{
    securityDefinitions: "{{}}"
}
```

egekorkan:
I also think the same. Keys whose values are objects or arrays can be simply omitted

egekorkan:
I have pushed my non working string replacement method. It is not used at the moment but I am working on it. The recursion does not work and needs some good-old debugging

egekorkan:
Ready for testing/merging. Validates all the TM examples in the current spec.

In the end, I gave up on the _elegant_ recursion but manually wrote where the replacements should occur ^^

sebastiankb:
From today's TD call we will merge the schema. Will be a good input for testing TMs during PlugFest.

Issue #1037, opened 2021-01-20
The "body" location value for security schemes is underspecified


mmccool:
When a security key is embedded in the body of a message (eg as a POST) it will typically be part of a larger structured payload, so a data schema is required and the identification of a particular element in that schema as the key (or other elements, there might be more than one, i.e. a client id AND a key) is needed.  This needs to be clarified in the definition of the "body" location specifier in security schemes.

mmccool:
One option here is to allow use of "name" for "body" to allow identification of a data schema element.  Note however this is not currently a required element for body (so we can't make it mandatory without breaking backward compatibility in 1.1)

egekorkan:
to annotate specific parts of the body, we can use JSON Pointers

mmccool:
Propose closing, BUT maybe want to follow up on JSON Pointers idea mentioned by Ege.  Issue is "name" needs to refer to a particular element of a data schema (but, over several different data schemas, so would be relative to the top of each schema)

PR #1058, opened 2021-02-22
Add JSON pointer assertion to definition of body sec location

mmccool:
Resolve https://github.com/w3c/wot-thing-description/issues/1037

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1058.html" title="Last updated on Mar 23, 2021, 10:02 PM UTC (fcf54c2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1058/24c2bf8...mmccool:fcf54c2.html" title="Last updated on Mar 23, 2021, 10:02 PM UTC (fcf54c2)">Diff</a>


egekorkan (commented):
Some comments

- I think an example showing a pointer pointing to a level deeper would be nice, i.e. `/auth/key`.
- Should there be a note somewhere outside the security saying that a DataSchema can contain security related information, please see section [section you just added] ?
- I think this is the first case where the data schema is referenced from somewhere else. So now I "worry" about having a hard link between the data schemas and the security container. If I have a TM that I deploy with different security configurations, I will need to change the TM and also the TD. A possible solution could be to change the wording a bit to say that the term `key` does not have to be present in the DataSchemas. So the name does not refer to an existing location but a location where the key and its value should be added. This way, the DataSchema containers will not contain the term key in the TD.

mmccool:
JSON pointer syntax should be updated to use a leading # mark for internal references.  See also the geolocation spec.

index.html
```diff
@@ -3202,6 +3206,88 @@ <h3><code>securityDefinitions</code> and <code>security</code></h3>
...
+Note: if the security key occurs in inconsistent locations, it will be
+necessary to use multiple security scheme definitions.
+<pre class="example">
+{
+    ...
+    "securityDefinitions": {
+        "apikey_body": {
+            "scheme": "apikey",
+            "in": "body",
+            "name": "/key"
```
mmccool, line 3231: Should have a leading #, e.g. and maybe should also be "deeper", e.g. nested structure, like `#/auth/key`, which means changing the data schema too.

index.template.html
```diff
@@ -2311,6 +2311,88 @@ <h3><code>securityDefinitions</code> and <code>security</code></h3>
 Using a specific prefix as in the above example
 for URI variables declared in security schemes can make it easier to avoid name conflicts.
 </p>
+<p>Security parameters might also be included along with the payload in some
+systems.  For example, suppose a system requires every payload to be a JSON
+object including a member named <code>key</code> whose value is a secret key.
```
mmccool, line 2316: if make deeper make sure to update this too.   Also mention the correct syntax for JSON pointer somewhere.

index.template.html
```diff
@@ -2311,6 +2311,88 @@ <h3><code>securityDefinitions</code> and <code>security</code></h3>
...
+Depending on the interaction, however, other elements of the JSON object might 
+vary.  This situation can be dealt with using the <code>body</code> security
+information location.
+Note that for this location, the <code>name</code> parameter is actually a JSON pointer
```
mmccool, line 2320: add reference to JSON pointer

index.html
```diff
@@ -1599,6 +1599,10 @@ <h2>Security Vocabulary Definitions</h2>
           <dt><code>body</code>:</dt>
           <dd>The parameter will be provided in the body of the request payload, with the data schema element  
             used provided by <code>name</code>.
+            <span class="rfc2119-assertion" id="sec-body-name-json-pointer">When used in the context of a 
+            <code>body</code> security information location, the value of <code>name</code> 
+            <em class="rfc2119" title="MUST">MUST</em> be in the form of a JSON pointer [[!RFC6901]] relative to
```
mmccool, line 1604: Add assertion (or just a note) about leading #?

mmccool:
So regarding this comment from Ege:
* A possible solution could be to change the wording a bit to say that the term key does not have to be present in the 
DataSchemas. So the name does not refer to an existing location but a location where the key and its value should be added. This way, the DataSchema containers will not contain the term key in the TD.

This kind of relates to the discussion we had recently (in the plugfest) that dataschemas in TDs are meant to indicate only required elements but that it is permissible for a producer to include additional information.  So what you're basically saying is that the JSON pointer in the security schema would indicate where an "additionalProperty" should be *inserted*.

I'm kind of leaning towards allowing both.  If the element exists in the data schema, then it refers to it; fine.  But it's acceptable to omit it, and then it is implicitly inserted.  This might make processing a TD more complex, however; validators will have to make sure to process the security schemes to get the full dataschemas, and generally keys will naturally be "required" elements for interactions to work.   In short, an "interaction validator" looking just at the data sent by a consumer and the data schemas in the TD may miss that required security information is missing.  

Security TF met and felt either solution is fine, but would like input from TD re validation.

mmccool:
So people were also uncomfortable with starting the JSON pointer with #, since that (normally) refers to the entire document root, which in this case is the TD.  But this JSON pointer refers to EACH data schema in each interaction in which it is used.  On the other hand, using a "raw" JSON pointer (e.g. "/auth/key") can also be confusing since it looks like a file... maybe.  A final option would be to define a special starting character like "~" (e.g. "~/auth/key") but this just seems like another thing to annoy people and which needs extra processing.

mmccool:
Anyhow... now I'm personally leaning towards leaving off the `#` since it's a relative pointer.  But I think we should include the `#` in other contexts that do refer to the root of the TD, i.e. links (including internal geolocation links as in the geolocation proposal).   But at least we should add a note to the current draft explaining this choice, and maybe giving an example of an "expansion" filling in the implied part of the JSON pointer for each interaction.

index.html
mmccool: Actually... it shouldn't, not a fragment identifier.

mmccool:
OK, I made the following changes:
- Added some assertions to the definition of "body" to allow insertions.
- Added an assertion to indicate that the referenced (or created) element needs to be of type "string" and be required.
- Added an assertion to allow the "-" array index (refers to an element past the end of an array, to allow insertions at the end; unfortunately there does not seem to be a mechanism for inserting things elsewhere in an array, e.g. at the beginning)
- Current version still allows referring to already existing data schema elements.
- Expanded the example to show use of both existing data schemas and insertions; this also shows the "creation" of a data schema!  Insertion into an "empty" (unspecified) data schema essentially creates it.

mmccool:
I personally consider this ready to go so am taking off the WIP tag.

sebastiankb:
from today's TD call: ok to merge this PR

Issue #980, opened 2020-10-05
Definition of contentEncoding should be of type xsd:string


takuki:
Both in context/json-schema-context.jsonld and context/td-context-1.1.jsonld

    "contentEncoding": {
      "@id": "jsonschema:contentEncoding",
      "@type": "xsd:unsignedInt"
    },

It should be xsd:string.

sebastiankb:
From today's TD call:
- we should fix this 
- @sebastiankb can provide a PR

sebastiankb:
the issue was solved by the merged PR

PR #1060, opened 2021-02-24
fix issue #980

sebastiankb:
see #980


sebastiankb:
From today's TD call we will merge this PR.

PR #1065, opened 2021-03-08
fix: the "required" keyword was placed incorrectly in the TM schema

spf00:
The current TM schema handles the "required" keyword not like the Thing Model draft describes.  https://w3c.github.io/wot-thing-description/#thing-model-td-required <br>

Currently the "required" keyword is placed inside the interactions when instead it should be placed on the interaction-level next to properties, actions and events. <br>

This PR updates the TM Schema to work correctly by creating a new "required_element" in the definitions and referencing it in the properties, actions and events sections.

### DIFF
--- a/validation/tm-json-schema-validation.json
+++ b/validation/tm-json-schema-validation.json
@@ -94,6 +94,12 @@
         }
       ]
     },
+    "required_element": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
     "dataSchema": {
       "type": "object",
       "properties": {
@@ -249,12 +255,6 @@
         "properties": {
           "additionalProperties": {
             "$ref": "#/definitions/dataSchema"
-          }
-        },
-        "required": {
-          "type": "array",
-          "items": {
-            "type": "string"
           }
         }
       }
@@ -611,12 +611,6 @@
         "properties": {
           "additionalProperties": {
             "$ref": "#/definitions/dataSchema"
-          }
-        },
-        "required": {
-          "type": "array",
-          "items": {
-            "type": "string"
           }
         }
       },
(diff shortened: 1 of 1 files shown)


egekorkan:
A couple of points:
- The TM Schema is generated automatically from the TD Schema using a script. Any change to the TM schema will be simply overwritten at some point. By the way, I have realized that this is not clearly indicated somewhere. Would you expect this information to be in the JSON Schema or somewhere in the GitHub (Readme) or in the spec document? I would guess not in the spec document since it is not information for the reader.
- From the experience in the plugfest, this use of `required` will change. This would mean that this PR would be probably closed or heavily needs to be modified
- The required inside the Data Schema should be kept as it is. It is used to validate when someone puts a required in the DataSchema that is used for listing required keys of an object. Without that one, one can write an ObjectSchema with `"required":123`, which would result in a broken JSON Schema

egekorkan:
Sorry for closing, I have mistakenly  clicked on "Close with comment" instead of "Comment"

sebastiankb:
From today's TD call:
- the `required` term is under discussion. Most likely it will be replaced by another approach
- currently, the TM schema is generated by a script. To update the schema we need to update the script that generates the TM schema (https://github.com/w3c/wot-thing-description/blob/main/validation/tmSchemaGenerator.js)
- keep this PR open until we have clarified the `required` situation during the vF2F meeting in next 2 weeks

sebastiankb:
Please checkout the new approach that was provided by this PR #1092. 
@egekorkan can you update the JSON Schema with this new required  + the tm:ref approach

sebastiankb:
a new required approach was introduce here #1092. Close this PR

Issue #1023, opened 2020-12-10
Introduce a canonical TD format


mlagally:
Define canonical TD representation that satisifes the following requirements:

- Adopt JCS (https://www.rfc-editor.org/rfc/rfc8785)
- Default values - omit, when possible
- Prefixes - compact form
- Define handling of unordered arrays
- Handling of Array-of-one-element: suppress the square brackets?

See the discussion in the Profile issue: https://github.com/w3c/wot-profile/issues/55

Please create a document section that can be referenced from the Profile spec.

sebastiankb:
>Prefixes - compact form

What does this mean exactly? I would expect in a canonical form that the prefixes are resolved with the corresponding namespaces. 
@danielpeintner I think this is done in Canonical XML, right?

danielpeintner:
Correct. In the ideal case both the following snippets should have the **same** canonical form
Note: the difference is in the prefix used: `"saref"` vs. `"sf"`

```
{
    "@context": [
        "https://www.w3.org/2019/wot/td/v1",
        { "saref": "https://w3id.org/saref#" }
    ],
    "title": "MyLampThing",
    "@type": "saref:LightSwitch",
    "securityDefinitions": {
        "basic_sc": {"scheme": "basic", "in": "header"}
    },
    "security": "basic_sc"
}
```

vs.

```
{
    "@context": [
        "https://www.w3.org/2019/wot/td/v1",
        { "sf": "https://w3id.org/saref#" }
    ],
    "title": "MyLampThing",
    "@type": "sf:LightSwitch",
    "securityDefinitions": {
        "basic_sc": {"scheme": "basic", "in": "header"}
    },
    "security": "basic_sc"
}
```

sebastiankb:
but the expected canonical form would be:


```
{
    "@context": [
        "https://www.w3.org/2019/wot/td/v1" 
    ],
    "title": "MyLampThing",
    "@type": "https://w3id.org/saref#LightSwitch",
    "securityDefinitions": {
        "basic_sc": {"scheme": "basic", "in": "header"}
    },
    "security": "basic_sc"
}
```

danielpeintner:
Yes correct, but we need to describe it somewhere (if it is not already done so somewhere)

mmccool:
See discussion in https://github.com/w3c/wot-thing-description/pull/943

PR #1085, opened 2021-04-01
WIP: Add Validation Section

mmccool:
Add a validation subsection to the end of the Serialization section.  Defines three levels of validation appropriate to different contexts.

* Resolves Issue https://github.com/w3c/wot-thing-description/issues/1043
* Relevant to Issue https://github.com/w3c/wot-discovery/issues/143 (mostly resolves it, although we would still need to add a reference from WoT Discovery to Intermediate Validation for directories that do semantic processing; to discuss, how to deal with extensions...)
* Relevant to Issue https://github.com/w3c/wot-discovery/issues/99 (mostly resolves it, although we still need to add references to WoT Discovery to cite the appropriate levels of validation).

NOTES: 
* also added missing id in immediately preceeding Identification section.
* adjacent to proposed canonicalization section so order will depend on which one is merged first

- [x] define Basic Validation
- [ ] discuss and refine Basic validation definition
- [x] define Intermediate Validation
- [ ] discuss and refine Intermediate validation definition
- [x] define Full Validation
- [ ] discuss and refine Full validation definition
- [ ] add correct section links
- [ ] reference JSON schema in appendix
- [ ] add special case for TDs with no extensions (may need additional assertions, i.e. not allowing extra keywords if not defined in an extension)
- [ ] define/reference how context and SHACL files are to be fetched over the network
- [ ] provide SHACL files for TD
- [ ] label appropriate assertions with Basic tag
- [ ] label appropriate assertions with Intermediate tag
- [ ] Document tags in RFC2119 introduction section
- [ ] update change log

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1085.html" title="Last updated on May 5, 2021, 3:37 PM UTC (f4f38cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1085/0e54c8f...mmccool:f4f38cf.html" title="Last updated on May 5, 2021, 3:37 PM UTC (f4f38cf)">Diff</a>

PR #1086, opened 2021-04-01
Add section to define Canonical serialization

mmccool:
Adds a section to define a canonical form for TD JSON serialization suitable for signing, etc.

* Resolves Issue https://github.com/w3c/wot-thing-description/issues/1023
* Relevant to Issue https://github.com/w3c/wot-profile/issues/55 (to resolve would have to add reference to this section from the Profile document)
* Relevant to Issue https://github.com/w3c/wot-thing-description/issues/940 (canonicalization is a prerequisite to signing)
* Relevant to Issue https://github.com/w3c/wot-discovery/issues/104 (directories should return canonical form)
* Related to PR https://github.com/w3c/wot-thing-description/pull/1085 defining validation (TDs should be validated before canonicalization; we have to decide what level is appropriate, however)

- [x] create section
- [x] add links to relevant issues from this PR
- [x] reference JSON Canonicalization Scheme (IETF [RFC8785](https://tools.ietf.org/html/rfc8785)) as foundation
- [x] Define specific ordering rules to make sure security definitions come before use and metadata is before interactions.  Note: this is only applied at the top level.  Ordering rules inside objects like securityDefinitions is not going to be covered, since it would be rather complex to define. (Edit: removed)
- [x] Define rule to say that elements that are array valued but may also be a single element are serialized without square brackets when there is only one element
- [x] Define rule to handle defaults (Edit: Current version requires all default values to be filled in, including for extensions)
- [x] Define rule to regenerate prefixes for extension vocabularies. (Edit: removed; add assertion that TD Processors must preserve prefixes instead)
- [x] Add appropriate TD Processor assertions (e.g. to preserve order of all arrays, etc)

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1086.html" title="Last updated on Apr 22, 2021, 3:55 PM UTC (584e7ea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1086/75e70b0...mmccool:584e7ea.html" title="Last updated on Apr 22, 2021, 3:55 PM UTC (584e7ea)">Diff</a>


egekorkan:
I think that this is not ready for review yet. One comment in advance regarding default value removal: For op values in a Property Form, the default value says:

> Array of string with the elements readproperty and writeproperty when readOnly and writeOnly are set to false or Array of string with the element readproperty when readOnly is set to true or Array of string with the element writeproperty when writeOnly is set to true.

The canonicalization algorithm should pay attention to this since it is not a simple default value on its own but depends on the data schema

mmccool:
No, not ready for review yet. Sorry, I assigned you for review, but if it's marked as draft it's not ready yet.

mmccool:
Anyhow I'll look at the default issue you mention.  But the basic rule will be if the value given would have been the default (however calculated; this is not the canonicalization algorithm's job...) then it will be omitted from the canonical form.

mmccool:
didn't consider connection to validation yet.  When/if that is merged we will have to consider whether a canonical TD must be validated first or not.

mmccool:
BTW I considered but rejected adding a step to canonicalize URIs, eg by normalizing case.  See https://www.w3.org/DesignIssues/Axioms.html.  This means that TD Processors also MUST NOT attempt to canonicalize URIs.  I am a bit worried though that some usual-but-wrong things like using special characters in URLs that need to be encoded might cause problems.

Also... I need to add some assertions.  Some of these need to go into TD Processing, things like "MUST NOT reorder arrays", etc.

BTW alternatively we could sort elements in arrays where order does not matter but then we would need an explicit list of which ones to sort (which would be annoying) since not all should be.

mmccool:
Anyway, if TDs need to be validated before canonicalization, then illegal URIs could be rejected and the problem I noted above goes away.

egekorkan:
@mmccool wrote:

>  I am a bit worried though that some usual-but-wrong things like using special characters in URLs that need to be encoded might cause problems.

Also URIs in protocols other than HTTP may have special meanings so I would advise against this.

mmccool:
On URIs:
- Basic validation will catch and reject illegal URIs
- But what if some protocol INSISTS on using illegal URIs?

index.html
```diff
@@ -3930,6 +3938,64 @@ <h2>Identification</h2>
...
+  using the template <code>e{n}</code> (where <code>{n}</code> is an integer sequence number starting
+  at 1), and use these prefixes in place of base URIs. Note: some systems discard the prefixes
+  and expand all vocabulary terms into URIs.  This rule allows the canonical form to be
+  generated even in this case. It does, however, lead to the loss of the original prefixes.</li>
+  <li>Apply the JSON Canonicalization Scheme [[!RFC8785]], which among other transformations,
+  sorts object members by name, removes white space, and normalizes number representations.</li>
+  <li>As an exception to the sorted order,
+  move <code>securityDefinitions</code> to just before <code>security</code> at the top level.
+  This is to make single-pass processing easier, so these definitions 
+  occur prior to their first use.</li>
```
farshidtz, line 3990: The second bullet is going to be very hard to implement. Basically, we will not be able to use existing processors which implement the likes of RFC8785 JSON Canonicalization Scheme.

I suggest leaving consumer requirements out of this. Otherwise, there will be many more reordering exceptions: e.g. `base` before `actions` and more when we dig into each object tree.

index.html
```diff
@@ -3930,6 +3938,64 @@ <h2>Identification</h2>
...
+  <li>Remove any elements that are given their default value. 
+  For example, the default value of <code>observable</code>
+  is <code>false</code>.  
+  If an input TD explictly assigns the value <code>false</code> to <code>observable</code>,
+  this element and the redundant value will be removed.</li>
+  <li>Rewrite values that can be expressed as either an array or as a single value 
+  as a single value if there is only one element.  
+  In other words, remove square brackets around
+  arrays of single elements where this is permitted.
+  </li>
```
farshidtz, line 3979: It will be a programming burden to process a defaults list and remove attributes from a complex object. What about extensions?
I would have suggested only removing `null` assignments if I were sure there are no domain-specific semantics behind them.

index.html
```diff
@@ -3930,6 +3938,64 @@ <h2>Identification</h2>
...
+  </p>
+  <p>Computing a canonical form directly from an RDF triple representation is
+  difficult so the starting point is the JSON serialization of a TD.  
+  The following sequence of transformations are then applied to this serialization: 
+  <ol>
+  <li>Remove any elements that are given their default value. 
+  For example, the default value of <code>observable</code>
+  is <code>false</code>.  
+  If an input TD explictly assigns the value <code>false</code> to <code>observable</code>,
+  this element and the redundant value will be removed.</li>
```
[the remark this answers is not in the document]
egekorkan: Actually the op keywords are not the default values. The `read-writeOnly` is not specified, thus both are false and an op's default value, in this case, would be `["readproperty","writeproperty"]` . Thus, the `op` should not be removed.

This is a very important aspect of default removal and insertion but is a detail at the same time, leaving room for errors. That is what I wanted to imply at https://github.com/w3c/wot-thing-description/pull/1086#issuecomment-813406125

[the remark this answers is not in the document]
danielpeintner: > Actually the op keywords are not the default values.

https://w3c.github.io/wot-thing-description/#example-42 needs to be fixed than?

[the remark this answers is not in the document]
egekorkan: Yes :grin:

index.html
mmccool: The problem is that when you pull things into a database, you will fill in all the default values.  Later you don't know whether a value was assigned during import or by the originator.  Would only apply to defaults defined in the TD spec, not in extensions.

index.html
mmccool: Good point about base.  We either need to do this for all such things, or not.

index.html
farshidtz: Why doesn't it apply to extensions? Extensions can also default values and the same issue arises when pulling that data out of such default-aware databases. The canonicalization algorithm should work on extended TDs too. 

I think this will be easier if TD spec provides the defaults in a simple machine-readable and extensible way. Maybe a TD with only default values to compare against the input TD during this stage of the algorithm?

index.html
mmccool: Anyhow, as stated this is broken, because "security" also needs to come before "action".  I think what I'll do is just specify the order the top-level members of Thing need to be listed in.  We can, however, separately decide if we even want to require an ordering different from just a plain sort.

index.html
mmccool: Defaults are handled in a special way by a TD processor, basically filling in missing items during import.  Extensions would not have defaults; we might even want to require this.

mmccool:
Regarding URLs, what I am thinking is that TDs should only ever allow "legal" URLs (this is in fact the current spec).  If we ever encounter some protocol that *requires* a special illegal character in a URL we can cross that bridge when we come to it; I'd suggest in that case adding a protocol binding element that specifies that a certain character should be decoded (reversing percent-encoding) before the URL is used with the given protocol.

mmccool:
I updated the last rule to give an explicit order to ALL top-level elements in a Thing, including extensions, to remove any ambiguity.  Along the way I noticed that in fact "base" is NOT part of the Thing information model!  I assume this is intentional but it means that many of the TDs that I have generated as examples are in fact incorrect.
I also added a note about illegal URLs.

The multiple-forms-collapsing-into-one is an interesting issue, perhaps I should also add something about that; e.g. after the rule for "default removal" state that "if default removal results in elements of the Forms array having the same value then duplicates should be removed".  Are there places other than Forms where something like this comes up?

farshidtz:
> Along the way I noticed that in fact "base" is NOT part of the Thing information model! I assume this is intentional but it means that many of the TDs that I have generated as examples are in fact incorrect.

It used to be in TD 1.0; I created an issue: #1091.

index.html
```diff
@@ -3931,6 +3939,93 @@ <h2>Identification</h2>
...
+    any extension terms (in sorted order),
+    <code>links</code>,
+    <code>securityDefinitions</code>,
+    <code>security</code>,
+    <code>forms</code>,
+    <code>properties</code>,
+    <code>actions</code>, and
+    <code>events</code>.
+  This reordering is to make single-pass processing easier, so that (for example) definitions 
+  occur prior to their first use.</li>
```
farshidtz, line 4016: I have a feeling that this is mostly to make TDs human readable and better looking in the spec document.

Most importantly and as mentioned before, suggesting RFC8785 JSON Canonicalization Scheme for only parts of object and a different sorting order for the rest means implementations will not be able to use available libraries or any kind of existing JSON serializer for that matter. Note that in most programming languages, JSON is purely the serialized output of an internal object (class/struct instance) or dictionary/map. For generic directories which need to blindly host and give back TDs with extensions (with and without namespace prefixes), anything other than an alphanumeric sorting will require complex and inefficient custom serialization.

I understand that single-pass processing may be a consumer requirement, but the suggested ordering doesn't match every use case. Imagine the following example from a smart city application:
```json
{
    "@context": [
        "<domain-specific-context-url>"
    ],
    "id": "urn:uuid:15eb6579-29b9-4329-a952-3a679eab8cc4",
    "type": "motionSensor", // alias for @type
    "site": "alpha", // extension
    "projectID": "80581098-5f7f-4e8b-be02-d1d82fb4b161", // extension
    "ifcZoneID": "ea971d5e-96df-4369-9ee5-7a5bac67437d", // extension
    "ifcSpaceID": "a7b3931f-aab5-4e4f-88de-ad369a2c4393", // extension
    ...
    "properties": {
       ...
     },
     ...
}
```
Here, the title(s) and description(s) have no use in single-processing. Instead, several domain-specific extensions need to be consumed to know the location of the thing before processing the rest.

Thus, I suggest not prescribing any custom ordering for canonicalization and relying fully on RFC8785. The servers which have single-processing consumers can provide appropriate custom ordering (which we cannot generalize) through server-driven content negotiation.

index.html
farshidtz: I think they would have (unless we are talking about different things). Some examples:
https://www.w3.org/TR/wot-thing-description11/#http-binding-assertions
https://www.w3.org/TR/wot-binding-templates/#mqtt-default-vocabulary-terms

danielpeintner:
> The multiple-forms-collapsing-into-one is an interesting issue, I will also add something about that for us to discuss; e.g. after the rule for "default removal" state that "if default removal results in elements of the Forms array having the same value then duplicates should be removed". Are there places other than Forms where something like this comes up?

I can see similar issues in most of the places where arrays are possible.

e.g.

`"type": ["Foo", "Foo"]`

_could_ be revised by removing duplicates and become

`"type": ["Foo"]`

which then based on the other rules MUST become

`"type": "Foo"`

index.html
```diff
@@ -3931,6 +3939,93 @@ <h2>Identification</h2>
...
+    <code>support</code>,
+    any extension terms (in sorted order),
+    <code>links</code>,
+    <code>securityDefinitions</code>,
+    <code>security</code>,
+    <code>forms</code>,
+    <code>properties</code>,
+    <code>actions</code>, and
+    <code>events</code>.
+  This reordering is to make single-pass processing easier, so that (for example) definitions
```
sebastiankb, line 4015: base is missing in the list.

index.html
```diff
@@ -3931,6 +3939,93 @@ <h2>Identification</h2>
...
+  preserves the order of elements in arrays.
+  There are some array-valued elements in TDs where the order is not meaningful, such
+  as the list of <code>scopes</code> in a <code>Form</code>.  
+  However, implementations of TD Processors need to preserve the order even in
+  such cases so a Canonical TD can be generated if needed.
+  Also, TDs should be validated before being canonicalized.  
+  This will require that URLs
+  in particular be fully confomant with the standard and must in particular 
+  percent-encode characters outside the basic set defined for URLs.
+</section>
```
sebastiankb, line 4028: an example would be good for better understanding the process

sebastiankb (changes requested):

sebastiankb:
From today's TD call:
- decide not to merge
- PR  should address the comments above
- talk about this issue next week again

farshidtz:
I just read [here](https://json-ld.org/spec/latest/json-ld/#sets-and-lists) that arrays in JSON-LD are unordered by default unless the property is "abbreviated by setting @container to @list in the context". This is important, otherwise, triplestores WILL NOT preserve the order.

danielpeintner:
> I just read [here](https://json-ld.org/spec/latest/json-ld/#sets-and-lists) that arrays in JSON-LD are unordered by default unless the property is "abbreviated by setting @container to @list in the context". This is important, otherwise, triplestores WILL NOT preserve the order.

Interesting.  [RFC 7159](https://www.rfc-editor.org/rfc/rfc7159.txt) for JSON states:
"An array is an **ordered** sequence of zero or more values."

index.html
mmccool: After further thought I am inclined to agree with you.  There are enough other things that make canonical form non-human-readable that it is moot.  Consumers will have to simply have enough memory to read an entire TD into memory before processing it.

index.html
mmccool: I agree, but note that a "correct" example would have no white space and so would be very hard to understand.  That said, perhaps a "pretty-printed" example can be given, maybe with a multiple-view example and a big warning that the pretty-printed version is NOT canonical.

mmccool:
Regarding arrays being unordered in JSON-LD by default: this relates to some assertions I need to add for TD Processors, one of which should be that "TD Processors MUST preserve the ordering of all arrays."  So this should be fixed in the TD context file.

mmccool:
Regarding defaults, I realized there is one big problem: defaults in extensions.  In particular, the protocol binding for HTTP, which is non-normative (I think, although we were discussing moving it into the TD spec...) has defaults, some of which are related in complicated ways to "op", which has its own defaults.  In general though, if extensions have default values for some things these will be unknown to the canonicalizer in general and it will be impossible for an automatic system to canonicalize such TDs.  This will show up specifically in protocol bindings.

After giving this a lot of thought I think the only solution is as follows: canonical form must not allow default values for anything, extensions included.  Defaults given for TD vocabulary must be "filled in".  This actually makes implementation of round-tripping through an RDF database easier as we already need to "fill in" default values when reading a TD into the database, so to export a canonical form we just do... nothing.

The only new bit is adding an assertion that if an extension vocabulary is used, someone or something needs to fill in all default values for that extension before the TD can be canonicalized.

Tomorrow I plan to make an update along these lines.  I also plan to remove the "reordering" step to place definitions before uses, and generate an example.

mmccool:
I'm also thinking it's best to remove the rule about removing duplicates.  If there are duplicates, for whatever reason, they should be preserved.  Note: we should check that RDF processors will respect this with our current set of definitions in the context file.  Let's not make assumptions...

mmccool:
Various updates:
- replaced rule to remove default values with another one saying they must always be included
- removed rule to reorder top level (now just sorted alphabetically)
- removed rule to rename prefixes.  This did not work for prefixes declared inside context files, like htv.
- added a rule to make use of Z mandatory in dateTime (otherwise changes to timezones would break canonicalization)
- added an example, both with and without whitespace
- added assertion that TD processors cannot reorder arrays
- added assertion that TD processors must preserve prefixes

Please review and comment.

farshidtz:
I find it much more pragmatic now. It may be useful to also describe a "minimal" form, i.e. canonical without defaults for applications with constrained bandwidth.

Also, we could benefit from a media type or media type [parameter](https://tools.ietf.org/html/rfc6838#section-4.3) to specify the canonical representation. For example, to request the canonical form from an HTTP server: `Accept: application/td+json;  form=canonical`

Or is the `application/td+json` expected to be the canonical form?
> All registered media types MUST employ a single, **canonical** data
   format, regardless of registration tree.
https://tools.ietf.org/html/rfc6838#section-4.4

mmccool:
Farshid,
* Regarding the "minimal" form, it will run into the same problem with defaults in extensions.  However, perhaps we could define it to only suppress "normative" defaults (e.g. as defined in the TD spec) but still require defaults be explicit for extensions.  If we did that, I would just include it in the base definition rather than having a separate minimal form.
* There definitely needs to be a way to request a canonical form during discovery both from directories and when reading a TD directly from a device.  I like the idea of a contentType "encoding" parameter.  We can make an issue to add that to the Discovery spec.  When I add signing I will also mandate that TDs need to be canonicalized before signing.  But in general TDs would not have to be canonical (they are a pain to read and edit).

farshidtz:
> Regarding the "minimal" form, it will run into the same problem with defaults in extensions. However, perhaps we could define it to only suppress "normative" defaults (e.g. as defined in the TD spec) but still require defaults be explicit for extensions. If we did that, I would just include it in the base definition rather than having a separate minimal form.

Right, I meant it as a best-effort basis. Unlike for canonicalization, applications can decide what to drop toward minimizing the TD size in order to save bandwidth. Leaving a few defaults behind doesn't break applications.

Anyway, this is an off-topic and can be discussed later in discovery if some use case demands it.

mmccool:
To clarify, the default rule needs to be able to survive round-tripping through an RDF database.  When a TD is in a database, we don't know if a given value was generated by a default or not.  The current rule, saying defaults must be filled in, just means we do NOT have to check for them when we output a TD from an RDF database (which makes the implementation easier).  Suppressing ALL defaults gets complicated, even if we leave out extensions, so at this point I'm inclined to favor the simpler-to-implement option, even if it is more verbose.

mmccool:
Anyway, I would like to suggest we merge this as a baseline, and create issues to discuss specific proposed changes to the default rule/minimal form, etc.  We also can deal with the other thing (how to specify that you want to fetch a canonical form) in the Discovery spec.  Note that I don't necessarily want to force a small device to be able to generate a canonical form, so it needs to be optional in peer-to-peer discovery (for unsigned TDs, anyway).  For directories we probably want the ability to generate a canonical form to be mandatory.  We should also make it clear that a Thing is free to return a canonical form in all cases if it wants (for a small device using a TD template, this would be the easiest way to support canonical form; just make sure the template is canonical).

mmccool:
@farshidtz Sorry, I think I misunderstood, you're talking about "minimal" form being a different thing (e.g. another section), not an option for canonicalization, correct?  I def want only a single canonical form ;)

mmccool:
MAY also want to add an assertion that TD Processors must not modify URLs (for example, attempting to "canonicalize" them by converting to all lower case, etc).  Input URLs should be legal as I've already noted.  Protocol bindings will have to invent some way to deal with protocols that insist on "illegal" URLs *only if it ever comes up,* e.g. by specifying percent-encoded characters that need to be decoded.

sebastiankb:
from today's TD call:
- group decide to merge
- fix the white spaces in second example
- add assertion that a TD processor must not modify the URLs 
- @mmccool will do the changes and merge this PR

Issue #1081, opened 2021-03-25
Introduce tmRef for TM


sebastiankb:
As discussed in yesterday's [TD session](https://github.com/w3c/wot/blob/main/PRESENTATIONS/2021-03-online-f2f/2021-03-24-WoT-F2F-TD-Sebastian.pdf) we need an import mechanism for TM definitions.

It is also proposed to rely on the SDF approach known as `sdfRef`. For WoT we would introduce `tmRef` (alternative name would be `tdRef` or `wotRef`).

sebastiankb:
PR merged

PR #1090, opened 2021-04-12
init tmRef 

sebastiankb:
also see #1081

- [x] introduction 
- [x] assertion about the copy mechanism
- [x] assertion about context type of tmRef
- [x] complex example

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1090.html" title="Last updated on Apr 21, 2021, 2:07 PM UTC (cb25206)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1090/24c2bf8...cb25206.html" title="Last updated on Apr 21, 2021, 2:07 PM UTC (cb25206)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5062,7 +5062,7 @@ in a WoT Thing Description MUST accurately describe the <a>WoT Interface</a> of 
     In the following there is a sample Thing Model for a lamp without any protocol and security information.
     </p>
     
-    <pre class="example" title="MyLampThingModel serialized in JSON">
+    <pre class="example" title="MyLampThingModel serialized in JSON" id="td-model-example-lamp">
     {
         "@context": ["http://www.w3.org/ns/td"], 
         "@type" : "ThingModel",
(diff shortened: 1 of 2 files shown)


JKRhb:
Thank you for adding this feature! It already looks very solid to me. I have one question, though: Should it be possible to import an definition from a TM that itself is extended or imported ? If so, maybe one sentence could be added to clarify to procedure when importing from such a TM (for example "all extensions and imports in the referenced TM have to be resolved before importing a definition").

sebastiankb:
I think, this makes sense. The process should be also explained when a TD are initiated from a TM (@relu91 maybe something also for #1095).

JKRhb:
> 
> 
> I think, this makes sense. The process should be also explained when a TD are initiated from a TM (@relu91 maybe something also for #1095).

Great, thank for your feedback on this!

One more thing: In SDF, you can override and extend imported definitions. In [section 6.2.1](https://tools.ietf.org/html/draft-ietf-asdf-sdf-05#section-6.2.1) of the current draft the following example is given:

```json
"sdfData":
     "length" : {
       "type": "number",
       "minimum": 0,
       "unit": "m",
       "description": "There can be no negative lengths."
     }
   ...
     "cable-length" : {
       "sdfRef": "#/sdfData/length",
       "minimum": 0.05,
       "description": "Cables must be at least 5 cm."
     }
```

I suppose this should also be possible in the case of TMs? If so, maybe a clarification and a similar example could be given in the WoT specification?

sebastiankb:
from today's TD call:
- not ready to merge
- we should support the [input](https://github.com/w3c/wot-thing-description/pull/1090#issuecomment-819530553) from Jan 
- in general we should point out that the 'semantic' should be not changed when overwrite definitions (e.g., do not change type string to number)

sebastiankb:
from today's TD call: decide to merge this PR

Issue #1080, opened 2021-03-25
Replace required with tmRequired for TM


sebastiankb:
As discussed in yesterday's [TD session](https://github.com/w3c/wot/blob/main/PRESENTATIONS/2021-03-online-f2f/2021-03-24-WoT-F2F-TD-Sebastian.pdf) the current [required solution](https://w3c.github.io/wot-thing-description/#thing-model-td-required) is problematic. 

Proposal is to rely on the the SDF approach known with `sdfRequired`. For WoT we would introduce `tmRequired` (alternative name would be `tdRequired` or `wotRequired`).

benfrancis:
I believe `tm:required` was also suggested?

It might be useful to provide a summary in this issue of why the [`required`](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.5.3) keyword from JSON Schema can not be used.

sebastiankb:
>I believe tm:required was also suggested?

Oh, I cannot member. I will check the minutes. It wouldn't make that much sense, though, because we have to introduce a new namespace with the prefix "tm". 

>It might be useful to provide a summary in this issue of why the required keyword from JSON Schema can not be used.

The term `required` is mainly used to indicate which elements within a JSON schema definition are mandatory when creating JSON instances. However, we also like to use `required` outside the JSON schema definition, e.g. which actions are mandatory. We would use the new term `required` outside of the interactions to simplify TD/TM processing (see slide 12 of my presentation). This also means that we need a JSON pointer approach, which is not specified in the JSON schema.

PR #1092, opened 2021-04-13
rename required to tmRequired + top level definition 

sebastiankb:
also see #1080

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1092.html" title="Last updated on Apr 21, 2021, 1:32 PM UTC (a33652b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1092/24c2bf8...a33652b.html" title="Last updated on Apr 21, 2021, 1:32 PM UTC (a33652b)">Diff</a>

### DIFF
--- a/context/td-context.jsonld
+++ b/context/td-context.jsonld
@@ -8,6 +8,7 @@
     "xsd": "http://www.w3.org/2001/XMLSchema#",
     "dct": "http://purl.org/dc/terms/",
     "htv": "http://www.w3.org/2011/http#",
+    "tm": "https://www.w3.org/2019/wot/tm#",
     "@vocab": "https://www.w3.org/2019/wot/td#",
     "license" : {
       "@id" : "http://purl.org/dc/terms/license"
--- a/index.html
+++ b/index.html
@@ -5295,14 +5295,13 @@ in a WoT Thing Description MUST accurately describe the <a>WoT Interface</a> of 
     <span class="rfc2119-assertion" id="td-model-required">
         To guarantee the implementation of particular kind of interaction models,
          <a>Thing Model</a> definitions MUST use the JSON member 
-        name <code>required</code> at interaction level 
-        (<code>properties</code>, <code>actions</code>, and <code>events</code>)
-        to provide the name of the interaction in a string-based array list.
+        name <code>tmRequired</code> as JSON array at the top level that provides JSON Pointer references to 
+        the required interaction model definitions.
     </span>    
 </p>
 
 <p>    
-The following sample shows the usage of <code>required</code> for the <a>Property</a> interaction <code>status</code>
+The following sample shows the usage of <code>tmRequired</code> for the <a>Property</a> interaction <code>status</code>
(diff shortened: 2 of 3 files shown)


egekorkan:
I see some issues:
- The reference to JSON Pointer seems to be missing
- 10.4 still uses `required`. In the meantime, 10.4 has typos and assertive words like have to, must that are not RFC assertions
- Title of the example uses `required`
- I think that it is not 100% clear that the pointer should resolve in the affordance object/map, i.e. the following: 

> To guarantee the implementation of particular kind of interaction models, Thing Model definitions MUST use the JSON member name tmRequired as JSON array at the top level that provides JSON Pointer references to the required interaction model definitions. 

Should be split into multiple assertions, something like:

> To guarantee the implementation of particular kind of interaction models, Thing Model definitions MUST use the JSON member name tmRequired. `tmRequired` MUST be a JSON array at the top level.  It MUST provide JSON Pointer references to the required interaction model definitions. The JSON Pointers MUST resolve to an entire interaction affordance Map definition.

This way, all of them can be separately tested and also understood by readers.

sebastiankb:
@egekorkan many thanks. I updated the text.

>10.4 still uses required. In the meantime, 10.4 has typos and assertive words like have to, must that are not RFC assertions

@relu91 can you take this point also in your PR #1095

relu91:
> @relu91 can you take this point also in your PR #1095

Renamed and the _assertive typos_ should be already fixed. ✔

relu91 (commented):
Did you evaluate to use a `json-ld` prefix instead of changing `required` to `tmRequired`?. With `json-ld` we could achieve the same in a more consistent pattern:
```js
{
    "@context": ["http://www.w3.org/ns/td",{
       "tm": "http://www.w3.org/ns/td"
     }], 
    "@type" : "tm:ThingModel",
    "title": "Lamp Thing Model",
    "description": "Lamp Thing Description Model",
    "tm:required": [
        "#/properties/status",
        "#/actions/toggle"
    ],
    "properties": {
        "status": {
            "description": "current status of the lamp (on|off)",
            "type": "string",
            "readOnly": true
```

The downside is that we have to force implementers to use `tm` as a prefix otherwise non-json-ld processors would not understand.

sebastiankb:
I see the advantages of this tm: approach and may show a clear separation what belongs to TD data model and TM data model. The disadvantage would be that tm: context would be very small, as we would only have tm:required and tm:Ref in it. For now, I would keep this in the TD model and have it as tmRequired. For the WoT 2.0 version, maybe we can think about a clearer separation and the TM itself will be defined in a separate document.

sebastiankb:
From today's TD call:
- not ready to merge
- it seems it make sense that we introduce an own namespace for TM
- tmReference can be defined as tm:reference
- this should be also true for tmRef #1090

sebastiankb:
from today's TD call: merge this PR

Issue #1047, opened 2021-02-10
Two step generation of the TD from a TM should be clear


egekorkan:
Currently, two things happen when a TD is generated from a TM (in no particular order, for now):

- Placeholders are replaced with a map. This is properly specified in the current version
- Required terms for a TD such as security definitions or forms are generated somehow. This is only shown in the examples in the current version but specified in the text. It should be made clear that this is left to implementation

Also as mentioned, the TM with no placeholder is a partial TD according to Scripting API. It would be good to combine the efforts here. In any case, these two generation steps should be made clear

Note: Thanks @relu91 for putting my concern into a proper format ::)

sebastiankb:
Actually, there a couple more actions which have to be covered. The PR is updated by:

![image](https://user-images.githubusercontent.com/13832739/108203310-9eae9380-7122-11eb-819a-8f43008b9d13.png)

egekorkan:
Some comments regarding TD (not TM) validation:
- Can a TD have an `@type` with a value of ThingModel? -> I guess not so a change in TD Validation Schema would be needed
- Can a TD have links with `rel:extend` ? -> I guess not so a change in TD Validation Schema would be needed

These have implications to the TD and I would say that such restrictions should be at least hinted in the core part of the document. If I am only interested in the TD and not in TM, how would I know these restrictions? So somewhere in Chapter 5 should say something like "A TD document cannot contain some name-value pairs that are reserved for the Thing Models. Please refer to the Thing Model chapter to find out the restrictions"

sebastiankb:
> Can a TD have an @type with a value of ThingModel? -> I guess not so a change in TD Validation Schema would be needed

based on point two above this should be replaced. But checking this in schema would be great.

> Can a TD have links with rel:extend

So far, it is only motivated for TM and should be removed in the TD (see last point)

relu91:
> Actually, there a couple more actions which have to be covered. The PR is updated by:
> 
> ![image](https://user-images.githubusercontent.com/13832739/108203310-9eae9380-7122-11eb-819a-8f43008b9d13.png)

I think that the algorithm looks good from the TD specification point of view. However, it would be really great to reflector it somehow so that we could reuse some steps in [this task](https://github.com/w3c/wot-scripting-api/issues/304).

Concretely, it would be sufficient to talk about PartialTD and move the "missing communication etc." to a separate algorithm since we have our own algorithm to concretely create a TD using runtime capabilities.

relu91:
[...]

PR #1095, opened 2021-04-13
Two step generation of the TD from a TM 

relu91:
This PR addresses the comments in #1047. In practice, it refactors the generation process of a Thing Description from a Thing Model in two distinct phases. In the first part, it creates a Partial TD. In other words, the process replaces every place holder and Thing Model specific keywords with the corresponding values. The resulting object is a Partial TD because it cannot be guaranteed that the original TM was fully complete (e.g., it did not have forms or security meta-data). Finally, the last phase describes how to take the resulting Partial TD (no extends, no place holders) and create a valid Thing Description. 

I consider this work as a first proposal and we might need to discuss a couple of points:
- Should we mandate a two-step generation process? I used the keyword MUST. We might need to consider if we just require the complete set of steps but not the fact that the processor MUST first have an intermediate Partial TD object. 
- The final step of the process is a little bit vague
- The algorithm does not mention any validation procedure, should we apply this process only to valid TMs?

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -4429,7 +4429,7 @@ instance.
     <p>
         <a>Thing Models</a> can be used as template to generate <a>Thing Description</a> based on the restriction defined in 
         Sections <a href="#sec-vocabulary-definition"></a> and <a href="#sec-td-serialization"></a>. During this process 
-        missing data such as communication and seceurity metadata have to be complemented to create valid <a>Thing Description</a> instances.
+        missing data such as communication and security metadata have to be complemented to create valid <a>Thing Description</a> instances.
         <span class="rfc2119-assertion" id="thing-model-td-generation-inconsistencies">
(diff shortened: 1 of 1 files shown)


sebastiankb:
From today's TD call:
- for each processing step we need separate assertions
- tmRequired needs to be changed to tm:required
- check the definition of partial TD with the Architecture document  and provide a reference

relu91:
> check the definition of partial TD with the Architecture document and provide a reference

partialTD is definined in [this line](https://github.com/w3c/wot-thing-description/pull/1095/files#diff-20c762491016637ceac07ca2c6c28cc19333d98b482bac4d863ac95cf88e9b6bR614) as other Architecture terms.

sebastiankb (approved):

sebastiankb:
From today's TD call we decided to merge this PR. Maybe there will be another PR which works on some wording.

Issue #1072, opened 2021-03-22
Move terminology definitions to architecture


mlagally:
Please move Vocabulary terminology definitions to Architecture:

https://github.com/w3c/wot-architecture/issues/575

Definition of Thing Model is already in architecture - should be double-checked and removed from TD

sebastiankb:
done

PR #1102, opened 2021-04-22
Update Terminology Section

sebastiankb:
also see #1072

@relu91 I also introduced _Partial TD_ in the list


sebastiankb:
from today's TD call we decided to merge this PR.

Issue #1101, opened 2021-04-22
jsonschema:StringSchema: pattern missing and maxLength is defined wrongly


fennibay:
In the jsonschema ontology:

1. [jsonschema:StringSchema](https://w3c.github.io/wot-thing-description/ontology/jsonschema.html#StringSchema) doesn't have a pattern property, although this is defined in [JSON Schema Spec.](http://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.3.3) -> pattern property should be added // see also #1094 
2. [jsonschema:maxLength](https://w3c.github.io/wot-thing-description/ontology/jsonschema.html#maxLength) states in text that it should apply to strings, but the domain is set to numbers and strings -> domain should be corrected to StringSchema

p.s. I logged this here as I couldn't find a separate github repo for the jsonschema ontology after discussion with @sebastiankb

egekorkan:
Although I generally do not oppose to adding JSON Schema terms, what is the use case here? We should always document addition of new features with corresponding use cases/requirements/

danielpeintner:
Question: Do we really need a use-case and/or requirement for each feature or try to add all of the JSONSchema terms.
I'm not sure, though how much work that would be...

sebastiankb:
I think, with regards to the origin from existing approaches like JSON Schema, we should not always provide use cases that have probably already been discussed there and are kind of obvious to be reused in TD as well. Here, the simple generic use case would be to allow restrictions on the data model definitions.

fennibay:
Sorry for my late comment. Use case is that we have some string property affordances that need to follow a certain pattern. Specifically these are some date time formats, but with added wildcards, so they cannot be represented with standard formats of JSON schema.

Generally, I think JSON schema ontology should provide whatever the JSON schema spec. is providing, with some exceptions like $ref, as that can be covered with RDF mechanisms already.

egekorkan:
In that case, we can simply just add all the JSON Schema terms of draft 7 and don't lose time for discussion and also stopping to confuse people when some terms are there and some are not. I had proposed this in Princeton Plugfest but it was rejected due to lack of need for the terms and having to maintain them all.

sebastiankb:
>I had proposed this in Princeton Plugfest but it was rejected due to lack of need for the terms and having to maintain them all.

[...]

Issue #1094, opened 2021-04-13
How to extend Data Schema terms while being JSON Schema compliant


egekorkan:
The TD Spec says:

> The data schema vocabulary definition is reflecting a very common subset of the terms defined by JSON Schema [JSON-SCHEMA]. It is noted that data schema definitions within Thing Description instances are not limited to this defined subset and may use additional terms found in JSON Schema using a TD Context Extension for the additional terms as described in § 7. TD Context Extensions, 

This is a bit too open for Data Schema terms that are meant to be JSON Schema terms. For example, assuming that JSON Schema has an OWL vocabulary at `"https://jsonschema.org/vocab"` and I want to use the term `pattern` of JSON Schema that allows the definition of regular expressions and that is not the TD vocabulary. If I write a TD like the following:
```js
{
    "@context": [
        "http://www.w3.org/ns/td",
        {
            "schema": "https://jsonschema.org/vocab",
        }
    ],
    ...
    "properties": {
        "phoneNumber": {
            "description": "Phone number to send SMS notifications to",
            "type": "string",
            "schema:pattern": "^(\\([0-9]{3}\\))?[0-9]{3}-[0-9]{4}$",
            "forms": [...]
        }
    }
}
```

A semantic processor would understand this but a JSON Schema processor would fail to understand the pattern keyword. I would have thus failed on using JSON Schema. The correct TD should be thus something like the following:

```js
{
    "@context": [
        "http://www.w3.org/ns/td",
        "https://jsonschema.org/vocab"
    ],
    ...
    "properties": {
        "phoneNumber": {
            "description": "Phone number to send SMS notifications to",
            "type": "string",
            "pattern": "^(\\([0-9]{3}\\))?[0-9]{3}-[0-9]{4}$",
            "forms": [...]
        }
    }
}
```

Thus, I would propose to change the note above (or add a new one) to restrict the usage of JSON Schema vocabularies to not be used in a Map but appended to the `@context` array.

P.S. This issue brought to you by https://github.com/w3c/wot-thing-description/issues/1093#issuecomment-818865847

sebastiankb:
Proposal: When using terms that are not deffined in DataSchema and do not have a prefix, it must be assumed that the origin is in the JSON schema. If this is not intended the terms must have a prefix with specific context.

fennibay:
So I understand the proposal here is to use the jsonschema:pattern (currently missing, see #1101) instead of [schema:pattern](https://schema.org/pattern), correct?

[...]

Issue #1093, opened 2021-04-13
Is there a way to set default values for uriVariables?


farshidtz:
Hi,

I was wondering if there is a way to set default values in TD for uriVariables?

The [URI Template](https://tools.ietf.org/html/rfc6570) does not support it after [draft v5](https://tools.ietf.org/html/draft-gregorio-uritemplate-05#section-2.5) and I couldn't find anything in the TD's DataSchema either.

Example Property with non-standard default values:
```json
{
    "uriVariables": {
        "offset": {
            "description": "The number of items to skip before the subset",
            "type": "number",
            "default": 0 
        },
        "limit": {
            "description": "The number of items in the subset",
            "type": "number",
            "default": 1 
        },
        "count": {
            "description": "Request the total number of items in the collection",
            "type": "boolean",
            "default": false
        },
        "sort_by": {
            "description": "The attribute used as comparator for sorting of the collection",
            "type": "string",
            "default": "id"
        },
        "sort_order": {
            "description": "Sorting order",
            "type": "string",
            "enum": ["ASC", "DESC"],
            "default": "ASC"
        }
    },
    "forms": [
        {
            "href": "/items{?offset,limit,count,sort_by,sort_order}",
            ...
        }
    ]
}
```

_Originally posted by @farshidtz in https://github.com/w3c/wot-discovery/pull/130#r590380965_

egekorkan:
JSON Schema allows setting default values in v7 onwards by the use of `default` keyword.

farshidtz:
So is it going to become a part of DataSchema through the "very common subset of the terms defined by JSON Schema"? 

Or I have to extend the TD context as Sebastian suggests? It isn't [here](https://www.w3.org/2019/wot/json-schema) so I'll have to find use https://schema.org/defaultValue.

@sebastiankb replied [here](https://github.com/w3c/wot-discovery/pull/130#discussion_r612464611):
> TD allows to use terms which are not defined in data schema yet. Information can be found here https://w3c.github.io/wot-thing-description/#sec-data-schema-vocabulary-definition

egekorkan:
> Or I have to extend the TD context as Sebastian suggests? It isn't here so I'll have to find use https://schema.org/defaultValue.

I am not sure how one extends Data Schema terms but schema.org is not the correct one. This is the correct reference at least: https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.9.2

[...]

Issue #956, opened 2020-08-18
JSON Schema validation issue wr.t. "format": "iri-reference"


danielpeintner:
I tried to use the JSON Schema file we have in the TD and used it with https://jsonschemalint.com.

At first the JSON schema seems fine but once I actually try to check a TD instance I get the following message

> `unknown format "iri-reference" is used in schema at path "#/definitions/anyUri"`

Could it be the that it has been renamed to `"uri-reference"` instead?
see https://github.com/json-schema-org/json-schema-spec/issues/355

Once I change it to `"uri-reference"` everything works as expected.

egekorkan:
The correct modeling is indeed `iri-reference` since but even the most popular json schema validation libraries do not support it as noted in the TD spec:
>  Please note that some JSON Schema validation tools do not support the iri string format. 

Playground uses the same validator as in jsonschemalint.com and there is the following note: https://github.com/thingweb/thingweb-playground/blob/master/playground-core/td-schema.json#L8

https://www.jsonschemavalidator.net/ does support `iri-reference` though :) Also, here is an example validation: https://www.jsonschemavalidator.net/s/pk2OEcK9

danielpeintner:
> https://www.jsonschemavalidator.net/ does support `iri-reference` though :) Also, here is an example validation: https://www.jsonschemavalidator.net/s/pk2OEcK9

I am not sure whether https://www.jsonschemavalidator.net/  actually supports (or ignores it) given that you can use `"format": "xxx-uri-reference"` also without any error.

According to http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.5 both is supported....
@handrews may I ask you what is the term to be used?

*My* goal is to have something that just works without any tweaks like [this](https://github.com/thingweb/thingweb-playground/blob/master/playground-core/td-schema.json#L8)

egekorkan:
> I am not sure whether https://www.jsonschemavalidator.net/ actually supports (or ignores it) given that you can use "format": "xxx-uri-reference" also without any error.

It actually supports it and JSON Schema spec says that if the format is unknown, the validation should pass. See here: https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7

To see that it works, you can have an é in on of the `href` values. If you change `iri-reference` to `uri-reference`, the validation will fail

handrews:
[...]

PR #1104, opened 2021-04-22
Address multiple DataSchema issues

sebastiankb:
This PR updates some JSON Schema / DataSchema issue such as:

- [x] #1101
- [x] #1093
- [x] #1094
- [x] #956

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1104.html" title="Last updated on May 5, 2021, 3:36 PM UTC (a5a80ea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1104/4d08013...a5a80ea.html" title="Last updated on May 5, 2021, 3:36 PM UTC (a5a80ea)">Diff</a>


egekorkan:
- The text for default says:

>  Supply a default JSON value. The value should validate against the schema in which it resides. 

I think that we should not mention JSON here since all other data schema terms are not for JSON values only. Also, default is one of those terms that is easy to apply to other payload formats like `text/plain`.

Also, the word `schema` should be replaced with Data Schema

- Regarding format:

> It is already removed, i.e. not under discussion anymore. We need a better explanation for this.


- Regarding assuming JSON Schema term in Data Schema:

First I had thought that it was fine but actually it makes the following example weird/incorrect:

```js
{
    "@context": [
        "http://www.w3.org/ns/td",
        "https://schema.org"
    ],
    ...
    "properties": {
        "myProperty": {
            "type": "string", //doesn't matter
            "longitude": -122.08585,
            "forms": [...]
        }
    }
}
```

Since `longitude` is specified without a prefix, according to the spec, this should be understood as a JSON Schema term. It is not totally incorrect since when a term is unknown by a JSON Schema processor, it should just ignore it.

Example from https://schema.org/longitude

sebastiankb:
> Since longitude is specified without a prefix, according to the spec, this should be understood as a JSON Schema term.

The full text is: 
>It is noted that data schema definitions within Thing Description instances are not limited to this defined subset and may use additional terms found in JSON Schema. It is assumed that such additional terms within DataSchema definitions that have not used a prefix have their origin in JSON Schema.

So the 'additional terms' refer to JSON schema terms and the longitude would not make sense here.

sebastiankb:
from today's TD call:
- for default we should provide a note that says that default is annotation 
- fix the conflict
- discuss this next week again

sebastiankb:
I removed the text
>It is assumed that such additional terms within DataSchema definitions that have not used a prefix have their origin in JSON Schema. 

again and provide the text of REC version again. We have to decide if we integrate all missing terms as discovered here #1119 for the next update. If so we can remove this paragraph for the CR then.

sebastiankb:
from today's TD call we decided to merge

Issue #877, opened 2020-02-03
No mechanism to indicate the "desired" form entry


danielpeintner:
We got the request in the Scripting API task force that scripts might want to choose which form is used. In the sample TD below a script would like to decide which form entry is used for a given interaction (e.g., `form[0]` for the interaction `toggle` because it uses HTTP).

```
{
	"@context": "https://www.w3.org/2019/wot/td/v1",
	"title": "MyLampThing",
	"securityDefinitions": {
		"basic_sc": {
			"scheme": "basic"
		}
	},
	"security": ["basic_sc"],
	"actions": {
		"toggle": {
			"forms": [{
					"href": "https://mylamp.example.com/toggle"
				},
				{
					"coap": "https://mylamp.example.com/toggle"
				}
			]
		}
	}
}
```

Besides the desired intent of a script e.g., preferring HTTP (which is not supported in scripting yet) there is also no way to indicate in a TD what form entry is _preferred_. WoT runtimes are free to pick any form while it seems that there is an _unwritten_ assumption that the order in the array might be an indicator.

Does it make sense to have a more explicit mechanism in a TD that indicates the preferred form entry or to make the _unwritten_ assumption more evident in the specification?

mmccool:
Even if no meaning is assigned to the order, I think we should have an assertion that processors should not reorder forms arbitrarily, just to future-proof against assigning a meaning to the order later.

zolkis:
> processors should not reorder forms arbitrarily

I think that is the right formulation (SHOULD NOT) since clients MAY still choose whichever Form found in the TD. 

However, as discussed in the Scripting call, it makes sense to be able to select the Form from an application, so the Scripting API will add that option to the interactions exposed by ConsumedThing.

Also, when specifying an ExposedThing with Forms, Scripting API implementations SHOULD NOT change the order clients specified for Forms. This is independent of the decision the TD TF has made and ensures compatibility with eventual later policies.

takuki:
In TD call on 2/14, we discussed and suggest to say something like "Clients may choose the first form entry that works for them." in TD specification. This does not introduce any new constraint.

sebastiankb:
@takuki can you provide a PR for this statement?

sebastiankb:
please checkout PR #1108 if it is working for you

sebastiankb:
PR merged

PR #1108, opened 2021-04-23
consumer forms selection 

sebastiankb:
also see #877

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1108.html" title="Last updated on Apr 28, 2021, 12:39 PM UTC (00336e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1108/24c2bf8...00336e1.html" title="Last updated on Apr 28, 2021, 12:39 PM UTC (00336e1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2097,9 +2097,11 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
           associated with the name <code>additionalResponses</code> that
           includes a value for the name <code>schema</code>.</span> 
           </p><p>
-          A <a>Thing Description</a> of a WoT producer may has multiple forms entries with, e.g., different 
-          protocol and/or content types declarations that a <a>Consumer</a> would also support. In that case
-          it is recommended that the <a>Consumer</a> chooses the first form entry that works for them.
+          A <a>Thing Description</a> of a WoT producer may have multiple forms entries with, e.g., different 
+          protocol and/or content types declarations that a <a>Consumer</a> could possibly support. In that case 
+          the Consumer may choose any form entry that works (e.g., the protocol and content type is supported) for them. 
+          When one form is chosen, it is expected that the <a>Consumer</a> will continue to use it as long as possible 
+          for every new interaction with the WoT producer.
           </p>
           </section>
 <section><h3><code>ExpectedResponse</code></h3><p>Communication metadata describing the expected
--- a/index.html
+++ b/index.html
@@ -2096,7 +2096,12 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
           "rfc2119" title="MUST">MUST</em> include an entry in the array
           associated with the name <code>additionalResponses</code> that
           includes a value for the name <code>schema</code>.</span> 
-          </p></section>
+          </p><p>
+          A <a>Thing Description</a> of a WoT producer may has multiple forms entries with, e.g., different 
+          protocol and/or content types declarations that a <a>Consumer</a> would also support. In that case
+          it is recommended that the <a>Consumer</a> chooses the first form entry that works for them.
+          </p>
+          </section>
 <section><h3><code>ExpectedResponse</code></h3><p>Communication metadata describing the expected
           response message for the primary response.</p><table class="def"><thead><tr><th><a>Vocabulary term</a></th><th>Description</th><th>Assignment</th><th>Type</th></tr></thead><tbody><tr class="rfc2119-table-assertion" id="td-vocab-contentType--ExpectedResponse"><td><code>contentType</code></td><td>Assign a content type based on a media type
                 (e.g., <code>text/plain</code>) and potential
--- a/validation/td-validation.ttl
+++ b/validation/td-validation.ttl
@@ -727,8 +727,8 @@
           associated with the name <code>additionalResponses</code> that
           includes a value for the name <code>schema</code>.</span> 
           </p><p>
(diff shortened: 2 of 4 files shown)


index.html
```diff
@@ -2096,7 +2096,12 @@ <h2>Hypermedia Controls Vocabulary Definitions</h2>
           "rfc2119" title="MUST">MUST</em> include an entry in the array
           associated with the name <code>additionalResponses</code> that
           includes a value for the name <code>schema</code>.</span> 
-          </p></section>
+          </p><p>
+          A <a>Thing Description</a> of a WoT producer may has multiple forms entries with, e.g., different
```
zolkis, line 2100: nit: may have

index.html
```diff
@@ -2096,7 +2096,12 @@ <h2>Hypermedia Controls Vocabulary Definitions</h2>
           "rfc2119" title="MUST">MUST</em> include an entry in the array
           associated with the name <code>additionalResponses</code> that
           includes a value for the name <code>schema</code>.</span> 
-          </p></section>
+          </p><p>
+          A <a>Thing Description</a> of a WoT producer may has multiple forms entries with, e.g., different 
+          protocol and/or content types declarations that a <a>Consumer</a> would also support. In that case
```
zolkis, line 2101: "could possibly support" (may be more clear to me, but you decide)

index.html
```diff
@@ -2096,7 +2096,12 @@ <h2>Hypermedia Controls Vocabulary Definitions</h2>
...
+          it is recommended that the <a>Consumer</a> chooses the first form entry that works for them.
```
zolkis, line 2102: I think this should be formulated in more generic terms, something like "In that case the <a>Consumer</a> may choose any form entry that works for them".

validation/td-validation.ttl
```diff
@@ -726,7 +726,12 @@
           "rfc2119" title="MUST">MUST</em> include an entry in the array
           associated with the name <code>additionalResponses</code> that
           includes a value for the name <code>schema</code>.</span> 
-          </p>"""^^rdf:HTML .
+          </p><p>
+          A <a>Thing Description</a> of a WoT producer may has multiple forms entries with, e.g., different
```
zolkis, line 730: nit: may have

validation/td-validation.ttl
```diff
@@ -726,7 +726,12 @@
           "rfc2119" title="MUST">MUST</em> include an entry in the array
           associated with the name <code>additionalResponses</code> that
           includes a value for the name <code>schema</code>.</span> 
-          </p>"""^^rdf:HTML .
+          </p><p>
+          A <a>Thing Description</a> of a WoT producer may has multiple forms entries with, e.g., different 
+          protocol and/or content types declarations that a <a>Consumer</a> would also support. In that case
+          it is recommended that the <a>Consumer</a> chooses the first form entry that works for them.
+          </p>
```
zolkis, line 733: ah, copy-paste. Then same comments as above :)

zolkis (approved):
Approved with nits and alternative formulations to consider.

egekorkan:
I agree with @zolkis comments, otherwise good :)

danielpeintner (changes requested):
LGTM besides comments from @zolkis

index.html
relu91: Should we also suggest a retaining policy? For example "When one form is chosen, it is expected that the consumer will continue to use it as long as possible for every new interaction with the Producer" ? 🤔  

Also _works for them_ feels a little bit vague but I understand that we might not need to be specific about implementation details.  What happens if the form chosen is unreachable? The consumer should give up or try again with another form?

relu91 (approved):
Just a little comment more than what @zolkis  already said.

index.html
sebastiankb: @zolkis 
>I think this should be formulated in more generic terms, something like "In that case the Consumer may choose any form entry that works for them".

I thought the idea was to provide a proposal, which one the Consumer should select if there are multiple possible.

@relu91 
>"When one form is chosen, it is expected that the consumer will continue to use it as long as possible for every new interaction with the Producer"

+1

index.html
sebastiankb: done

index.html
sebastiankb: done

index.html
zolkis: Yes, but "the first form entry that works for them" is equivalent to "any form entry".
Why is it a problem if the implementation chooses the form? (the one best matching its own context).
If we have generic guidelines for specific cases, we should be explicit about them, like when choosing between protocols, which would need defining a priority list for protocols (just an example).

index.html
zolkis: Still looks the same, don't forget you need to fix these comments in 2 places (copy-paste) - and I see you fixed it one place, not here. You can amend the previous commit I guess.

sebastiankb: sorry, I forgot the render script

sebastiankb:
updated PR based on feedback from @zolkis and @RealJoshue108

zolkis (approved):
Looks good.

The language is not concise, but that needs to be checked across the whole document.
For instance, instead of "it is expected" it would be nice to specify MAY (free to do or not to do) or SHOULD (recommended in all cases minus specific exceptions) or MUST (it is a compliance error not to do so).

sebastiankb:
from today's TD call we decided to include this PR

Issue #1068, opened 2021-03-10
uriVariables needs to be more limited


sebastiankb:
`uriVariables` should only typed with simple data types such as string and number. Complex types like arrays and object should not allowed. We need an assertion about this.

relu91:
Another thing that is confusing about `uriVariables` is the ability to use them also in action affordances. There, we have already the `input` field which specifies the required arguments of an action. I think we should rather rely on the input field and specify that the URI template can refer to first-level properties in the input data schema. For example, consider this odd Web Thing that has an action that takes in input id as URI variable and a payload:
```js
"configure":{
            "title": "configure",
            "description": "",
            "input": {
                "title": "payload",
                "type": "object",
                "properties": {
                    /**/
                }
            },
            "uriVariables": {
                "id":{
                    "type": "string"
                }
            },
            "output": {},
            "forms": [{
                "href": "http://example.com/subdevice{?id}"
            }]
        }
```

Essentially we could simplify the affordance like this: 
```js
"configure":{
            "title": "configure",
            "description": "",
            "input": {
                "title": "payload",
                "type": "object",
                "properties": {
                    "id": {
                        "type": "string"
                    }
                    /**/
                }
            },
            "uriVariables": {
                "id":{
                    "type": "string"
                }
            },
            "output": {},
            "forms": [{
                "href": "http://example.com/subdevice{?id}"
            }]
        }
```

A runtime just needs to know which property of the input should be inserted in the URI. Conflicts between the payload and the URI template can be easily resolved by using a different name. 

Regarding other interaction affordances (i.e., properties and events), having `uriVariables` it's a strong hint for a bad design choice as they are probably actions. I've not found anything on the document describing this possible issue. I think we should consider also this problem when talking about https://github.com/w3c/wot-thing-description/issues/1020.

egekorkan:
> Regarding other interaction affordances (i.e., properties and events), having uriVariables it's a strong hint for a bad design choice as they are probably actions.

The main use case I find for `uriVariables` in properties or events is to filter a key of an object, give an input for getting something or to simply describe already existing Web APIs. I think that the first two points are not really RESTful designs but they are what literally every weather API does... Examples of weather APIs below:
- https://openweathermap.org/current
- https://www.weatherbit.io/api/weather-current
- https://developer.accuweather.com/accuweather-forecast-api/apis : This one is actually not `uriVariables` but `uriTemplates` that we cannot describe yet in TDs, I think?

[...]

PR #1109, opened 2021-04-23
restrict more uri variables

sebastiankb:
also see #1068

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1109.html" title="Last updated on Apr 28, 2021, 9:46 AM UTC (767f5fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1109/24c2bf8...767f5fc.html" title="Last updated on Apr 28, 2021, 9:46 AM UTC (767f5fc)">Diff</a>


egekorkan:
Thanks for fixing the weird numbers that were popping up as well :)

I think that the text needs a tiny bit of polishing:

> Define URI query template variables as collection based on DataSchema declarations. The type of the DataSchema cannot be an ObjectSchema or an ArraySchema. 

It is not exactly clear that the keys in the uriVariables map cannot be of ObjectSchema or ArraySchema, i.e. it reads as if uriVariables cannot be an object, it always is. Not sure but maybe something like:

->  Define URI query template variables as a collection based on DataSchema declarations. The individual variables' (keys') DataSchema cannot be an ObjectSchema or an ArraySchema. 

There was also a missing `a` before collection :)

index.html
```diff
@@ -3640,6 +3641,12 @@ <h3><code>forms</code></h3>
 }
 </pre>
 
+<p>
+    <code>uriVariables</code> are mainly for properties and events. When retrofitting an existing system, it may be necessary to use
+    <code>uriVariables</code> for actions. In general, it is recommended to avoid <code>uriVariables</code> as much as possbile when a 
+    new WoT-based system is designed.
```
relu91, line 3647: Thanks for making this an explicit statement of the specs.

relu91 (approved):
Nicely done, except for @egekorkan's comment the PR looks good. Just a side note, in #1068 I mentioned the possibility to merge `uriVariables` in the `input` object for actions avoiding ambiguities.  It is fine that it is not addressed in this PR, but I would open a dedicated issue so that we can discuss this possibility. Maybe it could even be deferred for 2.0. What do you think?

index.html
sebastiankb: year, that would be a kind of bigger change and should be more addressed in the TD 2.0 version. I would suggest that you create a new issue and I label it with defer 2.0

sebastiankb:
new commit includes the CR from Ege

sebastiankb:
from today's TD call it was decided to merge this PR

PR #1112, opened 2021-04-28
Adding `tm:ref` and `tm:required` to tm schema generation

egekorkan:
I have worked on the TM Schema generator to include `tm:ref` and tm:required. Addresses #1065

It is marked as Draft since I am not sure about two aspects:

1. Where can one use `tm:ref`? Everywhere or only in an interaction? The examples are only for interactions.
2. I am not sure if JSON Pointer is the correct type for `tm:ref`. If I do the following validation at https://www.jsonschemavalidator.net , it does not accept that string as JSON Pointer:

```js
{
            "type":"string",
            "format": "json-pointer"
}
```
versus: `"http://example.com/BasicOnOffTM.tm.jsonld#/properties/onOff"`


egekorkan:
Oops I have committed some wrong stuff...

PR #1113, opened 2021-04-28
add tm:required and tm:ref generation

egekorkan:
I have worked on the TM Schema generator to include `tm:ref` and tm:required. Addresses #1065

It is marked as Draft since I am not sure about two aspects:

1. Where can one use `tm:ref`? Everywhere or only in an interaction? The examples are only for interactions.
2. I am not sure if JSON Pointer is the correct type for `tm:ref`. If I do the following validation at https://www.jsonschemavalidator.net , it does not accept that string as JSON Pointer:

```js
{
            "type":"string",
            "format": "json-pointer"
}
```
versus: `"http://example.com/BasicOnOffTM.tm.jsonld#/properties/onOff"`


sebastiankb:
>Where can one use tm:ref? Everywhere or only in an interaction? The examples are only for interactions.

good question. I would be open to have it also at top level to include existing metadata definition. 

> I am not sure if JSON Pointer is the correct type for tm:ref.

I checked with the JSON Schema definition [here](https://json-schema.org/understanding-json-schema/structuring.html) and you are right:
>The value of $ref is a URI-reference, and the part after # sign (the “fragment” or “named anchor”) is in a format called JSON Pointer.

So tm:ref is consisting of two patterns: an (optional) URI to external TM definition + JSON Pointer definition

We should update this in the spec. @egekorkan can you take this in this PR?

sebastiankb:
from today's TD call:
- group agree to have tm:ref everywhere
- however, we should be careful that the TM is still valid regarding imported definitions that is not causes some redundant definitions (e.g., defining title and import title again)
- we have to introduce a good explanation how the value of "tm:rel" is to interpret. we should also check the definition from sdfRef

egekorkan:
Before I forget for the next week, is `tm:required` restricted to `#/properties/`, `#/actions/`, `#/events/` ? I have assumed so and did the following in the PR: https://github.com/w3c/wot-thing-description/pull/1113/files#diff-82a902d1c513889daba3558ad2384eaac05f7b5c3c03167129951b9451e53857R1090

sebastiankb:
>Before I forget for the next week, is tm:required restricted to #/properties/, #/actions/, #/events/ ? I have assumed so and did the following in the PR:

I would support the idea that this is not limited to interactions. I might want to enforce that TD instances should also use a version container, etc.

egekorkan:
I have removed the pattern keyword that needed the values of `tm:ref` to start with `#/properties` etc.

I still need to add the part that handles ` group agree to have tm:ref everywhere `

egekorkan:
Now it is ready for review

sebastiankb:
from today's TD call we will merge this PR

PR #1103, opened 2021-04-22
Introduce Thing Icon

sebastiankb:
also see #1063

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1103.html" title="Last updated on Apr 28, 2021, 3:02 PM UTC (aefb0b5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1103/98f98fd...aefb0b5.html" title="Last updated on Apr 28, 2021, 3:02 PM UTC (aefb0b5)">Diff</a>

PR #1120, opened 2021-04-30
Add validation for icon links

relu91:
#1103 introduced the icon relation type for TD links together with a new attribute called `sizes`. The current specs says that the `sizes` is only allowed for links that have ` "rel":"icon"` but the JSONSchema was not updated. This PR updates the JSON  Schema and validate the `sizes` field using the proposed format:
> The value pattern follows {Height}x{Width} (e.g., "16x16", "16x16 32x32").

Thanks, @egekorkan for reviewing the proposal 👍🏻 .

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -766,12 +766,12 @@
                 },
                 {
                     "not": {
-                        "description": "A basic link element should not contain icon or extends",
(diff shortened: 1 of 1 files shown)


sebastiankb (approved):

egekorkan (approved):

sebastiankb:
from today's td call we decided to merge

PR #1129, opened 2021-05-07
More fixes to canonicalization

mmccool:
- Discuss use of base.  In particular, add a clarification that the previous assertion about not changing the expression of URLs *includes* not changing how URLs used for interaction endpoints are expressed as a combination of base and relative URLs.  This is also meant to address the point about base raised in https://github.com/w3c/wot-profile/pull/57
- Add an assertion stating that if a prefix is defined, it MUST be used in place of that URL.  This is to support stable round-tripping where we may expand prefixes into URLs and then on export have to "un-expand" them.  If a URL was used in place of a prefix in the input it would be a legal TD, but the unexpansion would replace it with the prefix, which would result in a serialization different than the input.
- <s>Update: No "security" at the Thing level, and added assertion that names of securityDefinitions must be preserved.  This avoids problems with "equivalent" TDs using different names for these definitions.  See comments below.</s> (REVERTED)

Note that some of these choices mean that there will be Canonical TDs defining the same "information" that are considered different.  For instance, "not changing the way URLs are expressed" means that TDs that split up base and (relative) interaction URLs differently will be considered different, even though they actually refer to the same endpoints.  Likewise not changing prefixes means we can have any number of TDs with different prefixes that refer to the same URLs for those vocabularies, and therefore technically have the same semantics.  Therefore the goal is not really to stably canonicalize the semantic information itself, but the *expression* (serialization) of a given set of information.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1129.html" title="Last updated on May 28, 2021, 1:27 PM UTC (ec5c17b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1129/f95085b...mmccool:ec5c17b.html" title="Last updated on May 28, 2021, 1:27 PM UTC (ec5c17b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4264,7 +4264,7 @@ On the <a>Thing</a> side, <a>Thing</a> is expected to return readable (non <code
   </p><p>
   Some of the choices leading to the efficient computation of a canonical form, 
   such as the suppression of white space and sorting of elements in objects 
-  do make the representation less human readable.
+  do make the representation less human readable, more verbose, and more redundant.
   In general, the canonical form presented here is optimized for machine-to-machine
   use cases such as signing.
   </p>
@@ -4303,6 +4303,17 @@ On the <a>Thing</a> side, <a>Thing</a> is expected to return readable (non <code
   Note that this also applies to extension vocabularies, e.g. for protocol bindings.
   If any such extension defines default values they must be given explicitly in the canonical form.
   </li>
+<!--
+  <li><span class="rfc2119-assertion" id="canonical-no-base">
+  In the <a>Canonical TD</a>,
+  <code>base</code> MUST NOT be used.</span>  
+  All URLs need to be given in absolute form.
+  </li>
+-->
+  <li><span class="rfc2119-assertion" id="canonical-prefix-mandatory">
+  In the Canonical TD, if a prefix is
+  defined it MUST be used in place of that URL.</span>
+  </li>
   <li><span class="rfc2119-assertion" id="canonical-singleton-arrays">
   In the <a>Canonical TD</a>,
   all values that can be expressed as either an array or as a single value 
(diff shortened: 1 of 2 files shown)


mmccool:
just realized there there is one other issue with "security" at the Thing level vs. at the Form level.  A TD with the same security (say, "nosec") in all Forms can be expressed in two ways, in fact many more than two.  So to avoid variants, will have to add an assertion that Canonical form expresses all security requirements at the Form level (resolving all overrides first) and does not use "security" at the Thing level.

mmccool:
Don't know if I have to add something about the names of securityDefintions being retained.  In *theory* you could define the same information using different names for securityDefinitions, as long as the mapping was the same (these are flattened before being pulled into an RDF database and some extra work is needed to unflatten them when reserializing, too).   Probably to be safe I should add another assertion that forbids processors from changing these names.

mmccool:
Note that we can make the names of defined objects part of the information model to avoid the disconnect between information and its expression with this last point (and for prefixes).  See the round-tripping discussion linked above.

mmccool:
@sebastiankb although you changed the title to WIP, I feel this PR is ready to merge myself (pending review).  Of course I MAY have overlooked other little corner cases for canonicalization (the nature of the problem...) but don't have any more right now.

mmccool:
Note there is a similar issue with additionalExpectedResponses and schemaDefinitions wrt to canonicalization and not changing the names of schema definitions.  However, I can't add such an assertion here until PR https://github.com/w3c/wot-thing-description/pull/1130 is merged, and can't add it to https://github.com/w3c/wot-thing-description/pull/1130 until this PR is merged.  So we should merge both and then I can fix this gap.  I've created an issue to follow up on this: https://github.com/w3c/wot-thing-description/issues/1150

index.html
```diff
@@ -4403,20 +4415,55 @@ <h2>Canonicalization</h2>
   <span class="rfc2119-assertion" id="processors-preserve-prefixes">
   In order to support canonicalization, implementations 
   of TD Processors MUST preserve the prefixes used for extension vocabularies.</span>
-  This includes preserving and expressing a vocabulary term as a URL if no prefix was used.
+  This includes preserving and expressing a vocabulary term as a URL 
+  if no prefix is defined for that URL.
+  However, we need to
+  support stable reserialization in TD Processors that expand all
+  prefixes into URLs and then need to collapse them back to prefixes
```
sebastiankb, line 4422: we need to grantee that normalized prefixes are used

sebastiankb (changes requested):

index.html
```diff
@@ -4330,7 +4330,7 @@ <h2>Canonicalization</h2>
   </p><p>
   Some of the choices leading to the efficient computation of a canonical form, 
   such as the suppression of white space and sorting of elements in objects 
-  do make the representation less human readable.
+  do make the representation less human readable, more verbose, and more redundant.
```
danielpeintner, line 4333: This addition looks contradictory to me. Why should sorting and/or suppressing white-spaces make it more verbose and redundant?

index.html
```diff
@@ -4369,6 +4369,18 @@ <h2>Canonicalization</h2>
   Note that this also applies to extension vocabularies, e.g. for protocol bindings.
   If any such extension defines default values they must be given explicitly in the canonical form.
   </li>
+  <li><span class="rfc2119-assertion" id="canonical-form-security">
+  In the <a>Canonical TD</a>,
+  <code>security</code> MUST NOT be used at the <code>Thing</code> level.</span>
```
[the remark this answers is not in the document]
mmccool: OK, let me look into that.   You're right, we changed this; it is mandatory at the top level, and optional at the form level.  So the only thing we have to worry about is people "overriding" it at the Form level with the same scheme mentioned at the top level (which would be semantically equivalent, but redundant...)

index.html
mmccool: yes, maintaining prefixes is a TD processor requirements (this includes directories).  This can be done though by treating the @context as a regular object...

sebastiankb:
from today's TD call:
- PR not ready to merge
- @mmccool checks the comments above
- step 5 have to be removed. The solutions should be the same as for the securityDefintions.

index.html
mmccool: I will try rewording that.  Suppression of whitespace does not make it more verbose, but other choices may.   Or I may just delete that addition.

index.html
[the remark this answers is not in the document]
mmccool: There are two possible solutions here:
1. Leave the definitions alone.  If "security" appears that the Form level, it must appear in the output after round-tripping, even if it is the same as the object at the "Thing" level.  We would have to forbid TD processors from "optimizing" TDs by removing such redundant definitions.
2. Require removal of redundant definitions.
Based on the discussion, it sounds like people would prefer 1.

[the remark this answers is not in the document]
mmccool: I applied solution 1, adding an assertion that TD processors must not remove (or add) "security" fields in Forms, even if they are redundant.

index.html
mmccool: decided to just take those extra words out.

index.html
mmccool: I *think* this was already taken care of in the draft, please confirm and resolve the conversation if so.

sebastiankb (approved):

sebastiankb:
from today's TD call:
- @mmccool addressed the comments above in latest commit
- decide to merge to have a new baseline about canonicaliziation

PR #1155, opened 2021-06-02
Draft Testing Report for TD 1.1

mmccool:
PR for new TD 1.1 test report, after June 2021 Testfest.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1155.html" title="Last updated on Aug 4, 2021, 1:46 PM UTC (9563f25)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1155/4e1496e...mmccool:9563f25.html" title="Last updated on Aug 4, 2021, 1:46 PM UTC (9563f25)">Diff</a>


mmccool:
So while updating this recently I had the following thoughts:
- We probably DO want to make a PR to merge eventually, e.g. with the draft implementation report.  The aim is to get that done by June 18 (and probably have a prelim version ready by June 11)
- We previously agreed to test the WD version, but I accidentally rebased this PR against main, which is however (at this point) very close to the WD version.   We can either do two reports (e.g. against main, which we can keep up to date, and against the WD) or focus on the WD version and then update to main after the Testfest.
- The report currently includes results from TD1.0 which includes implementations from some vendors that have not been active recently, but whose implementations may still be valid and usable as test inputs.  We will have to decide how to deal with those.  Since TD 1.1 is supposed to be mostly compatible with TD1.0, TD1.0 inputs are still valuable test cases.

mmccool:
Decided to go ahead and use this PR to actually submit a draft implementation report when it's ready (which is not yet).

sebastiankb:
from today's TD call:
- before merge we should use a new report file like report11.html otherwise we overwrite the 1.0 report version

mmccool:
Ready to merge.  Updates to better handle versioning:
- Archive of old report (and all input files) copied to publication/rec/testing from wot-td-1.0 branch
- Copy of HTML file for report copied to testing/report.html so existing URL in TD1.0 REC points to the right file
- Assertion testing script (assertion.js) that generates the report updated to output to testing/report11.html
- URL in index.template.html updated to point to report11.html
- render and assertion test report generator re-run, output to testing/output.log.  Note there are some errors in there that should be addressed (missing RFC2119 classes in some assertions, for example)

sebastiankb:
from today's TD call, we decided to merge this PR

PR #1157, opened 2021-06-03
Fix ContentEncoding and ContentMediaType

### LINKED ISSUE #1116: contentEncoding and contentMediaType should be moved under string schema
While not being 100% sure, I think that `contentEncoding` and `contentMediaType` should be moved to strings. They do not make sense for an object or array. Maybe number or integer can be used but I doubt it. JSON Schema spec is slightly ambiguous about it but all the examples are with strings

https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-8

relu91:
This PR fixes #1116. It moves `contentEncoding` and `contentMediaType` under `StringSchema` class. They were already placed currently in the ontology but not in the shacl file.

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-thing-description/pull/1157.html" title="Last updated on Jun 17, 2021, 7:11 AM UTC (eb024f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1157/075fcc6...relu91:eb024f8.html" title="Last updated on Jun 17, 2021, 7:11 AM UTC (eb024f8)">Diff</a>


validation/td-validation.ttl
```diff
@@ -1728,6 +1714,20 @@
...
+        sh:maxCount 1 ;
+        sh:order 14 ;
+    ] ;
+    sh:property [
+        sh:path jsonschema:contentMediaType ;
+        skos:definition """Specifies the MIME type (e.g., image/png, audio/mpeg) of the contents of a string value, as described in RFC 2046."""^^rdf:HTML ;
+        sh:datatype xsd:string ;
+        sh:maxCount 1 ;
+        sh:order 15 ;
+    ] ;
```
danielpeintner, line 1730: I think those 2 properties should come after jsonschema:pattern.
see https://github.com/w3c/wot-thing-description/blob/4e25ce3fe80dde6eb613f37dd7f8aa39fea24057/validation/td-validation.ttl#L1731

I would also change the sh:order to 4 and 5 so that it matches the sequence...

danielpeintner (changes requested):
Other than my minor re-ordering/clean-up proposal it looks good!

egekorkan:
The preview looks good to me :)

validation/td-validation.ttl
relu91: Right! fixed 👍🏻

danielpeintner (approved):
LGTM

sebastiankb:
from today's TD session:
- group decided to merge this PR
- just small fixes: the list of  \"7bit\", \"8bit\", \"binary\", \"quoted-printable\" and \"base64\" should be as an example and provided by the string type.

relu91:
>just small fixes: the list of "7bit", "8bit", "binary", "quoted-printable" and "base64" should be as an example and provided by the string type.

@sebastiankb  Should be fixed now!

sebastiankb:
many thanks! as agreed yesterday, I'm going to merge this PR.

PR #1158, opened 2021-06-04
Proposal: use version field in schemas

relu91:
Currently, the schema version is contained in the `title` field. This results in weird names when using schemas for the generation of TypeScript types. It would beneficial to have a clean name and a dedicated field just for the version. I found no official way to handle this, but the approach that I choose is similar to what npm is doing for package.json files.


egekorkan (approved):

sebastiankb:
from today's TD call:
- group decided to merge this PR
- we should be always remember that we change version in the schemas

PR #1163, opened 2021-06-07
add success to addResp and refactor addResp into definitions

egekorkan:
the `success` term was missing and the `additionalRepsponses` were copy pasted instead being referenced in the definitions


sebastiankb:
from today's TD session, group decided to merge this PR

PR #1168, opened 2021-06-09
fix: add wotsec context on securityDefinitions property

ethieblin:
Should fix the context to obtain the same result as the RDF example #1167


vcharpenay (approved):
the next TD version will likely change how `securityDefinitions` should be processed in RDF, but this fix is satisfactory until then.

PR #1196, opened 2021-07-20
Add social media icons to Readme

egekorkan:
Following https://github.com/w3c/wot-marketing/issues/188

Ideally, I would like to repeat this in our other repositories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 <div align="center">
   
 [![Follow on Twitter](https://img.shields.io/twitter/follow/W3C_WoT.svg?label=follow+W3C_WoT)](https://twitter.com/W3C_WoT)
-<img alt="Stack Exchange questions" src="https://img.shields.io/stackexchange/stackoverflow/t/web-of-things?style=plastic">
+[![Stack Exchange questions](https://img.shields.io/stackexchange/stackoverflow/t/web-of-things?style=plastic)]( https://stackoverflow.com/questions/tagged/web-of-things)
   
 </div>
 
--- a/README.md
+++ b/README.md
@@ -1,14 +1,10 @@
-# Specification 'Web of Things (WoT) Thing Description'
+# Web of Things (WoT) Thing Description
 
-<div align="center">
-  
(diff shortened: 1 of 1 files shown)


danielpeintner (approved):
LGTM.

The idea from @mmccool  was to add at the same time some general information link .. see *my* example below.

### ~~~ Possible Proposal ~~~

General information about the Web of Things can be found on https://www.w3.org/WoT/.

<div align="center">

[Follow on Twitter](https://twitter.com/W3C_WoT)
Stack Exchange questions

</div>

README.md
```diff
@@ -1,5 +1,12 @@
 # Specification 'Web of Things (WoT) Thing Description'
 
+<div align="center">
+  
+[![Follow on Twitter](https://img.shields.io/twitter/follow/W3C_WoT.svg?label=follow+W3C_WoT)](https://twitter.com/W3C_WoT)
+<img alt="Stack Exchange questions" src="https://img.shields.io/stackexchange/stackoverflow/t/web-of-things?style=plastic">
```
JKRhb, line 6: Very cool addition to the readme :) I was wondering, however, if the badge should also link to Stackoverflow (for example https://stackoverflow.com/questions/tagged/web-of-things). If I am not mistaken, clicking on the badge just opens the SVG in another tab.

danielpeintner: @JKRhb you are right, a surrounding `<a>` is needed to work as expected.

JKRhb: Using Markdown syntax I guess this could work?

```suggestion
[![Stack Exchange questions](https://img.shields.io/stackexchange/stackoverflow/t/web-of-things?style=plastic)]( https://stackoverflow.com/questions/tagged/web-of-things)
```

egekorkan:
@JKRhb and @danielpeintner thank you for the contributions. I have added both of this information as a small section above the old text and separated them with a line. It looks pretty good now

mmccool:
Do think the icons should be left-aligned rather than centered, it's less messy.

mmccool:
C - contrast
R - repetition
A - alignment <--> left alignment
P - proximity

egekorkan:
In the call of 21.07, we have decided to:
- We will merge this
- Other repositories should follow suit
- @relu91 proposed adding the logo as well. This will be discussed later.

Issue #1186, opened 2021-07-06
Issues with Thing Model schema


JKRhb:
At the moment there are no `"required"` fields in the JSON Schema for Thing Models. However, the current draft states that

> Thing Model definitions MUST use the keyword @type at top level and a value of type string or array that equals or respectively contains tm:ThingModel

which is not reflected by this approach (a TM without a top-level `@type` field is valid at the moment).

I am also wondering if it could make sense to also declare `@context` and maybe `title` mandatory.

egekorkan:
`@type` not being mandatory is a bug that I will fix. Regarding making the other fields mandatory, I remember discussing that everything required should be removed. However, the text reads as if `@context` should be mandatory.

JKRhb:
I just noticed another TM related bug in the schema: Specifying a `@type` for an affordance somehow does not work at the moment except when the value is `tm:ThingModel". I created a minimal example illustrating the issue here: https://www.jsonschemavalidator.net/s/HUCDN9HO

JKRhb:
I discovered another issue with the TM schema: At the moment it is not possible to use `rel` in a link object. See the following minimal example: https://www.jsonschemavalidator.net/s/yFlOOPVN

JKRhb:
`multipleOf` seems not to be working as well. The following definition is not valid at the moment:

```json
{
    "@context": ["http://www.w3.org/ns/td"],
    "@type": "tm:ThingModel",
    "properties": {
        "foo": {
            "title": "Invalid",
            "multipleOf": 2
        },
        "bar": {
            "title": "Invalid",
            "multipleOf": 2.0
        },
        "baz": {
            "title": "Valid, because not an integer",
            "multipleOf": 2.2
        }
    }
}
```

Also see: https://www.jsonschemavalidator.net/s/lX5pEnaQ

PR #1204, opened 2021-07-30
Fix definition of multipleOf in TD schema

JKRhb:
In #1186 I noticed (among other issues) that the current definition of multipleOf in the JSON schemas does not work as expected. So far the following (TM) definition would be invalid:

```json
{
    "@context": ["http://www.w3.org/ns/td"],
    "@type": "tm:ThingModel",
    "properties": {
        "foo": {
            "title": "Invalid",
            "multipleOf": 1
        },
        "bar": {
            "title": "Invalid",
            "multipleOf": 2.0
        },
        "baz": {
            "title": "Valid, because not an integer",
            "multipleOf": 2.2
        }
    }
}
```

This PR replaces the `oneOf` in the definitions of `multipleOf` with an `anyOf` and adds an `exclusiveMinimum` of 0 to align the schema more closely to the [specification](https://w3c.github.io/wot-thing-description/#td-vocab-multipleOf--NumberSchema) that states:

> The value must [be] strictly greater than 0.

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -207,16 +207,7 @@
                     "minimum": 0
                 },
                 "multipleOf": {
-                    "anyOf": [
-                        {
-                            "type": "integer",
-                            "exclusiveMinimum": 0
-                        },
-                        {
-                            "type": "number",
-                            "exclusiveMinimum": 0
-                        }
-                    ]
+                    "$ref": "#/definitions/multipleOfDefinition"
                 },
                 "properties": {
                     "additionalProperties": {
@@ -247,6 +238,18 @@
                     }
                 }
             }
+        },
+        "multipleOfDefinition": {
+            "anyOf": [
+                {
+                    "type": "integer",
+                    "exclusiveMinimum": 0
+                },
+                {
+                    "type": "number",
+                    "exclusiveMinimum": 0
+                }
+            ]
         },
         "form_element_property": {
             "type": "object",
@@ -604,16 +607,7 @@
                     "minimum": 0
                 },
                 "multipleOf": {
-                    "anyOf": [
-                        {
-                            "type": "integer",
-                            "exclusiveMinimum": 0
-                        },
-                        {
-                            "type": "number",
-                            "exclusiveMinimum": 0
-                        }
-                    ]
+                    "$ref": "#/definitions/multipleOfDefinition"
                 },
                 "properties": {
                     "additionalProperties": {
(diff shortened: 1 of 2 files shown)


egekorkan (approved):

egekorkan:
Thanks! :)

JKRhb:
> Thanks! :)

Thank you for the approval! :)

I am also adding some (potential) fixes to the TM generation script at the moment and would soon open a PR that maybe requires a bit more discussion in the working group. I hope this approach is okay for you, otherwise let me know :)

danielpeintner:
@egekorkan I am just wondering.... do we have some kind of tests?
e.g., JSON instances that fail/pass w.r.t. the TD/TM schema?
I think having such (automatic) tests would be *very* beneficial

relu91 (approved):
By the way, the PR does also a small refactoring of the `additionalResponses` term. Now the term refers to a global definition called `additionalResponsesDefinition` (which is good). Well done!

egekorkan:
@danielpeintner wrote:

> @egekorkan I am just wondering.... do we have some kind of tests?
e.g., JSON instances that fail/pass w.r.t. the TD/TM schema?
I think having such (automatic) tests would be very beneficial

So I do have tests for TDs at playground at: https://github.com/thingweb/thingweb-playground/tree/master/packages/core/examples  . These are executed at each PR. Ideally, each change to the TD schema is reflected there and tested automatically.

I will also add TM tests since obviously there are wrong things happening everywhere.

JKRhb:
> By the way, the PR does also a small refactoring of the `additionalResponses` term. Now the term refers to a global definition called `additionalResponsesDefinition` (which is good). Well done!

I think this has actually already been part of the TD schema and is only integrated in this PR because of the regeneration of the TM schema so I can't take credit for that ;) Speaking of which, however, I was thinking about refactoring `multipleOf` as a global definition. Would this be a useful addition to this PR?

JKRhb:
> I was thinking about refactoring `multipleOf` as a global definition. Would this be a useful addition to this PR?

I added such a refactoring now in 902c813a7c6280fc40dd522044e2e7eb783e5398

sebastiankb (approved):

sebastiankb:
from today's TD call:
- we decided to merge this PR, @JKRhb many thanks
- @JKRhb are you interested for applying as invited expert? Would be cool ;-)

JKRhb:
> * @JKRhb are you interested for applying as invited expert? Would be cool ;-)

Thank you very much for this opportunity! :) I will hand in an application soon :)

PR #1205, opened 2021-07-30
Fix issues in TM schema generation script

### LINKED ISSUE #1186: Issues with Thing Model schema

JKRhb:
This PR tries to deal with a couple of issues that I have identified in #1186.  The concrete issues this PR solves are the following:

* The TM specific `@type` definition is now only used at the top level
* `@context` and `@type` are now required fields at the top level
* Using `rel` and `sizes` in `links` is now possible as well as `{"rel": "tm:extends"}`
* It is now possible to define `securityDefinitions` which led to invalid schemas using the old approach

To realize the last bullet point I commented out the `removeEnums` function which prevented the definition of any `SecurityScheme`. In my opinion all enums should actually still be included in TMs as otherwise a lot of value of the TM validation is lost (you could define a TM that will become an invalid TD after instantiation). I guess this aspect of the PR, however, should be discussed by you in the working group before going forward with this PR.

The PR does not include a newly generated schema yet. I would wait for this step until #1204 has been merged.

Once this PR is merged, it closes #1186.


validation/tmSchemaGenerator.js
```diff
@@ -57,7 +57,7 @@ let tdSchema = JSON.parse(fs.readFileSync('validation/td-json-schema-validation.
 // do all the manipulation in order
 let tmSchema = staticReplace(tdSchema)
 tmSchema = removeRequired(tmSchema)
-tmSchema = removeEnum(tmSchema)
+// tmSchema = removeEnum(tmSchema)
```
egekorkan, line 60: not sure why this is commented out

validation/tmSchemaGenerator.js
```diff
@@ -79,6 +79,18 @@ function staticReplace(argObject){
     argObject.title = "Thing Model"
     argObject.description = "JSON Schema for validating Thing Models. This is automatically generated from the WoT TD Schema."
     argObject.definitions.type_declaration = {
+        "oneOf": [{
```
egekorkan, line 82: normal type declaration is not needed anymore right?

validation/tmSchemaGenerator.js
```diff
@@ -110,9 +122,14 @@ function removeRequired(argObject) {
     // remove required if it exists and is of array type.
     // check for array is needed since we also specify what a required is and that it is an object
     if (("required" in argObject) && (Array.isArray(argObject.required))){
-        // need to decide whether to delete or replace it with ""
-        // delete is "cleaner" but "" is more explicit
-        delete argObject.required;
+        if (!(argObject["required"] == "sizes" || argObject["required"] == "rel")) {
```
egekorkan, line 125: Could you add some comments on what is happening here?

egekorkan (changes requested):

validation/tmSchemaGenerator.js
JKRhb: I actually separated the top level `@type` definition from the one for other levels because otherwise you have to use `tm:ThingModel` in `@type` fields of every level. This new approach, however, removes, allows you to use `tm:ThingModel` in *any* `@type` field. If this should be kept as is we could also just use the existing definition from the TD schema.

validation/tmSchemaGenerator.js
JKRhb: Sure! Will do :)

validation/tmSchemaGenerator.js
JKRhb: I commented it out only for now as I wasn't sure if you want to keep this or some other kind of enum removal in the schema generation. I would propose keeping the enum values in but maybe this is something you want to discuss in the working group first.

validation/tmSchemaGenerator.js
JKRhb: I added a comment in c7b5fe5cbe9c24ad76c96c89a8fe15c305e53fb6

JKRhb:
To make the changes to the generation script more transparent I added a preliminary new version of the TM schema in 8874ec279f3fbc129bba54fbc659d7a654387580. After all changes are made, this step should be repeated.

validation/tmSchemaGenerator.js
JKRhb: After thinking about it, one purpose of the enum removal is probably to enable the use of placeholders, I guess? Maybe the enums could be turned into an `anyOf` instead that additionally contains a regex item with a pattern like `"{{PLACEHOLDER}}"`? (Or rather, more formal, `\{{2}[A-Za-z]+\}{2}`).

JKRhb: > After thinking about it, one purpose of the enum removal is probably to enable the use of placeholders, I guess?

An example that illustrates the current problem would be the following: https://www.jsonschemavalidator.net/s/qFot0kvR

validation/tmSchemaGenerator.js
egekorkan: Oh I see, what you did is indeed correct :)

validation/tmSchemaGenerator.js
JKRhb: > I will try my take on this and commit to this branch

Awesome, thank you!

sebastiankb:
from today's TD call the group agrees to merge this PR

@egekorkan and  @JKRhb many thanks

egekorkan (approved):

Issue #1173, opened 2021-06-10
Unclear assertion on TMs


egekorkan:
I do not understand the following assertion:

>  If it is intended to override an existing JSON name-value pair definition from tm:ref, the same JSON name MUST be used at the same level of the tm:ref declearation that provides a new value. 

The example provided overwrites a property it references, which is allowed. Am I not allowed to reference a property and overwrite a sub element of its dataschema? Or is it about referencing a property inside an action? In any case, this limitation is not very clear.

sebastiankb:
Thanks for the hint. In the next update I will be more clear. In addition I will also have an illustration to make the overwriting more clear.

sebastiankb:
PR was merged today

Issue #1165, opened 2021-06-08
Minor spelling and grammar errors


ajs124:
I was reading chapter 10 and noticed a typo in §10.4: "placholder" should be "placeholder".

Also in §10.1, 
```
The Thing Model enables:
- onboard and manage […]
- simulate of […]
- develop common […]
- combining multiple […]
- implementation support […]
```
I'm not a linguist and English isn't my first language, so I lack the vocabulary to describe what exactly is wrong with this, but it would be news to me if this is correct. I really don't think something can "enable onboard and manage" or "enable develop".

Please tell me if this is the wrong way to report this kind of thing, if you don't care, etc.

sebastiankb:
Thanks for pointing that out and yes, that is the right place. I will take care in the next update.

Issue #1152, opened 2021-05-27
tm:ref pattern should be more clear


sebastiankb:
So far, `tm:ref` only [say](https://w3c.github.io/wot-thing-description/#thing-model-extension-import) it is a JSON Pointer. However, it can be a combination of an URL (location of the Thing Model) and JSON Pointer (to address position within the Thing Model definition)

sebastiankb:
PR was merged today

PR #1207, opened 2021-08-04
Updates for TM Chapter

sebastiankb:
This PR addresses:

- [x] removing basic text about TM (text is provided in [Arch 1.1](https://github.com/w3c/wot-architecture/commit/36a8c29ec391438d85961f96633f9dabec1a9dfd))
- [x] clearification about overwriting
- [x] TM composition / subModels with links 
- [x] #1173
- [x] JSON Pointer + Path definition for tm:Ref #1152
- [x] typo fix (Thing model --> Thing Model)
- [x] addresses https://github.com/w3c/wot-thing-description/issues/1252
- [x] addresses #1165

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1207.html" title="Last updated on Nov 3, 2021, 3:46 PM UTC (b0c36a8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1207/4e1496e...b0c36a8.html" title="Last updated on Nov 3, 2021, 3:46 PM UTC (b0c36a8)">Diff</a>


egekorkan:
Could you also address #1173 in this PR?

JKRhb:
> Could you also address #1173 in this PR?

Maybe [RFC 7386](https://datatracker.ietf.org/doc/html/rfc7386) (JSON Merge Patch) could be referenced in this context?

index.template.html
```diff
@@ -5009,6 +4990,38 @@ <h2>Required</h2>
...
+            "rel": "subthing",
+            "href": "./ventilation.tm.jsonld",
+            "instanceName": "ventilation"
+          },
+          {
+            "rel": "subthing",
+            "href": "./led.tm.jsonld",
+            "instanceName": "led"
+          }
+        ]
```
JKRhb, line 5020: @sebastiankb Thank you for adding a compose mechanism! I think this already looks promising. Some initial thoughts:

I was wondering if an approach similar to the one chosen by SDF (using objects and things in the TM itself) could be an alternative (or addition?) as well. Using a similar structure could make it easier to reuse models from SDF and therefore also from other ecosystems.

Also, I guess you would have to "flatten" the `subthing` links contained in a TM when converting such a TM to a TD, right? In this case, would the `instanceName` be used as a prefix of all affordances? And would top-level fields like the `@context` be merged into one? Or would the TM that is used for composition overwrite the values from the referenced TMs?

sebastiankb: Thanks for the feedback. Definitely, I would be open for evaluation. Can you share an example how you can refer to existing models without defining again within the object container?  Would this example work:

```
    {
        "@context": "http://www.w3.org/ns/td",
        "@type": "tm:ThingModel",
        "title": "Smart Ventilator Thing Model",
        "version" : { "model" : "1.0.0" },
        "tmObject": {
           "ventilation" :  { "tmRef": "./ventilation.tm.jsonld"}
            "led": {"tmRef": "./led.tm.jsonld"}
          }
      }
```

JKRhb: Very nice :) However, maybe we need to discuss if in this case the `tm:ref` should refer to the whole Thing Model or rather to another object within the TM (for example using `"ventilation" :  { "tmRef": "./ventilation.tm.jsonld#tmObject/ventilation"}`. I think both approaches could make sense. Having multiple `@context` fields in the case of referencing a complete TM wouldn't be a problem, would it?

sebastiankb: In my understanding, each object entry is an own Thing. Usually, for each Thing should be an own TM/TD. The original idea is to provide a way to describe that your system is based on a set of subsystems which each of them would have its own TD/TM definition. If TDs are generated this hierarchy can be provided by the links container again:

```json
"links": [{
     "rel": "item",
     "href": "path/to/ventilation.td.json"
 },
 {
     "rel": "item",
     "href": "path/to/led.td.json"
 },
```

sebastiankb:
TM Composition section is kind of ready. I would be happy to get some feedback.

index.html
```diff
@@ -6030,6 +6035,435 @@ <h2>Required</h2>
...
+                }
+            }
+           </pre>
+         </aside>
+
+
+    <p>
+        Such composition can be introduced by the usage of the <code>links</code> container: 
+        <span class="rfc2119-assertion" id="tm-compose-submodel">
+            A parent <a>Thing Model</a> MUST use at least one <code>links</code> entry with <code>"rel":"tm:submodel"</code>
```
thjaeckle, line 6147: Why is this defined as `MUST`? I am no expert in specifications, but as this is formulated here, this sounds very non-optional to me.
It may rather define an arbitrary amount of links, or do I miss something?

👍 for the `tm:submodel` - much clearer than `subthing`

index.html
```diff
@@ -6030,6 +6035,435 @@ <h2>Required</h2>
...
+            }
+           </pre>
+         </aside>
+
+
+    <p>
+        Such composition can be introduced by the usage of the <code>links</code> container: 
+        <span class="rfc2119-assertion" id="tm-compose-submodel">
+            A parent <a>Thing Model</a> MUST use at least one <code>links</code> entry with <code>"rel":"tm:submodel"</code> 
+            that targets to a (sub-) or child <a>Thing Model</a> that is be composed.
```
thjaeckle, line 6148: I would simplify to only `that targets to a sub Thing Model ...`. 
The "(sub-)" in parenthesis is confusing and introducing "child" as a new term here as well.

`that is be composed` sounds strange to me.

index.html
```diff
@@ -6030,6 +6035,435 @@ <h2>Required</h2>
...
+            Optional an <code>instanceName</code> can be provided to associate an individual name to the composed (sub-) <a>Thing Model</a>.
+        </span> 
+        This is useful when multiple same <a>Thing Model</a> definitions are composed and needs to be distinguised. 
+    </p>
+
+
+    <p>
+        Different strategies can be followed to generate <a>Thing Descriptions</a> from composed <a>Thing Model</a> definitions. The default recommendation is to generate from each
+         parent and sub/child <a>Thing Model</a> a corresponding <a>Thing Descriptions</a> (also see <a href="#thing-model-td-generation"></a>). 
+       The composition relation can be reflected by the <code>collection</code> and <code>item</code> relation types in the links container of the
```
thjaeckle, line 6160: is this really a "The composition relation `can` be reflected"? our should this be `SHOULD`?

thjaeckle:
> TM Composition section is kind of ready. I would be happy to get some feedback.

I added some inline comments to the composition section. In general I already really like and and it's good that you added the "inline TD" example as an option 👍

sebastiankb:
@JKRhb 
> Maybe RFC 7386 (JSON Merge Patch) could be referenced in this context?

done

Many thanks for the hint.

index.html
sebastiankb: Here I would keep it optional if Thing providers want to share what the relationship looks like with other Things during run time. This may also be avoided for security reasons.

sebastiankb:
> Is #1165 handled in this PR?

done

I will put this PR on tomorrow's TD agenda and ask for approval to merge.

egekorkan:
Actually, does this contradict with the following assertion?

> The links array can use the entry with "rel" : "type" maximum one.

If I use `type` it can be used only once in total and if I want to explicitly say that I am implementing multiple TMs, it should be with `tm:submodel`?

Also, I have realized that `tm:submodel` is not in the rel values table, thus requesting for changes

index.template.html
```diff
@@ -4855,91 +4856,460 @@ <h3>Extension and Import</h3>
...
 <p>
-    <a>Thing Model</a> can specify which terms should be taken in a TD instance, but their values are unspecific and are first known during TD instantiation. 
-    In such a case the placeholder labeling MAY be used in Thing Model that MUST be substituted with a concrete value when TD instance is created 
-    from the Thing Model. 
-    <span class="rfc2119-assertion" id="tm-placeholder">
-        The string-based pattern of the placeholder MUST follow <code>{{\PLACEHOLDER_IDENTIFIER}}</code>, 
-        which should contain a placeholder identifier name. The identifier name can be used to identify the placeholder for the substitution process.
+    Such composition can be introduced by the usage of the <code>links</code> container. 
+    <span class="rfc2119-assertion" id="tm-compose-submodel">
+        If it is desired to provide information that a <a>Thing Model</a> consists of one or more (sub-)<a>Thing Models</a>, the <code>links</code> entries MUST use the <code>"rel":"tm:submodel"</code>
```
egekorkan, line 4993: this is missing from the second table at https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1207.html#link

egekorkan (changes requested):

egekorkan (approved):
opening a new issue for my request for changes so that this PR can be merged

sebastiankb:
from today's TD call, group decided to merge

PR #1212, opened 2021-08-06
fix: maxItems,minItems are xsd:nonNegativeInteger

ioggstream:
## This PR

- fixes maxItems,minItems datatype xsd:decimal -> nonNegativeInteger


danielpeintner:
The proposed fix seems reasonable to me.

While looking at it I stumbled over `IntegerSchema`

https://github.com/w3c/wot-thing-description/blob/09a0fec43ff431ab2117da893d726751c11cfee5/ontology/jsonschema.html#L204-L205

Should that be `^^xsd:integer` also or is this construct used for NumberSchema also?

sebastiankb:
> While looking at it I stumbled over `IntegerSchema`

yes, seems to be also a mistake.

sebastiankb:
from today's TD call:
- many thanks @ioggstream for this hint
- this PR only changes the html representation, however, we need also to change the ontology files
- we should also fix the IntegerSchema type as mentioned [above](https://github.com/w3c/wot-thing-description/pull/1212#issuecomment-895264718)
- @danielpeintner will have a look, will also ask @vcharpenay for support

danielpeintner:
Note: the bug was purely in a `pre` code section, the ontology was already correct, e.g., 

https://github.com/w3c/wot-thing-description/blob/a2a435dc1e5400d5e60946f4f72cc13946bb02af/ontology/jsonschema.ttl#L151-L158

W.r.t. IntegerSchema I mentioned https://github.com/w3c/wot-thing-description/pull/1212#issuecomment-895264718 I reached out to @vcharpenay

danielpeintner:
> W.r.t. IntegerSchema I mentioned [#1212 (comment)](https://github.com/w3c/wot-thing-description/pull/1212#issuecomment-895264718) I reached out to @vcharpenay

see https://github.com/w3c/wot-thing-description/issues/1219
fixed in this PR also

sebastiankb (approved):

sebastiankb:
from today's TD call we decided to merge this PR

@ioggstream many thanks for the fix

PR #1222, opened 2021-09-08
Add a new action model proposal

relu91:
This PR introduces a third hypermedia control proposal per Actions as we discussed today. Please don't comment the proposal here, below you can find the related issue.


danielpeintner:
I think we should merge this proposal PR

danielpeintner (approved):

PR #1228, opened 2021-09-15
add new action model proposal in other formats

relu91:
This is a follow-up of #1222 it seems that the ODP file generated by google drive cannot be open by PowerPoint. This PR adds PDF and pptx formats.


danielpeintner (approved):
Checked pptx and it now works for me

Issue #1231, opened 2021-09-20
Payload format for meta operations


egekorkan:
I had mentioned this in #1200 but it applies to all meta operations. The payload formats of operations like `readallproperties` have responses from the Thing that contain the values of multiple affordances. There are multiple issues around this:
1. How the payload looks like is badly documented for these operations. There is only one "theoretical" sounding paragraph just **above** https://w3c.github.io/wot-thing-description/#interactionaffordance. There are no examples and there is no mention of operations like `subscribeallevents` or `unsubscribeallevents`.
2. Events and observable properties: Should the name of the affordance appear in the map of the payload? For example, if I am observing all properties but only 1 has changed (which is the likely case), should I still get a map with all the property names (unchanged ones will be empty?) or should their names be omitted from the map? I would have done the second one by intuition but the spec requires the first one. For example, the first one would look like this:
```json
{
  "myProp1": 12,
  "myProp2": null
}
```
3. What happens when some of the affordances have non JSON payloads?
4. (small point) The following text is not clear:
> ObjectSchema instance contains each data schema of the PropertyAffordances

I think that it should not say `each data schema` but `payload`?

benfrancis:
> 1. How the payload looks like is badly documented for these operations. There is only one "theoretical" sounding paragraph just **above** https://w3c.github.io/wot-thing-description/#interactionaffordance.

I agree this part of the spec needs work. These normative statements are quite buried in the text and lack examples. Even though I know they are there I always have to dig for a while to find them.

> There are no examples and there is no mention of operations like `subscribeallevents` or `unsubscribeallevents`.

In the case of the Core Profile the `subscribeallevents` and `unsubscribeallevents` operations using Server-Sent Events don't actually have a payload so it's a bit of a non-issue https://w3c.github.io/wot-profile/#subscribeallevents

For implementations which do have a payload, can you think of a reasonable default payload schema constructed from the schemas of  `subscribeevent` and `unsubscribeevent` operations? Bearing in mind this schema is probably just for the subscription operation, not the actual event payloads (see below).

If something other than the default should be used, it's not clear to me where a schema should be defined. It's possible to define an `ExpectedResponse` in a top level `Form` and we have previously discussed adding a `schema` member to `ExpectedResponse` like the one in `AdditionalExpectedResponse`. There's currently no way to define a _request_ schema in a top level form though.

[...]

Issue #1200, opened 2021-07-22
Is there a need for `invokeanyaction`/`queryallactions` operations?


benfrancis:
Now that we have `subscribeallevents`/`unsubscribeallevents` operations it is possible to have top level forms in a Thing Description for both properties and events. E.g.

```json
  "forms": [
    {
      "op": ["readallproperties", "writeallproperties"],
      "href": "./properties"
    },
    {
      "op": ["subscribeallevents", "unsubscribeallevents"],
      "href": "./events"
    }
  ]
```

In addition to these top level `Properties` and `Events` endpoints, [WebThings](https://webthings.io) also provides a top level [`Actions`](https://webthings.io/api/#actions-resource) endpoint for things. This endpoint has a couple of functions:
- Invoke an action of any type
- Get a list of all pending action requests for all action types

The Actions endpoint is really just provided as a convenience since:
- Actions can be invoked on individual `Action` endpoints
- A list of all pending action requests can be produced by individually querying every action resource and concatenating the results into a list

I'm curious to know whether people would find value in `invokeanyaction`/`queryallactions` operations such that actions could also have a top level form, e.g.

```json
  "forms": [
    {
      "op": ["invokeanyaction", "queryallactions"],
      "href": "./actions"
    }
  ]
```

I think it would be neat to have all three, but how useful is it?

(The alternative is that we drop this feature from WebThings).

egekorkan:
I would say that once querying actions is possible, such an endpoint would be needed. `invokeanyaction` is a bit weird since that is what the forms in the actions are for

benfrancis:
@egekorkan wrote:
> `invokeanyaction` is a bit weird since that is what the forms in the actions are for

The same argument could apply to querying actions, as a `GET` request on the action endpoint could return a list of all pending action requests for that action.

I think there are three possible use cases for "querying" an action:
1. Getting an individual [`ActionStatus`](https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/89.html#ActionStatus) resource regarding an individual action request (e.g. `GET /actions/fade/1935-5939-ngu3`)
2. Getting a list of pending action requests for a given action (e.g. `GET /actions/fade`)
3. Getting a list of pending action requests for all actions (e.g. `GET /actions`)

sebastiankb:
I would be ok for having `queryallactions` as a new op type. However, I also do not see the motivation for `invokeanyaction` either. Is the idea to invoke multiple actions at the same time?

[...]

PR #1232, opened 2021-09-21
Add `queryallactions` operation

benfrancis:
This is the `queryallactions` operation proposed in https://github.com/w3c/wot-thing-description/issues/1200.

You may be pleased to know this is probably the final operation I'm going to propose for Thing Description 1.1. There are other potential candidates (like `invokeanyaction`, `cancelallactions`, `queryactioninstance`, `cancelactioninstance` and `readpastevents`) but I think we can live without them.

I think `queryallactions` is particularly important because without it there may be no way for a Consumer to enumerate ongoing actions, which may mean that only the Consumer which invokes an action can ever query its status and if that Consumer should crash, it may lose the reference to a dynamic resource needed to do so. This operation is probably also needed for the Core Profile, see https://github.com/w3c/wot-profile/issues/99.

There's still some discussion about the payload format for meta operations, but this is being discussed in #1231.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-thing-description/pull/1232.html" title="Last updated on Sep 21, 2021, 4:45 PM UTC (ca6c192)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1232/37a0a29...benfrancis:ca6c192.html" title="Last updated on Sep 21, 2021, 4:45 PM UTC (ca6c192)">Diff</a>


sebastiankb:
from today's TD call, the group decided to merge this PR

PR #1239, opened 2021-10-06
Update impl.csv for 2021.10 Plugfest results

mmccool:
Proposed new structure; "Systems" names need to be updated to match CSV input filenames


sebastiankb:
from today's TD call the group decided to merge this PR. 

the "impl-" prefix by the other companies have to removed as well. Should be done by another PR.  

In addition, in the last column all organization should provide the implementation that they belong to.

Issue #1237, opened 2021-10-04
JSON schema misses ExpectedResponse definitions 


danielpeintner:
The JSON schema [here](https://w3c.github.io/wot-thing-description/#json-schema-for-validation) or [here](https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json) misses definitions for 
* [ExpectedResponse](https://w3c.github.io/wot-thing-description/#expectedresponse) and
* [AdditionalExpectedResponse](https://w3c.github.io/wot-thing-description/#additionalexpectedresponse)

danielpeintner:
Note: I trapped into the same issue as in https://github.com/w3c/wot-thing-description/issues/1235

ExpectedResponse is there but as an _anonymous_ type

https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json#L300-L307

I think it would be good to have all terms in the TD spec also being named types in JSON schema.

Issue #1235, opened 2021-09-30
JSON Schema issue / request


danielpeintner:
## Issue
[OAuth2SecurityScheme](https://w3c.github.io/wot-thing-description/#oauth2securityscheme) states `flow` can be string (e.g., code, client, or device)

The JSON schema seems to require `"code"` only
see https://github.com/w3c/wot-thing-description/blob/0600e4384c129bf1dc1fa4fa1a34074d6b684afa/validation/td-json-schema-validation.json#L1143-L1148

## Request
As you might know we use the JSON schema to auto-generate TypeScript definitions so that they can be used in the Scripting API and in other places, see https://github.com/w3c/wot-scripting-api/tree/main/typescript/thing-description

At the moment the outcome for all security schemes is a big  list of possible anonymous types since they are anonymous types in [JSON schema](https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json#L807-L1156) also.
see https://github.com/w3c/wot-scripting-api/blob/main/typescript/thing-description/thing-description.d.ts#L32-L124

to facilitate re-use it might be beneficial to have proper names like
* NoSecurityScheme
* BasicSecurityScheme
* OAuth2SecurityScheme
* ...

so that anyone can re-use these definitions and/or point to it.

Do you think it is possible to create global types in the JSON schema so that we are getting those named types?

relu91:
I can volunteer looking into this 👍🏻

PR #1250, opened 2021-10-13
Fix JSON Schema issues

relu91:
As described in #1235 and #1237, this PR introduces dedicated type definitions for anonymous types in `securityDefinitions` and `form-element`.

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-thing-description/pull/1250.html" title="Last updated on Nov 10, 2021, 3:27 PM UTC (0c529e5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1250/0600e43...relu91:0c529e5.html" title="Last updated on Nov 10, 2021, 3:27 PM UTC (0c529e5)">Diff</a>

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -1131,7 +1131,7 @@
                             ]
                         },
                         "flow": {
-                           "oneOf": [
+                           "anyOf": [
                             {
                                 "type": "string"
                             },
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -251,6 +251,14 @@
                 }
             ]
         },
+        "expectedResponse": {
+            "type": "object",
+            "properties": {
+                "contentType": {
+                    "type": "string"
+                }
+            }
+        },
         "form_element_property": {
             "type": "object",
             "properties": {
@@ -298,12 +306,7 @@
                     "$ref": "#/definitions/scopes"
                 },
                 "response": {
-                    "type": "object",
-                    "properties": {
-                        "contentType": {
-                            "type": "string"
-                        }
-                    }
+                    "$ref": "#/definitions/expectedResponse"          
                 },
                 "additionalResponses": {
                     "$ref":"#/definitions/additionalResponsesDefinition"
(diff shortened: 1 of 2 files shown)


sebastiankb:
@relu91 what is the status of this PR

relu91:
> Exactly that's why I think we should use the example field as we are doing for subprotocol. But Daniel's proposal might work better with the TS-type generator.

I have to check whether is more convenient to use `examples` or @danielpeintner's solution. After that is ready thank you for pinging me!

relu91:
I checked and it seems that from the TS prospective `examples` are completely ignored. Therefore, I opted for @danielpeintner's solution:
```js
"flow": {
                           "oneOf": [
                            {
                                "type": "string"
                            },
                            {
                                "type": "string",
                                "enum": [
                                    "code",
                                    "client",
                                    "device"
                                ]
                            }]
                        }
                    },
```
The enum helps competition tools (like Intellisense) to suggest possible keywords. I think it might be useful. Do you still think we should just stick with `string` type alone?

egekorkan:
I don't like it a lot to introduce this validation for intellisense but I wouldn't mind too much. A small `$comment` would be good to record why this is done like this.

validation/td-json-schema-validation.json
```diff
@@ -1141,15 +1131,49 @@
                             ]
                         },
                         "flow": {
-                            "type": "string",
-                            "enum": [
-                                "code"
-                            ]
+                           "oneOf": [
```
egekorkan, line 1134: Actually it doesn't work since `code` would validate both of the schemas: https://www.jsonschemavalidator.net/s/VVtuuHK5

Does `anyOf` work with Intellisense

egekorkan (changes requested):

validation/td-json-schema-validation.json
relu91: Thanks, good point! I've just checked with `anyOf` and it seems to produce the same TypeScript code. Works well with Intellisense too. Updating...

egekorkan (approved):

danielpeintner (approved):
LGTM now

sebastiankb:
from today's TD call, going to merge this PR.

Another PR will be provided by @relu91 for TM JSON Schema

PR #1264, opened 2021-11-03
Fix 948 (add OAuth2 flow examples)

### LINKED ISSUE #948: We need an OAuth2 example for TD 1.1
for better understanding how the OAuth2 security scheme is working a good example would be good

relu91:
Close #948

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-thing-description/pull/1264.html" title="Last updated on Dec 2, 2021, 4:50 PM UTC (e2c966d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1264/0600e43...relu91:e2c966d.html" title="Last updated on Dec 2, 2021, 4:50 PM UTC (e2c966d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3361,6 +3361,75 @@ In summary, this requires the following:
     ...
 }
 </pre>
+<p>
+    <a href="#oauth2securityscheme">OAuth2SecurityScheme</a> can specify different <code>flow</code>s, the next two examples report
+    the typical configuration of the <code>client</code> and the <code>device</code> flows. Typically, 
+    the <code>client</code> flow is used for machine to machine communication, consequently the example 
+    shows a fictional Split Air Conditioner that can be controller by his companion thermostat using the 
+    pre-shared client id and client secret. On the other hand, the second example describe a <a>Thing Description</a>
+    of a device that offers the <code>device</code> flow for other Things that do not have a rich user 
+    interface. From the <a>Thing Description</a> prospective <code>device</code> flow is similar to <code>code</code>
+    but complaint clients will act differently following OAuth 2.0 specification. 
+</p>
+<pre class="example">
+{
+   ...
+   "title": "Split Air Conditioner"
+   "securityDefinitions":{
+      "oauth2_sc":{
+         "scheme":"oauth2",
+         "flow":"client",
+         "token":"https://auth0tenat/token",
+      }
+   },
+   "security":[
+      "oauth2_sc"
+   ],
+   "actions":{
+      "activate":{
+         "description":"Start cooling the room",
+         "forms": [ ... ]
+      }
+   }
+}
+</pre>
+
+<pre class="example">
+{
+    ...
+    "securityDefinitions": {
+        "oauth2_sc": {
+            "scheme": "oauth2",
+            ...
+            "flow": "device",
+            "authorization": "https://device.com/authorization/",
+            "token": "https://device.com/token",
+            "scopes": ["limited", "special"]
+        }
+    },
+    "security": "oauth2_sc",
+    ...
+    "properties": {
+        "status": {
+            ...
+            "forms": [{
+                "href": "https://scopes.example.com/status",
+                "scopes": ["limited"]
+            }]
+        }
+    },
+    "actions": {
+        "configure": {
+            ...
+            "forms": [{
+                "href": "https://scopes.example.com/configure",
+                "scopes": ["special"]
+            }]
+        }
+    },
+    ...
+}
+</pre>
 
 <p>A Thing can require an onboarding process that results in the Consumer requiring an API key
     to interact with the Thing. This API key can be included in the request to the Thing in different ways 
--- a/index.html
+++ b/index.html
@@ -3319,7 +3319,7 @@ In summary, this requires the following:
   with a token containing the <code>special</code> scope.
   Scopes are not identical to roles, but are often associated with them;
   for example, perhaps only those in an administrative role are authorized to perform "special" interactions.
-  Tokens can have more than one scope are issued by dedicated web services to users.
+  Tokens can have more than one scope and are issued by dedicated web services to users.
   In this example, an administrator could be issued tokens with both the <code>limited</code> and <code>special</code> scopes,
   while ordinary users could be provided with tokens with the <code>limited</code> scope.
   </p>
@@ -3330,8 +3330,7 @@ In summary, this requires the following:
     "securityDefinitions": {
         "oauth2_sc": {
             "scheme": "oauth2",
-            "flow": "code",
-            "authorization": "https://example.com/authorization",
+            "flow": "client",
             "token": "https://example.com/token",
             "scopes": ["limited", "special"]
         }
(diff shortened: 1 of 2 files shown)


index.template.html
```diff
@@ -2335,6 +2335,75 @@ <h3><code>securityDefinitions</code> and <code>security</code></h3>
...
+}
+</pre>
+
+<pre class="example">
+{
+    ...
+    "securityDefinitions": {
+        "oauth2_sc": {
+            "scheme": "oauth2",
+            ...
```
egekorkan, line 2377: these three dots are not needed right?

egekorkan (changes requested):
mostly minor comments. Thank you for better explaining and exemplifying the OAuth2 flows :)

egekorkan (approved):

sebastiankb:
@mmccool would be great to have a look on this PR. From my side looks good.

index.html
```diff
@@ -3321,8 +3321,8 @@ <h3><code>securityDefinitions</code> and <code>security</code></h3>
   for example, perhaps only those in an administrative role are authorized to perform "special" interactions.
   Tokens can have more than one scope.  
   In this example, an administrator would
-  probably be issued tokens with both the <code>limited</code> and <code>special</code> scopes,
-  while ordinary users would only be issued tokens with the <code>limited</code> scope.
+  probably issue tokens with both the <code>limited</code> and <code>special</code> scopes,
```
mmccool, line 3324: "be issued" is correct in both cases.  These tokens are provided TO the administrator and the user in both cases by the authentication server.  Rather than "be issued" it might be clearer to say "be provided with".  We can also get rid of "would probably" and say "could" so: "could be provided with".  Again, for both roles.

index.template.html
```diff
@@ -2295,8 +2295,8 @@ <h3><code>securityDefinitions</code> and <code>security</code></h3>
   for example, perhaps only those in an administrative role are authorized to perform "special" interactions.
   Tokens can have more than one scope.  
   In this example, an administrator would
-  probably be issued tokens with both the <code>limited</code> and <code>special</code> scopes,
-  while ordinary users would only be issued tokens with the <code>limited</code> scope.
+  probably issue tokens with both the <code>limited</code> and <code>special</code> scopes,
```
mmccool, line 2298: same problem as above.  "be issued" was correct, IMO.  Or, use the alternative wording mentioned above.

index.template.html
```diff
@@ -2335,6 +2334,74 @@ <h3><code>securityDefinitions</code> and <code>security</code></h3>
...
 </pre>
+<p>
+    <a href="#oauth2securityscheme">OAuth2SecurityScheme</a> can specify different <code>flow</code>s, the next two examples report
+    the typical configuration of the <code>client</code> and the <code>device</code> flows. Typically, 
+    the <code>client</code> flow is used for machine to machine communication, consequently the first example below
+    shows a fictional Split Air Conditioner that can be controller by his companion thermostat using the 
+    pre-shared client id and client secret. On the other hand, the second example describe a <a>Thing Description</a>
+    of a device that offers the <code>device</code> flow for other Things that do not have a rich user 
+    interface. From the <a>Thing Description</a> prospective <code>device</code> flow is similar to <code>code</code>
+    but compliant clients will act differently following OAuth 2.0 specification.
```
mmccool, line 2345: This should say HOW these flows are different.  In particular technically both code and device require actual user input (in slightly different ways...) as part of the process, and IMO are probably best suited for onboarding processes.  Only "client" technically allows authentication to be delegated to a service without directly involving the user.  We are in the middle of updating the security best practices document to make this clearer but I expect that we will explicitly recommend that "client" be used for M2M purposes and "device" or "code" used only for onboarding or other cases where there is a display or browser involved.  IMO "client" is the important flow.  It might also be worthwhile to note (again) that implicit and password are deprecated (and not even in the TD spec, you have to use an extension).

sebastiankb:
from today's TD, @mmccool will do another editorial read and do some changes in the text.

mmccool (changes requested):
I think this needs some fixes before it can be merged:
1. issue -> be issued (or "be provided with")
2. I think we should JUST have a client flow example, as it's the only one that really makes sense for M2M, e.g. orchestration.  The other flows are mostly useful for onboarding and other initial setup, in my opinion.  So I would just add a comment "Please see the Security Best Practices for a discussion of usage of the other flows".  At best I think a normal interaction could have BOTH client and another flow, so it can be interacted with directly from a browser as well as from orchestration.  But it's generally easier to just use client in this case, too.

mmccool:
I added a few comments [to the open issue about updating the OAuth2 discussion](https://github.com/w3c/wot-security-best-practices/issues/5) in order to update the Security Best Practices document to align with what's said in the TD spec.

index.html
relu91: I interpreted the sentence the other way around. Administrators and Users will send tokens TO the remote web thing, probably my misunderstanding is due to the fact that the introduction text before the sentence refers to interactions between the client and the remote webthing:
> As another more complex example, OAuth 2.0 makes use of scopes. These are identifiers that may appear in tokens and must match with corresponding identifiers in a resource to allow access to that resource (or Interaction Affordance in the case of W3C WoT). For example, in the following, the status Property can be read by Consumers using bearer tokens containing the scope limited, but the configure Action can only be invoked with a token containing the special scope.

My suggestion is to apply your changes but also change the previous sentence from:
>  Tokens can have more than one scope
to
> Tokens can have more than one scope and are issued by dedicated services to users. 

Or something similar.

relu91:
Thank you @mmccool for your comments. 
> I added a few comments to the open issue about updating the OAuth2 discussion in order to update the Security Best Practices document to align with what's said in the TD spec. While I personally think "client" is probably what almost everything should use in almost all cases we may want to think a little bit more about possible interactions where we want to explicitly require user input every time; maybe for certain safety-critical or drastic, irreversible actions (e.g. self-destruct) then "code" or "device" would be appropriate. Even for onboarding, I'm now thinking because of the need to delegate to a non-human M2M consumer "client" is probably appropriate.

I agree with your general feeling that the `client` flow seems the most appropriate. For example, I had a hard time implementing the `code` flow inside node-wot and in the end, I didn't implement it at all (for different reasons). As you suggested we may move the discussion to the Security task force. Meanwhile, I'll update the PR accordingly to your suggestions.

sebastiankb:
from today's TD call:
- only a client example would be enough
- @mmccool will discuss this topic in the next security call

sebastiankb:
from today's TD call:
- this PR was discussed in the security call
- until next week this PR will be updated based on the resolutions from the security call

relu91:
I should have covered the points described [here](https://github.com/w3c/wot-thing-description/issues/948#issuecomment-981652199)

index.html
mmccool: In the device and code flows, the client is issued tokens by an authentication server, and then provides these later to the web api of the device it is trying to talk to.  The "be issued" here is referring in that case to the interaction between the client and the authentication server, not the client and the device.  So I think the confusion is we are talking about different stages of the process.

I avoided saying in this sentence "provided by the authentication server" however because only "device" and "code" flows actually use authentication servers.  The "client" flow does not (in fact disallows specifying it in the security scheme!).  Since we care mostly about the "client" flow this is important.  In the "client" flow the consumer/client would have been *provisioned* with tokens with the right scopes at some point by some other entity, i.e. during onboarding.

Anyway... the original text was correct, and was worded to cover these cases (not being overly specific about where the token came from, just that it was issued by some other entity).

sebastiankb (approved):

sebastiankb:
from today's TD call:
- decided to merge
- however, the wording "be issued" should be changed since this has a different meaning. In the next Security TF call this will be clarified and most likely another PR will be provided with an small text update @mmccool @relu91

PR #1271, opened 2021-11-10
fix iri to iri-reference and add further instructions

egekorkan:
did not run `npm run render` but can be reflected in the index.html in someone's next PR.


sebastiankb:

Issue #942, opened 2020-08-05
Example 35 is incorrect or has inconsistency


egekorkan:
The Example 35 of the spec has the following TD:
```json
{
    "@context": "https://www.w3.org/2019/wot/td/v1",
    "id": "urn:dev:ops:32473-Thing-1234",
    "title": "WebhookThing",
    "description": "Webhook-based Event with subscription and unsubscribe form.",
    "securityDefinitions": {"nosec_sc": {"scheme": "nosec"}},
    "security": ["nosec_sc"],
    "events": {
        "temperature": {
            "description": "Provides periodic temperature value updates.",
            "subscription": {
                "type": "object",
                "properties": {
                    "callbackURL": {
                        "type": "string",
                        "format": "uri",
                        "description": "Callback URL provided by subscriber for Webhook notifications.",
                        "writeOnly": true
                    },
                    "subscriptionID": {
                        "type": "string",
                        "description": "Unique subscription ID for cancellation provided by WebhookThing.",
                        "readOnly": true
                    }
                }
            },
            "data": {
                "type": "number",
                "description": "Latest temperature value that is sent to the callback URL."
            },
            "cancellation": {
                "type": "object",
                "properties": {
                    "subscriptionID": {
                        "type": "integer",
                        "description": "Required subscription ID to cancel subscription.",
                        "writeOnly": true
                    }
                }
            },
            "uriVariables": {
                "subscriptionID": { "type": "string" }
            },
            "forms": [
                {
                    "op": "subscribeevent",
                    "href": "http://192.168.0.124:8080/events/temp/subscribe",
                    "contentType": "application/json",
                    "htv:methodName": "POST"
                },
                {
                    "op": "unsubscribeevent",
                    "href": "http://192.168.0.124:8080/events/temp/{subscriptionID}",
                    "htv:methodName": "DELETE"
                }
            ]
        }
    }
}
```

For the unsubscription, the corresponding form seems to use `uriVariables` via `"href": "http://192.168.0.124:8080/events/temp/{subscriptionID}",`. There is also the corresponding `uriVariables` object. However, there is also the cancellation object. Does it really mean that I should send the subscriptionID in the payload and in the body of the message? I am guessing that the cancellation object needs to be removed.

P.S. I think there is a related issue to this in TD but I couldn't find it. Here is one from Scripting API: https://github.com/w3c/wot-scripting-api/issues/208

sebastiankb:
good point. 

[...]

PR #1279, opened 2021-11-16
print *all* code parts  on print

danielpeintner:
fixes https://github.com/w3c/wot-thing-description/issues/1059

Note: at the moment only index.html is changed for testing purposes.

With this change
* JS tabs, like  https://w3c.github.io/wot-thing-description/#example-4 and 
* CSS tabs like https://w3c.github.io/wot-thing-description/#example-56
are no longer hidden on print.

A downside is that it does not show a proper title before each code part but instead in the beginning only.
Please try to print index.html and you should see what I mean.

***
<a href="https://pr-preview.s3.amazonaws.com/danielpeintner/wot-thing-description/pull/1279.html" title="Last updated on Nov 18, 2021, 10:42 AM UTC (be95723)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1279/815fc8e...danielpeintner:be95723.html" title="Last updated on Nov 18, 2021, 10:42 AM UTC (be95723)">Diff</a>

PR #1283, opened 2021-11-17
update webhooks example 

sebastiankb:
also see #942

This PR provides two variations of the webhook example and updates the explainer text for better understanding.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1283.html" title="Last updated on Nov 22, 2021, 3:23 PM UTC (657f298)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1283/815fc8e...657f298.html" title="Last updated on Nov 22, 2021, 3:23 PM UTC (657f298)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1283.html" title="Last updated on Jan 12, 2022, 5:02 PM UTC (bf9d149)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1283/815fc8e...bf9d149.html" title="Last updated on Jan 12, 2022, 5:02 PM UTC (bf9d149)">Diff</a>


sebastiankb:
from today's TD call:
- we need to check if the URI variables can work in that way
- we need also to check the definition of the uriVariables and the right wording
- @egekorkan will check this and if this is aligned in the TD spec

egekorkan:
I think that this is definitely better. However, there I have some comments:
- I still find it counter-intuitive to use `readOnly` and `writeOnly` in this way. It is not that we are really writing or reading. It is more about the direction of the message. Our definition of `readOnly` says:
> Boolean value that is a hint to indicate whether a property interaction / value is read only (=true) or not (=false).
- If we go with this example of `uriVariables`, the definition of `uriVariables` need a small update to remove the word `query`. Currently we have:
> Define URI query template variables as collection based on DataSchema declarations. The individual variables DataSchema cannot be an ObjectSchema or an ArraySchema.
I propose to have:
> Define URI template variables according to RFC 6570 as collection based on DataSchema declarations. The individual variables 
DataSchema cannot be an ObjectSchema or an ArraySchema.

- If we add the point above, I propose to add another example below https://w3c.github.io/wot-thing-description/#example-31 that shows the current use of `uriVariables`

sebastiankb:
> I still find it counter-intuitive to use readOnly and writeOnly in this way. It is not that we are really writing or reading. It is more about the direction of the message.

yes, it is thinking around the corner. Needs to improved in the future, maybe should be also deeper addressed by the binding template.

> If we go with this example of uriVariables, the definition of uriVariables need a small update to remove the word query.

I will do so.

> If we add the point above, I propose to add another example below https://w3c.github.io/wot-thing-description/#example-31 that shows the current use of uriVariables

Im ok with that. Can you provide an example?

egekorkan:
I would propose the simple example below that I have somewhat taken from https://developer.accuweather.com/accuweather-forecast-api/apis by replacing location key with city name:
```js
{
    "@context": "http://www.w3.org/ns/td",
    ...
    "properties": {
        "weather": {
            ...
            "uriVariables": {
                "city": { 
                    "type": "string", 
                    "description": "City name to find the weather information for" }
                }
            },
            "forms": [{
              "href": "http://example.org/weather/{city}",
              "htv:methodName": "GET"
            }]
        },
        ...
    },
    ...
}
```

The explanatory text above the example can be:
> Alternatively, `uriVariables` can be used for replacing a part of the URI as specified by [RFC6570]. An example TD is provided below where a valid request to get the forecast of Bogota, Colombia would be an HTTP GET request to `http://example.org/weather/bogota`

As I like examples in general, we can mix the two examples into one more complex one, i.e. 
```js
{
    "@context": "http://www.w3.org/ns/td",
    ...
    "properties": {
        "weather": {
            ...
            "uriVariables": {
                "city": { 
                    "type": "string", 
                    "description": "City name to find the weather information for" }
                },
                "unit": { 
                    "type": "string", 
                    "enum":["Fahrenheit","Celsius"],
                    "description": "Desired unit for the temperature value" }
                }
            },
            "forms": [{
              "href": "http://example.org/weather/{city}/{?unit}",
              "htv:methodName": "GET"
            }]
        },
        ...
    },
    ...
}
```

An example request would be `http://example.org/weather/bogota/?unit=Celsius`

index.html
```diff
@@ -7065,76 +7063,137 @@ <h3>Webhook Event Example</h3>
               which must be submitted through the <code>subscribeevent</code> form.
               The read-only parameter <code>subscriptionID</code> is returned by the subscription.
               The <i>WebhookThing</i> will then periodically POST to this callback URI with a payload defined by <code>data</code>.
```
mlagally, line 7065: I think this language needs a bit of clarifications.
It is obvious that the data is *contained* in the event message, but it could be packaged into an envelop that contains metadata, such as timestamp, originator etc.

Please see https://github.com/w3c/wot-profile/pull/150 for details.

For clarity and better rendering in a printed/PDF version of the document, I suggest to split the example into two examples.

mlagally (changes requested):

index.html
sebastiankb: > For clarity and better rendering in a printed/PDF version of the document, I suggest to split the example into two examples.

This is not needed anymore since #1279 provides a setup that allows to print all examples behind the tabs.

sebastiankb:
from today's TD call:

- @egekorkan example provided here will be provided by an separated PR since it belongs to the uriTemplate section. 
- this PR  will be not merged. @mlagally comments will be taken into account to improve the reading

sebastiankb:
@mlagally I added a note that points out your mentioned comment + that the WoT WG is working on a definition.

sebastiankb:
from today's TD call:
- decided to merge
- however, maybe we should improve the example or add another example that provides another option about implementing a Webhook based approach (also see #1323)

Issue #1274, opened 2021-11-10
Where to place the HTTP Default Binding and which format


egekorkan:
In the Binding part of the TD call of 10.11.2021, there was a small discussion on where to put the HTTP default binding. It was based on the https://github.com/w3c/wot-binding-templates/issues/135 . There are some things to discuss in general:

- [ ] Where should this information be? In the call, @danielpeintner mentioned that the reason to put them in the TD spec in the first place was to show how protocol-specific information is conveyed and that HTTP was the most tested one so far.
- [ ] How should this information be formatted? In the Binding Templates, all protocol bindings have the same way of specifying the default values, i.e. for a given WoT operation, what is its default protocol method/options. In the TD spec, the default values for `htv:methodName` are put in a table and their values depend on the value of the `op`. These two ways are shown in the screenshots below, first one is the TD spec, second is the binding templates HTTP. They contain the same information but in different ways:

1.
![image](https://user-images.githubusercontent.com/20195407/141201984-5929ea56-7820-4fa9-9c54-c0f975d1fbc9.png)

2. 
![image](https://user-images.githubusercontent.com/20195407/141201964-31415e58-c5c9-4000-90f8-5e26469539f6.png)

We should choose one and stick with it. I personally like the second approach, i.e. the one in the binding templates since it is easier to read and reason.

relu91:
> Where should this information be? In the call, @danielpeintner mentioned that the reason to put them in the TD spec in the first place was to show how protocol-specific information is conveyed and that HTTP was the most tested one so far.

Although I acknowledge the fact that leaving it inside the TD spec has the advantage to be normative, I would prefer to move as much information as possible to its Protocol Binding document. The advantage is to have one single place to maintain and point people to read. IMHO Thing Description specification should leave any protocol-specific information to Protocol Binding documents or Profile documents.

benfrancis:
Note that this topic was originally raised in https://github.com/w3c/wot/issues/996, which led me to file https://github.com/w3c/wot-binding-templates/issues/135 and https://github.com/w3c/wot-thing-description/issues/1259 to consider which specification it should be removed from.

In terms of format, I prefer the table from the Protocol Binding Template specification which is organised by operation type. This is how we structure the [protocol binding in the WoT Profile specification](https://w3c.github.io/wot-profile/#protocol-binding) too.

@relu91 wrote:
> IMHO Thing Description specification should leave any protocol-specific information to Protocol Binding documents or Profile documents.

[...]

PR #1301, opened 2021-11-25
update Section 8.3

sebastiankb:
addresses issue https://github.com/w3c/wot-architecture/issues/639

- provide reference to URI scheme RFC3986
- note about IANA

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1301.html" title="Last updated on Dec 8, 2021, 2:56 PM UTC (aa70cad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1301/96220b6...aa70cad.html" title="Last updated on Dec 8, 2021, 2:56 PM UTC (aa70cad)">Diff</a>


benfrancis:
I understand that this change is intended to make it possible to describe existing IoT protocols like MQTT and Modbus in a Thing Description, and whilst I don't personally agree with that I accept that others want to do it.

My concern with this wording though is that because the terms "defacto" and "established" are so open to interpretation it essentially introduces a loophole which can allow _any_ protocol to be a WoT protocol, including non-Internet protocols. In fact it is so open that it arguably makes the assertion meaningless.

If I've understood correctly then someone could theoretically make up their own unofficial URI scheme for a non-IP protocol like Zigbee or Z-Wave and then create a protocol binding template which uses that URI scheme. This blows open the number of protocols that a WoT Consumer might be expected to use even further and removes the compromise we reached to ensure that this couldn't happen.

I'm afraid I don't have an alternative proposal which can work for the MQTT and Modbus use cases but not for the Zigbee and Z-Wave use cases (other than getting those standards bodies to register official URI schemes with IANA), but I'm interested to hear others' views on this.

sebastiankb:
from today's TD call:
- not going to merge. Some improvements will be done based on @benfrancis comment above
- @egekorkan will provide also the location in the binding spec which also address this aspect --> this link will be also covered by the note then

egekorkan:
I sort of forgot what I exactly had to do but I have added the link for the relevant part in the binding spec

egekorkan (approved):

index.html
```diff
@@ -5089,7 +5089,8 @@ <h3>Protocol Bindings</h3>
             <span class="rfc2119-assertion" id="bindings-requirements-scheme">
                 Every form in a WoT Thing Description MUST follow the requirements
                 of the <a>Protocol Binding</a> indicated by the URI scheme [[!RFC3986]] of its
-                <code>href</code> member.
+                <code>href</code> member as indicated by the Binding Templates specification at <a href="https://w3c.github.io/wot-binding-templates/#creating-a-new-protocol-binding">Creating a New Protocol Binding</a>.
```
mlagally, line 5092: This language is difficult to understand, since it refers the entire binding specification. The implication on the URI scene should be clarified.

egekorkan: It shouldn't be the case, for me the relative part is shown when I click on https://w3c.github.io/wot-binding-templates/#creating-a-new-protocol-binding

sebastiankb:
from today's TD call:
- decided to merge
- we need to double check with @mlagally what in particular should be improved based on the comment [above](https://github.com/w3c/wot-thing-description/pull/1301#discussion_r764957131) --> another PR can take this into account then
- another discussion was about if the section 8.3.1. should be moved to the binding document. there is a discussion in issue https://github.com/w3c/wot-thing-description/issues/1274
- in general we should decide what should be normative in TD regarding protocol binding topics.

mlagally (changes requested):
This language contains a reference to the binding template spec in a RFC assertion.
The sentence needs to be split and move the binding templates out of the normative part.

PR #1304, opened 2021-11-25
Remove TD Canonicalization

mmccool:
This PR
- Removes the definition of "Canonical TDs"
    - as discussed; the problem is that the current definitions do not deal with named objects, and without doing so it is not possible to define a canonical form (since names could be arbitrary without changing the meaning), but we can't inline SecurityScheme objects into security without breaking compatibility.  When this feature is added, e.g. in TD 2.0, we can bring canonicalization back).  We *could* define rules to generate names, and that was done in an earlier version, but it gets ugly and complicated and allows for a security hole when partial signing is used (if only the use is signed but not the definitions or vice-versa).
- Removes the section defining canonicalization
- Removes mention of this from the "Recent updates"

I also updated the implementation report (and did a little cleanup) since removing canonicalization also removes a number of assertions.  
- Updates the implementation report   
- This PR also removes assertions.js (it will be available and run separately).  
- Also, the report generator was updated to put a table of all TD assertions in testing/assertions.csv

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1304.html" title="Last updated on Nov 25, 2021, 5:09 PM UTC (568a973)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1304/ea62552...mmccool:568a973.html" title="Last updated on Nov 25, 2021, 5:09 PM UTC (568a973)">Diff</a>


mmccool:
Note: renaming would work without a security hold if we supported only signing of entire TDs, which is at any rate only what is supported in current standards like JWS.  But JWS has other problems.  On balance I think it's better to wait until JSON-LD signing is completed and use that.  Ideally that would also include handling for default values etc. but we will have to work with the JSON-LD WG to make sure it meets our requirements.  Also, as noted above, we probably want to still allow object inlining everywhere a named definition can be used in TD2.0, as this is more secure when partial signing is allowed (and also less verbose in simple cases).

sebastiankb (approved):

mmccool:
sorry for mixing up the canonicalization issues and the impl report update.  yes, yes, those have been two different PRs.

sebastiankb:
from today's TD call, group decided to merge.

For record: This canonicalization section can be seen in branch https://github.com/w3c/wot-thing-description/tree/canonicaliziation-2021-12-01

Issue #1311, opened 2021-12-06
[validation] Reorganize schema to have one inherited form schema


egekorkan:
Coming from https://github.com/eclipse/thingweb.node-wot/pull/550#issuecomment-984796415

egekorkan:
The PR has fixed this already, closing

PR #1319, opened 2021-12-08
Make forms inherited from a base form

egekorkan:
Addresses #1311

Some basic tests are available at https://www.jsonschemavalidator.net/s/lNpA9CYs


danielpeintner (approved):
LGTM!

Note: I recommend ignoring whitespaces since the PR touches lots of them...

danielpeintner (commented):
forms for property should point to `form_element_property`

while the definition of `form_element_property `should have the "allOf" construct.

```
{
  "allOf": [{ "$ref": "#/definitions/form_element_base" }],
  "properties": {
     "op": {
        ...
     }
  }
}
```

Moreover, `form_element_base` might also need the `op` values.. in this case defined as string ... with no enumerations. Thinking of it in an hierarchical fashion root/property/actions/events should be more specific.

I "believe" doing so it might work..

relu91 (changes requested):
I agree with @danielpeintner having allOf inside form_element_property it is more self-contained. 

Regarding ts typings it may be beneficial to have a generic Form type: 
```js
{
  // schema ...
   form : {
      oneOf: [
           {  $ref : "#/definitions/form_element_property" },
           {  $ref : "#/definitions/form_element_action" },
           {  $ref : "#/definitions/form_element_event" },
           {  $ref : "#/definitions/form_element_root" },
       ]
   }
}
 ```
 Although this would not have any validation purposes, it should solve also: https://github.com/eclipse/thingweb.node-wot/pull/550

egekorkan:
Three points about the comments above:

- `Although this would not have any validation purposes` : This is sadly correct and I would not support this direction since the schema would lose its main purpose. One possible way forward is to do some preprocessing before generating TS types 🙈 
-` form_element_property should have the "allOf" construct.` : I am not sure what is meant here but I have done this: https://www.jsonschemavalidator.net/s/Fy7VkyFE based on https://github.com/json-schema-org/json-schema-spec/issues/348
- `Moreover, form_element_base might also need the op values.. in this case defined as string ... with no enumerations. Thinking of it in an hierarchical fashion root/property/actions/events should be more specific.` I have fixed it at the link above  This is what you mean right?

Some other comments:

- What is needed is that a schema that validates and still generates TS types accordingly. Since the op value depends on the affordance it is placed in and that JSON Schema cannot "backtrack" to add requirements, I do not know how to really fix this.

egekorkan:
Converting to a draft since I have found some major issues with the tests. Basically none of the tests were working, now they do and reveal an issue with the regex.

`const placeholderPattern = "^.*{{2}[ -~]+}{2}.*$";` seems to be not a valid regex. I get `Invalid regular expression ... Lone quantifier brackets`

I need to understand it properly before making it available

egekorkan:
I have learned that I have to encapsulate `{` into `[}]` to avoid regex errors. Ready to merge

sebastiankb:

Issue #1316, opened 2021-12-08
Removing Overview.html file


egekorkan:
There is the following file in the repo that I don't know the use of. If I use the GitHub search function for this file, I find the following:

https://github.com/w3c/wot-thing-description/issues/14
https://github.com/w3c/wot-thing-description/blob/24c2bf864d3cfa0814808bc9eb31cd5d15b84d41/publication/wd3/manifest.txt

It would be good to remove this file since it is cluttering the repo and more annoyingly clutters the search results in IDEs

PR #1328, opened 2021-12-15
Delete Overview.html

egekorkan:
Fixes https://github.com/w3c/wot-thing-description/issues/1316

This file was a static version (after respec) used for publication purposes. It is now stale and can be removed.


sebastiankb:

Issue #1323, opened 2021-12-09
Missing event/notification affordance


mlagally:
In the context of the recent profile conversations it became apparent to me, that the TD is currently missing an important affordance. 

When we have two servients that communicate via events, one is the event producer and the other the event consumer. The TD allows to describe the producer interface (subscribe/unsubscribe) but it is missing the notify affordance, i.e. an affordance that receives events.

This is the blocking point to describe Event *consumers* with TDs. As a workaround, the profile can define action endpoints actions/notify, however this is quite unnatural and an extension of the current WD is desirable.

egekorkan:
Do you mean something like describing an endpoint of a server where the event notifications should be sent by the Thing? I sort of agree that this is missing but I do not know how we can describe it in the TD. TD describes the Thing for **any** Consumer in the end. 

If this is coming from the WebHook perspective, I think that this kind of information should not be really visible in the WoT. Analogy with MQTT:
The Thing is programmed with the certain topics it will publish or subscribe to. In the case of Webhook, it would be like having static URLs programmed into the Thing to send messages to. There can be mechanisms implemented in MQTT that allows the Thing to publish to certain topics however we do not have a standardized way.

There can be a bit of overlap with the opening connection mechanism that will hopefully come to TD 2.0? See https://github.com/w3c/wot-thing-description/issues/878, https://github.com/w3c/wot-thing-description/issues/1242, https://github.com/w3c/wot-thing-description/issues/1070

egekorkan:
I would also say that this is not an affordance but an operation

benfrancis:
This is something I've noticed too when trying to describe anything other than a Thing as an HTTP server.

All of the current values for `op` describe messages sent from a Consumer to a Producer.

Messages sent from a Producer to a Consumer can't easily be described by a `Form` because `Form`s represent endpoints of a Producer (the Thing the Thing Description describes), not the Consumers that may consume it. As @egekorkan alluded to, even if a Consumer had a URI that could be used as the `href` member, that URI would likely be different for every Consumer so can't be described by a static `Form` in the Thing Description.

[...]

PR #1329, opened 2021-12-15
Event consumer: Alternative 1 - event operation 

mlagally:
Event consumer operation to enable describing a servient with an event listener notification interface in a TD.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1329.html" title="Last updated on Dec 15, 2021, 9:46 PM UTC (6a563db)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1329/21ce84a...6a563db.html" title="Last updated on Dec 15, 2021, 9:46 PM UTC (6a563db)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1434,13 +1434,15 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
           containing a combination of these terms.
           </span></p></section>
 <section><h3><code>EventAffordance</code></h3><p>An Interaction Affordance that describes an event
-          source, which asynchronously pushes event data to
+          source and an event consumer, which asynchronously pushes event data to
           <a href="#dfn-consumer" class="internalDFN" data-link-type="dfn">Consumers</a> (e.g., overheating
           alerts).</p><table class="def"><thead><tr><th><a>Vocabulary term</a></th><th>Description</th><th>Assignment</th><th>Type</th></tr></thead><tbody><tr class="rfc2119-table-assertion" id="td-vocab-subscription--EventAffordance"><td><code>subscription</code></td><td>Defines data that needs to be passed upon
                 subscription, e.g., filters or message format for
                 setting up Webhooks.</td><td>optional</td><td><a href="#dataschema"><code>DataSchema</code></a></td></tr>
-<tr class="rfc2119-table-assertion" id="td-vocab-data--EventAffordance"><td><code>data</code></td><td>Defines the data schema of the Event instance
+<tr class="rfc2119-table-assertion" id="td-vocab-data-response--EventAffordance"><td><code>data</code></td><td>Defines the data schema of the Event 
                 messages pushed by the Thing.</td><td>optional</td><td><a href="#dataschema"><code>DataSchema</code></a></td></tr>
+                <tr class="rfc2119-table-assertion" id="td-vocab-data-response--EventAffordance"><td><code>dataresponse</code></td><td>Defines the data schema of the Event response
+                    messages sent be the consumer in a response to a data message.</td><td>optional</td><td><a href="#dataschema"><code>DataSchema</code></a></td></tr>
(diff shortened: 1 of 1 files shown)


egekorkan:
My review for both proposals: postpone to TD2.0 and avoid any kind of discussion that would imply more discussion time hence less time for making sure that the current specification is well specified. 

Opinion in general:
Even though this is adding _simply_ another operation, the mechanism and the behavior behind is not well explained nor the use case is specified beforehand nor anything was tried out (I do not mean tested/validated, simply just tried out) in a plugfest. Even if this is possible and implementable, we did not consider any other approaches and we should simply not approach a standardization activity like TD this way which tries to make sure that we can describe Thing interfaces in a generic way and not specify the first way that comes to mind.

mlagally:
@egekorkan 

> Opinion in general:
> Even though this is adding _simply_ another operation, the mechanism and the behavior behind is not well explained nor the use case is specified beforehand nor anything was tried out (I do not mean tested/validated, simply just tried out) in a plugfest. Even if this is possible and implementable, we did not consider any other approaches and we should simply not approach a standardization activity like TD this way which tries to make sure that we can describe Thing interfaces in a generic way and not specify the first way that comes to mind.

The use case has been discussed several times in recent profile conversations.
It is a scalable event mechanism, and a mechanism to describe consumers (which are also things) with a TD in a natural way. 

I don't understand your concerns wrt. implementability. IMHO introducing a single additional operation that takes an input and output payload should not be too hard. What am I missing?

egekorkan:
It is not a simple operation like adding `queryaction`, i.e. initiated from the Consumer to the Thing with a certain protocol request. It is telling the Thing to do something later on, thus a paradigm shift when done with a non-standard protocol. In order to implement it, I need to include an HTTP server in my Consumer that I was not doing until now. This implies opening ports etc. Also, you are not considering **at all** what the security requirements of the Server (now the Consumer) is and what the request-response content type, methods are.

Please implement it first before claiming it is a trivial implementation. It is indeed trivial from a technical point of view but it is not from an architectural one. It is also trivial to write a single script for a Thing and Consumer that exhibit this behavior but not so trivial when done as part of an existing implementation with a certain software architecture. You should at least provide some examples on how this looks like in a TD and a textual description of a couple of sentences on what it means.

@mlagally you have also edited only the index.html that will be removed in any next PR.

** For anyone who is reviewing: This looks like a small PR but it is a false impression. **

**In any case, please postpone this topic until TD2.0 and not steal time from the publication process. I am objecting this PR no matter what changes in its content.**

mlagally:
> In any case, please postpone this topic until TD2.0 and not steal time from the publication process. I am objecting this PR no matter what changes in its content.

@egekorkan Is this a personal comment or a statement on behalf of Siemens?

egekorkan:
> @egekorkan Is this a personal comment or a statement on behalf of Siemens?

It is a personal comment.

sebastiankb:
@mlagally it is quite hard to understand how this proposal should work. As already mentioned it would be helpful to have a practical example. I think, a sequence diagram as provided [here](https://w3c.github.io/wot-binding-templates/bindings/protocols/http/index.html#subscribe-notify-and-unsubscribe-event-http-binding-with-server-sent-event-subprotocol) would be great.

mlagally:
@sebastiankb THanks for your constructive proposal.
I will provide a sequence diagram will illustrate the propsal in issue #1323

sebastiankb:
this PR is replaced by #1377

egekorkan:
TD Call 13.12:
- The issue https://github.com/w3c/wot-thing-description/issues/1323 is already tracking this discussion. 
- Changing the index.html alone would not suffice anyways so the PR is not mergeable as is.
- A discussion should happen first before agreeing on one of the PRs (i.e. https://github.com/w3c/wot-thing-description/pull/1330)
- The same will be commented at the other alternative
- Closing

PR #1331, opened 2021-12-16
fix: move CBOR content types their own table row in section 5.3.2

JKRhb:
In the table in section 5.3.2 the CBOR content types are listed in a row labelled "JSON" which is probably incorrect. This PR provides a fix for this issue.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-thing-description/pull/1331.html" title="Last updated on Feb 16, 2022, 4:53 PM UTC (44ee6e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1331/21ce84a...JKRhb:44ee6e3.html" title="Last updated on Feb 16, 2022, 4:53 PM UTC (44ee6e3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1537,6 +1537,7 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
                      <code>application/xml</code> <br />
                      <code>application/senml+xml</code> <br />
                      <code>application/exi</code> <br />
+                     <code>application/senml-exi</code> <br />
                   </td>
                </tr>
             </tbody>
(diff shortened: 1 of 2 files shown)


danielpeintner:
Mhh, not sure If I would agree with the proposed change.

CBOR to me falls into the JSON category... CBOR can be converted to JSON and JSON can be converted to CBOR.
I know that CBOR is more powerful in some regards but still I think flagging it as `efficient` JSON format is fine.

Otherwise we would need yet another category or EXI .. which we currently flag as XML format.

sebastiankb:
from today's TD call decided to merge

Issue #1053, opened 2021-02-17
Consider adding DataSchema to response and additionalResponses


mmccool:
When there are additionalResponses, and *maybe* also when there is just a "response", the output DataSchema may not be the same as the input DataSchema.   For example, even when getting a property, an error response may not contain the property but an error code and message instead.

However, right now ExpectedResponse only has a ContentType.  This is OK if the response is unstructured (eg JPEG image), but it the response is JSON or XML a DataSchema is needed.

I propose adding an optional "data" field with type "DataSchema" to both "response" and "additionalResponses"

mmccool:
One problem: actions already have an "input" and "output" data schema, so adding another data schema inside "response" would be redundant and confusing.  We can just avoid using it for properties since if the response is not the same as the input then we can just call it an action.  So we could do it only in additionalResponses, but then I would have to create a subclass of ExpectedResponse in the ontology, etc. to model this.

sebastiankb:
from today's TD call: @mjkoster can support here since they had the same issue in OneDM

mjkoster:
It seems like the error case is the most important to allow additional response types, where "the usual programming is interrupted" and a different payload is sent, for example in the case of Modbus errors. 

The important distinction is that there is special header information like a response code that tells the consumer to look for the special data payload. In the case of diverse payloads that occur "normally", the anyOf construct might be more suitable. 

In other cases, we want to model optionality in the TM but will choose one for the instance and TD.

farshidtz:
> One problem: actions already have an "input" and "output" data schema, so adding another data schema inside "response" would be redundant and confusing. 

Also, events already have a `data` field for output data schema, equivalent to actions' `output`.

Seems like all interactions need a schema for the default response and this could go into Interaction Affordance, deprecating event `data` and action `output`.

mjkoster:
I'd like to consider the logical structures here, and try to drive to a concrete proposal.

There are input and output schemas for actions, and data schemas for properties. Events have a notification schema. These schemas are all part of the application data model for interactions. Only the formatting is expected to change when the protocol binding changes (array vs. object for multiple items, key names, contentType...).

Subscription schemas and cancellation schemas for events are arguably part of the protocol binding, because they would be different for different protocols. I recommend that we try to work subscription and cancellation into these schema extensions for forms, also bc that's where the ops for them are defined.

[...]

Issue #302, opened 2018-11-23
How do you cancel or query the state of an action request?


benfrancis:
In Mozilla's [Web Thing API](https://iot.mozilla.org/wot), an action can be requested using an HTTP `POST` request on an [Action resource](https://iot.mozilla.org/wot/#action-resource) to create an [ActionRequest resource](https://iot.mozilla.org/wot/#actionrequest-resource). The Action resource is essentially an action queue, consisting of multiple ActionRequest resources.

The response to the `POST` provides a unique URL for the ActionRequest resource, which can then have its status queried with a `GET` or be cancelled with a `DELETE`. A list of all current requests can be retrieved by a `GET` on the Action resource.

How would this API be described in a Thing Description following the current draft specification? Or is there another intended way to achieve these use cases?

benfrancis:
To provide some additional context for people who don't want to read the Web Thing API specification...

To request an action, the Web Thing API uses a POST request, e.g. 

```
POST https://mythingserver.com/things/lamp/actions/fade
Accept: application/json

{
  "fade": {
    "input": {
      "level": 50,
      "duration": 2000
    }
  }
}
```

Response:
```
201 Created

{
  "fade": {
    "input": {
      "level": 50,
      "duration": 2000
    },
    "href": "/things/lamp/actions/fade/123e4567-e89b-12d3-a456-426655"
    "status": "pending"
  }
}
```

You can get a list of current action requests.

Request:
```
GET /things/lamp/actions/fade
Accept: application/json
```

Response:
```
200 OK
[
  {
    "fade": {
      "input": {
        "level": 50,
        "duration": 2000
      },
      "href": "/things/lamp/actions/fade/123e4567-e89b-12d3-a456-426655",
      "timeRequested": "2017-01-25T15:01:35+00:00",
      "status": "pending"
    }
  },
  {
    "fade": {
      "input": {
        "level": 100,
        "duration": 2000
      },
      "href": "/things/lamp/actions/fade/123e4567-e89b-12d3-a456-426655",
      "timeRequested": "2017-01-24T11:02:45+00:00",
      "timeCompleted": "2017-01-24T11:02:46+00:00",
      "status": "completed"
    }
  }
]
```

You can get the status of an action request.

Request:
```
GET /things/lamp/actions/fade/123e4567-e89b-12d3-a456-426655
Accept: application/json
```

Response:
```
200 OK
{
  "fade": {
    "input": {
      "level": 50,
      "duration": 2000
    },
[...]

PR #1341, opened 2021-12-21
schema term in expected response

sebastiankb:
also see #1053

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1341.html" title="Last updated on Dec 21, 2021, 4:38 PM UTC (00eef3d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1341/21ce84a...00eef3d.html" title="Last updated on Dec 21, 2021, 4:38 PM UTC (00eef3d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1394,8 +1394,8 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
 <li><a href="#actionaffordance"><code>ActionAffordance</code></a></li>
 <li><a href="#eventaffordance"><code>EventAffordance</code></a></li></ul></section>
 <section><h3><code>PropertyAffordance</code></h3><p>An Interaction Affordance that exposes state of the
-          Thing. This state can then be retrieved (read) and
-          optionally updated (write). Things can also choose to
+          Thing. This state can then be retrieved (read) and/or updated (write).
+           Things can also choose to
           make Properties observable by pushing the new state after
           a change.</p><table class="def"><thead><tr><th><a>Vocabulary term</a></th><th>Description</th><th>Assignment</th><th>Type</th></tr></thead><tbody><tr class="rfc2119-table-assertion" id="td-vocab-observable--PropertyAffordance"><td><code>observable</code></td><td>A hint that indicates whether Servients hosting
                 the Thing and Intermediaries should provide a
--- a/index.html
+++ b/index.html
@@ -2040,7 +2040,7 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
                 <td>
                     0..*
                 </td>
-                <td>Point to the web app manifest of a web application which provides a user interface with which a user can interact with the Thing (also see [[APPMANIFEST]]).</td>
+                <td>Point to a software implementation of the TD.</td>
                 <td>
                     <a href="https://www.iana.org/assignments/link-relations">IANA Link Relation</a>
                 </td>
--- a/index.html
+++ b/index.html
@@ -1201,7 +1201,7 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
           "td-context-ns-thing-mandatory">The <code>@context</code>
           name-value pair <em class="rfc2119" title=
           "MUST">MUST</em> contain the anyURI
-          <code>https://www.w3.org/2019/wot/td/v1</code> either
+          <code>https://www.w3.org/2022/wot/td/v1.1</code> either
           directly when of type <code>anyURI</code> or as first
           element when of type <a href="#dfn-array" class=
           "internalDFN" data-link-type="dfn">Array</a>.</span>
@@ -1209,7 +1209,7 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
           "td-context-ns-thing-optional">When <code>@context</code>
           is an <a href="#dfn-array" class="internalDFN"
           data-link-type="dfn">Array</a>, the anyURI
-          <code>https://www.w3.org/2019/wot/td/v1</code> <em class=
+          <code>https://www.w3.org/2022/wot/td/v1.1</code> <em class=
           "rfc2119" title="MAY">MAY</em> be followed by elements of
           type <code>anyURI</code> or type <a href="#dfn-map"
           class="internalDFN" data-link-type="dfn">Map</a> in any
--- a/index.html
+++ b/index.html
@@ -1197,7 +1197,7 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
 <tr class="rfc2119-table-assertion" id="td-vocab-schemaDefinitions--Thing"><td><code>schemaDefinitions</code></td><td>Set of global-based named data schemas.
                 To be used in a local <code>schema</code>
                 name-value pair such as inside an 
-                <code>AdditionalExpectedResponse</code> object.</td><td>optional</td><td><a>Map</a> of <a href="#dataschema"><code>DataSchema</code></a></td></tr></tbody></table><p><span class="rfc2119-assertion" id=
+                <code>ExpectedResponse</code> and <code>AdditionalExpectedResponse</code> object.</td><td>optional</td><td><a>Map</a> of <a href="#dataschema"><code>DataSchema</code></a></td></tr></tbody></table><p><span class="rfc2119-assertion" id=
           "td-context-ns-thing-mandatory">The <code>@context</code>
           name-value pair <em class="rfc2119" title=
           "MUST">MUST</em> contain the anyURI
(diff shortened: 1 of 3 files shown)


sebastiankb:
@benfrancis I double check the issue #1053 . It seems there is a mixup with `ExpectedResponse` and `AdditionalExpectedResponse`.  The examples in #1053 relying on `AdditionalExpectedResponse` where we already have `schema` term in there. It seems that we do not need anymore this PR

benfrancis:
Please see https://github.com/w3c/wot-thing-description/issues/302 and https://github.com/w3c/wot-profile/pull/91 for more detail on use cases

egekorkan:
I think we should solve this in a better way. If we have schema in forms level, for the response of an action, we can have the schema in the output at the action level or in the forms level. Putting more and more "stuff" into forms make the whole WoT more protocol specific. I would like to have maybe query/input and query/output in the action level rather than in the form level. That was part of my hypermedia proposal at https://github.com/w3c/wot-thing-description/tree/main/proposals/hypermedia-control-2

benfrancis:
@egekorkan wrote:
> I think we should solve this in a better way. If we have schema in forms level, for the response of an action, we can have the schema in the output at the action level or in the forms level. Putting more and more "stuff" into forms make the whole WoT more protocol specific. I would like to have maybe query/input and query/output in the action level rather than in the form level. That was part of my hypermedia proposal at https://github.com/w3c/wot-thing-description/tree/main/proposals/hypermedia-control-2

That seems reasonable. Data schemas do already exist at the forms level for `AdditionalExpectedResponse`, but I agree it might be cleaner for a protocol-agnostic schema to be provided at the interaction affordance level.

Does a query action operation really need a separate input and output though? Could there just be one data schema for querying an action status like reading a property?

```
{
  "actions": {
    "fade": {
      "input": {...},
      "output": {...},
      "query": {...}
    }
  }
}
```

It does also beg the question of how a Consumer should interpret the meaning of the `queryaction` response payload, since it could contain anything. In the Core Profile those semantics are defined out of band in the specification.

Note that we don't have the need for a "cancel" data schema in the Core Profile since there's no payload (just a `DELETE` request on the action URL), but I can imagine that might be needed for some protocol bindings.

egekorkan:
@benfrancis wrote: 
> Does a query action operation really need a separate input and output though? Could there just be one data schema for querying an action status like reading a property?

I think that is also possible. We can start with the simple approach and improve if there are limitations.

> It does also beg the question of how a Consumer should interpret the meaning of the queryaction response payload, since it could contain anything.

I would imagine some semantic annotations would be needed if this needs to be flexible.

Note: We can be flexible and prescriptive on anything about hypermedia, there is not a really widely adopted way to do it.

index.html
```diff
@@ -1189,22 +1189,22 @@ <h2>Core Vocabulary Definitions</h2>
                 those defined in <code>securityDefinitions</code>.
                 These must all be satisfied for access to
                 resources.</td><td>mandatory</td><td><a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a> or <a>Array</a> of <a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a></td></tr>
-<tr class="rfc2119-table-assertion" id="td-vocab---Thing"><td><code></code></td><td>Set of named security configurations
+<tr class="rfc2119-table-assertion" id="td-vocab-securityDefinitions--Thing"><td><code>securityDefinitions</code></td><td>Set of named security configurations
                 (definitions only). Not actually applied unless
                 names are used in a <code>security</code>
-                name-value pair.</td><td>mandatory</td><td><a href="#securityscheme"><code>SecurityScheme</code></a> or <a>Array</a> of <a href="#securityscheme"><code>SecurityScheme</code></a></td></tr>
+                name-value pair.</td><td>mandatory</td><td><a>Map</a> of <a href="#securityscheme"><code>SecurityScheme</code></a></td></tr>
```
mlagally, line 1195: That's a significant change for implementations.
How does this change correspond to the expected response and the issue that is addressed here?
It looks like an unrelated bug fix.

index.html
```diff
@@ -1189,22 +1189,22 @@ <h2>Core Vocabulary Definitions</h2>
...
 <tr class="rfc2119-table-assertion" id="td-vocab-profile--Thing"><td><code>profile</code></td><td>Indicates the WoT Profile mechanisms followed by this Thing Description and the corresponding Thing implementation.</td><td>optional</td><td><a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#anyURI"><code>anyURI</code></a> or <a>Array</a> of <a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#anyURI"><code>anyURI</code></a></td></tr>
-<tr class="rfc2119-table-assertion" id="td-vocab-schemaDefinitions--Thing"><td><code>schemaDefinitions</code></td><td>Set of named data schemas.
-                To be used in a <code>schema</code>
-                name-value pair inside an 
-                <code>AdditionalExpectedResponse</code> object.</td><td>optional</td><td><a>Map</a> of <a href="#dataschema"><code>DataSchema</code></a></td></tr></tbody></table><p><span class="rfc2119-assertion" id=
+<tr class="rfc2119-table-assertion" id="td-vocab-schemaDefinitions--Thing"><td><code>schemaDefinitions</code></td><td>Set of global-based named data schemas.
+                To be used in a local <code>schema</code>
```
mlagally, line 1198: What is a "global-based" and "local" security schema?
The term global-based is not used anywhere else in the document - It would be good to explain.

index.html
```diff
@@ -2395,7 +2396,13 @@ <h2>Hypermedia Controls Vocabulary Definitions</h2>
...
                 (e.g., <code>text/plain</code>) and potential
                 parameters (e.g., <code>charset=utf-8</code>) for
-                the media type [<cite><a class="bibref" data-link-type="biblio" href="#bib-rfc2046" title="Multipurpose Internet Mail Extensions (MIME) Part Two: Media Types">RFC2046</a></cite>].</td><td>mandatory</td><td><a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a></td></tr></tbody></table></section>
+                the media type [<cite><a class="bibref" data-link-type="biblio" href="#bib-rfc2046" title="Multipurpose Internet Mail Extensions (MIME) Part Two: Media Types">RFC2046</a></cite>].</td><td>mandatory</td><td><a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a></td></tr>
+<tr class="rfc2119-table-assertion" id="td-vocab-schema--ExpectedResponse"><td><code>schema</code></td><td>Used to define the output data schema
```
mlagally, line 2400: This is the same text as for the "AdditionalExpectedResponse".
What is the difference between ExpectedResponse and AdditionalExpectedResponse?
Why does only the AdditionalExpectedResponse have a status code?
What does "default" mean for these two types?

validation/td-validation.ttl
```diff
@@ -170,26 +170,26 @@
...
+                <code>ExpectedResponse</code> and <code>AdditionalExpectedResponse</code> object."""^^rdf:HTML ;
         sh:node :DataSchemaShape ;
         sh:order 18 ;
     ] ;
     skos:scopeNote """<p><span class="rfc2119-assertion" id=
           "td-context-ns-thing-mandatory">The <code>@context</code>
           name-value pair <em class="rfc2119" title=
           "MUST">MUST</em> contain the anyURI
-          <code>https://www.w3.org/2019/wot/td/v1</code> either
+          <code>https://www.w3.org/2022/wot/td/v1.1</code> either
```
mlagally, line 184: That seems to be an unrelated change, which should be in a separate PR.

index.html
mlagally: How is backwards compatibility with TD1.0 guaranteed?

sebastiankb: This is a bug fix which was already fixed in a previous PR. I dont' know why this shows-up again but this change is correct and fulfill TD 1.0:

image

index.html
sebastiankb: > This is the same text as for the "AdditionalExpectedResponse".

Yes, it the same intention.

> What is the difference between ExpectedResponse and AdditionalExpectedResponse?

ExpectedResponse can be defined only once. AdditionalExpectedResponse can be defined as array and may be need to define, e.g., different error responses messages (I think this was also requested by you).

>  Why does only the AdditionalExpectedResponse have a status code?

Adding this to ExpectedResponse would violate backwards compatibility.

> What does "default" mean for these two types?

See https://w3c.github.io/wot-thing-description/#sec-default-values

validation/td-validation.ttl
sebastiankb: strange. It seems a race condition with other PRs. The namespace https://www.w3.org/2022/wot/td/v1.1 is correct

sebastiankb:
from today's TD call:
- it seems this PR is not be stable enough to be merged
- it needs further discussion about the motivation 
- decided no to have this for the 2nd WD

egekorkan:
We will close this but it is referenced at https://github.com/w3c/wot/blob/main/planning/ThingDescription/td-next-work-items/usability-and-design.md#reusable-element-design . A new discussion will be started around this whole topic since it is more than just reusable schemas.

Issue #1356, opened 2022-01-18
Fix Respec errors


mlagally:

danielpeintner:
We see do have the following 5 warnings

![grafik](https://user-images.githubusercontent.com/11281461/150342370-58ff5d04-0e7b-4fae-80a2-a4ee3874e1cb.png)

PR #1359, opened 2022-01-20
refactor: ignore definitions that are not used

danielpeintner:
fixes https://github.com/w3c/wot-thing-description/issues/1356

### DIFF
--- a/index.html
+++ b/index.html
@@ -685,9 +685,9 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
   <p>The fundamental WoT terminology such as
   <dfn>Thing</dfn>,
   <dfn>Consumer</dfn>,
-  <dfn>Thing Description</dfn> (<dfn>TD</dfn>),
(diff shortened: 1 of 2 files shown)


danielpeintner:
Note: we do have 2 remaining issues 

grafik


Properly fixing it would require to split  [9. Security and Privacy Considerations](https://w3c.github.io/wot-thing-description/#sec-security-consideration) into **two** sections 
* Privacy Considerations
* Security Considerations

I don't think we should do that in the current state. We could also simply ignore this check.
Any thoughts?

sebastiankb:
> I don't think we should do that in the current state. We could also simply ignore this check.
Any thoughts?

would be ok for me. @mmccool what do you think?

sebastiankb (approved):

sebastiankb:
from today's TD call:
- decided to merge
- discussion about the security and privacy section is moved to this issue [here](https://github.com/w3c/wot-thing-description/issues/1363)

Issue #1348, opened 2022-01-10
Review Security and Privacy Considerations


mmccool:
The security and privacy considerations section needs to be reviewed and updated.  In particular, there is a lot of discussion about ids being mutable, etc. which needs to be made consistent with discussion of ids elsewhere, and whether they refer to the Thing or the TD, how they are handled in Directories, handling of anonymous TDs, etc.  There is also a problem with mutual authentication and local networks, and also no reference to the Security Best Practices document (which is not however, feasible until we actually publish a first draft!).

Deadline: by end of January.  Anyone who has a thought on additional considerations should add a note here and ideally links to related issues.  We also need to consider where each consideration goes; each WoT deliverable should have a similar issue to this:
* https://github.com/w3c/wot-architecture/issues/672
* https://github.com/w3c/wot-discovery/issues/254

mmccool:
Considerations to add:
- Use of TD metadata to scan for vulnerable devices.  Security metadata is provided.  This is not secret information, but may make it easier for an attacker to identify devices with relatively weak security.  Discussion: this also allows for audits of security.  Mitigation: TDs can be protected from unauthorized access using directories.   Check: is stronger security than what is stated in the TD permitted, i.e. can a TD state that a Thing uses nosec but then requires OAuth using protocol negotiation?  We could add an "undef" scheme to be clearer that security is negotiated using the protocol, but this would require a normative change.  NOTE: currently, under section 8.1, there are assertions REQUIRING the security information in the TD to be accurate and complete, so it's not possible to omit security information; a TD which is not complete in this regard is considered invalid.  So we would either have to soften these or add an exception for the (proposed) "undef" scheme.   Note that "undef" however would be a form of "security through obscurity", and makes it harder to automate when we want to ADD extra security (e.g. forwarding a LAN thing through a proxy to the outside...).
- Use of inappropriate combinations, e.g. passwd with no encryption (e.g. http vs https).  Mitigation: TDs should be audited for where they are deployed and their acceptable threat level and warnings generated.

mmccool:
Another consideration:
- Metadata embedded in ids.  (Check, may be discussed in discovery too).  Ids should not contain embedded metadata such as the type of device or the owner.   In fact to avoid collisions they should be either cryptographically generated or provided during the onboarding process with a hub, etc.  NOTE: some of the examples may still have IDs that violate this rule.  Should check and replace with suitable UUID.

mmccool:
[...]

PR #1360, opened 2022-01-24
Update Security and Privacy Considerations

### LINKED ISSUE #1363: Split "9. Security and Privacy Considerations" into two sections?
from @danielpeintner (also see https://github.com/w3c/wot-thing-description/pull/1359):

Note: we do have 2 remaining issues 

grafik

Properly fixing it would require to split  [9. Security and Privacy Considerations](https://w3c.github.io/wot-thing-description/#sec-security-consideration) into **two** sections 
* Privacy Considerations
* Security Considerations

I don't think we should do that in the current state. We could also simply ignore this check.
Any thoughts?

mmccool:
Currently a placeholder for WIP - do not merge.
* Resolves https://github.com/w3c/wot-thing-description/issues/1348
* Resolves #1363

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1360.html" title="Last updated on Feb 7, 2022, 2:52 PM UTC (1c038a7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1360/114644a...mmccool:1c038a7.html" title="Last updated on Feb 7, 2022, 2:52 PM UTC (1c038a7)">Diff</a>


sebastiankb:
please also consider https://github.com/w3c/wot-thing-description/issues/1363

mmccool:
Updated to implement split into two sections, security and privacy considerations.  Note: the id's still all contain "sec-security-considerations".  I'll probably update the ids as well but only after I check that there I won't break any links (at least *within* the document...).

j1y3p4rk:
I reviewed it, and the privacy part is well written. 

In the security consideration part, I have minor change requests. 

In 9.1 TD Interception and Tempering Security Risk- mitigation part, "mutual authenticated channels" --> "mutual authenticated **secure** channels" because actually mutual authentication does not help to mitigate MITM attack but secure (encrpyted) channel established after mutual authentication does. 

With the same reason, I would like to ask to change in the chapter 9.2 Context Interception and Tempering Security Risk - Mitigation part, "obtained through authenticated channels" --> "obtained through secure channels established by mutual authenication" something like that..

If discussion is needed, we can do that in the security call.

mmccool:
- Probably need a generic security consideration around mitigating all [OWASP Top 10 security](https://owasp.org/Top10/)issues (see also Discovery)

mmccool:
> please also consider #1363

Done

mmccool:
I went ahead and added the following sentence to the end of 9.2:
"If it is necessary to fetch a context file, an implementation may also attempt to use HTTPS (HTTP over TLS) even when only an HTTP URL is given."
Please comment if you feel this is a bad idea and I can revert.  A lot of the time use of an http URL is just a convention (a standardized "name" for the context) and when actually fetching content is it expected/permitted to upgrade security measures.

mmccool:
I wonder also if somewhere in the TD spec (need to check if the JSON-LD spec already says this) it should say that HTTP and HTTPS URLs, if otherwise the same, are assumed to describe the same context.  Note the TD spec itself uses HTTPS context URLs (and really, we do want to avoid downgrading that to HTTP by accident, so a validator should reject HTTP versions, so this has to be worded that an *upgrade* is permitted, not that they are equivalent).

sebastiankb:
from today's TD call:
- group decided to merge this PR and have this for the WD 
-  it is not completed yet, after the WD release there will be further updates
- we need to resolve the conflict before before merge, @mmccool work on this and will merge it later 
-

danielpeintner:
Shall this PR be closed since https://github.com/w3c/wot-thing-description/pull/1387 has been merged?

mmccool:
Closing, replaced with PR #1387

PR #1370, opened 2022-01-28
Delete diff.html

egekorkan:
I am guessing that this should be either deleted or tracked with an issue to be updated. I would still say that this should be deleted and any diff should be simply done via GitHub.

Also see https://github.com/w3c/wot-scripting-api/issues/369


sebastiankb (approved):

sebastiankb:

PR #1254, opened 2021-10-26
fixes #1247 to allow type to be an array

egekorkan:
***
<a href="https://pr-preview.s3.amazonaws.com/egekorkan/wot-thing-description/pull/1254.html" title="Last updated on Nov 3, 2021, 3:55 PM UTC (cd2a52b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1254/f848688...egekorkan:cd2a52b.html" title="Last updated on Nov 3, 2021, 3:55 PM UTC (cd2a52b)">Diff</a>

PR #1388, opened 2022-02-09
revert multiple types

egekorkan:


egekorkan:
Merging as it is actually reverting the PR #1254

Issue #1216, opened 2021-09-01
URLs and IDs should not have embedded metadata


mmccool:
- there needs to be a privacy consideration recommending not embedding device type, etc. into IDs or URLs since these can leak to unauthorized parties
- the examples need to be updated to follow this recommendation (e.g. the first example embeds WoTLamp in the ID which violates this recommendation; a "Lamp" is not a problem, but pretend it was an "InsulinPump", then it would be leaking medical PII)
- recommended practice is to use "random" UUIDs which also helps avoiding collisions (which are a headache for TD Directory services which, unfortunately IMO, use TD IDs as database IDs, so...)

Issue #1401, opened 2022-02-21
Consider Disallowing HTML in Strings to Avoid Injection Attacks


mmccool:
See 

Basically, if HTML markup is allowed in strings such as title and description, and such strings are used to create HTML-based dashboard UIs using a template, then HTML in the TD will be interpreted in the context of the browser session produced by the dashboard generator (i.e. the hub).  It is possible for a malicious TD producer to embed scripts in HTML in various ways, making this a form a cross-site scripting attack.

I am writing a Security Consideration that will recommend that developers generating dashboards sanitize all strings drawn from TDs by replacing HTML special characters like the angle brackets and quotes with ampersand-escaped equivalents.  However, it is my opinion that it would be even better to disallow HTML in strings so a validator will reject them.  This does eliminate one mechanism for handling bidi strings identified in https://www.w3.org/TR/string-meta/ for internationalization, but, oh well: there's no guarantee that strings from TDs will be interpreted in an HTML context anyway, so HTML markup should not be relied up to set the initial direction of strings anyway.

mmccool:
To be clear this is pretty much best practice for web site development to not trust ANY string from outside.  So we don't HAVE to disallow HTML in TD strings, it would just be safer.

mmccool:
See https://github.com/w3c/wot-thing-description/pull/1402.  This includes some text pointing out that if HTML markup is sanitized as it should be, then it cannot be depended upon to set initial text rendering direction.

mmccool:
An argument against disallowing HTML is that HTML is not the only kind of markup that could cause problems, and its impossible to disallow ANY form of markup.  So the onus is on the developer using any external string with care.

sebastiankb:
The [application/td+json Media Type Registration](https://w3c.github.io/wot-thing-description/#media-type-section)  sections already have a security considerations about this topic. Maybe we can also point to this section?

mmccool:
Propose closing.  Can't check for all markup, PR #1402 say that HTML in strings can't be depended on (will normally be sanitized) and that's enough.

mmccool:
I suggest we take option 1, consolidating security considerations down in the S&P sections (and reference the appropriate considerations from the IANA section), and make the S&P considerations normative.  I'll create a separate issue for this since I want to close this one since the specific issue in the description has been resolved already.

PR #1402, opened 2022-02-21
Additional Security and Privacy Considerations

### LINKED ISSUE #1405: Consolidate Security and Privacy Considerations (moving IANA security considerations)
There are security and privacy considerations listed under the IANA considerations that are not listed under the Security Considerations and Privacy Considerations sections, and there are also similar ones that are not cross-referenced.  In addition there are assertions under the IANA section (which appears to be normative) but not under the Security and Privacy sections, which are informative.  There is also an additional consideration, about the sanitization of strings drawn from TDs when used in templates, which is different than any already listed under the IANA considerations (although it is similar to the eval() issue, it is not the same), but seems like it should be part of the IANA considerations also.

Security and Privacy reviews expect the security and privacy considerations to be consolidated.

I propose we:
- Make the Security Consideration and Privacy Considerations sections normative.
- Move the Security and Privacy considerations under the IANA considerations down to the appropriate S&P section, replacing or updating any existing text for similar considerations.
- Adding references to the appropriate S&P Considerations from the IANA considerations.
- Add *normative* assertions as appropriate to other security and privacy considerations, but at least for the string sanitization one to be consistent with the other assertions from existing the IANA section.
- Review the document and consolidate any other security and privacy considerations hiding in other sections

mmccool:
New security and privacy considerations:
- [x] Script Injection security risk.  Related to Issue #1401.  Includes some text under discussion of language handling point out that since it is expected that systems generating HTML will sanitize strings from TDs that HTML cannot be depended upon to control rendering, i.e. by setting initial text direction.
- [x] ID Metadata privacy risk.  Resolves https://github.com/w3c/wot-thing-description/issues/1216
- [x] Vulnerability Auditing security risk (there is an issue for this somewhere, need to track down)
- [x] Consolidate IANA Security Considerations (under consideration).  Would resolve #1405, and would also add an additional Security Consideration discouraging use of eval() to parse TDs, and would also pull in a bunch of assertions, making the S & P sections normative, and would also require adding cross-references.

The Security TF may add more after finishing our review.  Please do not merge until this is marked as complete (all checkboxes marked off above, draft status removed, WIP removed from title).

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1402.html" title="Last updated on Mar 17, 2022, 1:48 PM UTC (e185bf5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1402/1d65bde...mmccool:e185bf5.html" title="Last updated on Mar 17, 2022, 1:48 PM UTC (e185bf5)">Diff</a>


mmccool:
Should probably mention SQL injection under the area for "other markups" in Injection Attacks.  Since it's impossible to check for ALL possible markup languages in strings during TD validation that is not the answer.  USERS of strings drawn from TDs need to check for malicious values before using them.

mmccool:
I would also like to change "Related to" to "Resolved" since I don't think we can add "HTML markup MUST not be used in strings" as a TD assertion; it won't solve the problem anyway.  This PR adds a note that HTML can't be depended upon to solve internationalization problems, since sanitization will disable it.

mmccool:
Propose closing.  Can't check for all markup, PR #1402 say that HTML in strings can't be depended on (will normally be sanitized) and that's enough.

mmccool:
Added requirement to resolve Issue #1405 (assuming option 1, IANA -> S&P, but we could also use option 2, redundant considerations, which would be easier but redundantly redundant.)

Citrullin (approved):

sebastiankb:

Issue #1394, opened 2022-02-13
`name` and `in` fields for `BasicSecurityScheme` and `DigestSecurityScheme` needed?


JKRhb:
Dealing with the implementation of the defined SecuritySchemes, I noticed that the fields `name` and `in` for both the`DigestSecurityScheme` and `BasicSecurityScheme` seem to not correspond to [RFC 7616](https://datatracker.ietf.org/doc/html/rfc7616) and [RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617) where only the inclusion in an HTTP header (whose name is supposed to be `Authorization`) is specified.

Is it intentional for the TD specification to deviate from the RFCs? In [Open API](https://swagger.io/docs/specification/authentication/), where this part of the specification seems to have originated from, these fields are not present, if I am not mistaken.

sebastiankb:
from today's TD call:
- @mmccool @j1y3p4rk can you check this
- we should also check, if there some change needed, if it causes some backwards compatibility problems

JKRhb:
> I am not sure which part is deviating RFCs. From my perspective, it's correct.
> If @JKRhb provides details (e.g., chapter number of RFCs, how it should be in your opinion) will help to understand better.

From my understanding, both RFCs only specify that you can include the credentials in an `Authorization` (or `Proxy-Authorization`) header field (also see [RFC 7235](https://datatracker.ietf.org/doc/html/rfc7235#section-4.2)). I couldn't find any information regarding alternative header names or the inclusion in a query, a cookie, the body etc. but I might be wrong.

I guess the solution for this issue would be to simply remove the `name` and the `in` field from the specification for these two SecuritSchemes or to explicitly state that these fields can be used to create a custom scheme that (probably) deviates from the two RFCs.

mmccool:
Suggest we add a statement that "Usage of these fields MUST be in accordance with RFC-such-and-such".  The problem seems to be additional possible values in schemes which are not allowed in certain protocols.  In general, the values permitted in a scheme must match the protocols of the interactions with which they are associated. 

We should, however, check if we make anything *mandatory* that is explicitly disallowed by a particular RFC.

In general, while we want to maintain backward compatibility if at all possible, fixing any security issues must take precedence.   For TD 2.0 though we probably want to have a complete review and possible do-over.

mmccool:
[...]

PR #1419, opened 2022-03-02
Add `auto` value for the `in` field of `SecurityScheme`s

JKRhb:
As discussed in the latest security calls and in #1394, this PR adds a new possible value called `auto` to the `in` field of `SecurityScheme`s. This has to be seen as a workaround (that probably still requires some discussion) and should be resolved in a more general manner in TD 2.0. For now, this PR can serve as a first step for more discussions on this topic.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-thing-description/pull/1419.html" title="Last updated on Mar 2, 2022, 3:23 PM UTC (4e0ea05)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1419/1d65bde...JKRhb:4e0ea05.html" title="Last updated on Mar 2, 2022, 3:23 PM UTC (4e0ea05)">Diff</a>


egekorkan:
Call of 02.03:
- We will leave more time for review but it looks fine now
- This is a bit of a workaround, we need a better solution for TD 2.0

mmccool:
Discussed in Security Call today:
- We are not sure this is the correct solution to this specific problem (adjusting security schemes to different protocols that may have different locations for information or may want to negotiation locations) and is not a good solution to another problem we have (not wanting to distribute security information in advance to avoid vulnerability scans)
- Having "in = header" as a default is ok for HTTP, actually.
- A better solution to the "vulnerability scanning" problem (which is listed as a security consideration) would be to add an "auto" *SecurityScheme* which would indicate that there IS security (unlike nosec) but that it would be negotiated by the protocol upon a connection attempt (which is now things work right now in nearly all protocols).
- It would be better to have security schemes with protocols; for each protocol, there should be a set of security schemes that match them.  For example, generalizing "basic" to other protocols (that might have different defaults) is the root of the first problem above.  One way to do this (in the future, anyway, for TD2.0) would be to put security scheme names into the protocol-specific namespace, e.g. htv, or a subspace (htvsec).
- Extensions in general are a better way to go since they can easily deal with NEW security schemes that might appear after the spec is published.
- Also, appropriate values for elements in a security scheme (e.g. cryptographic suites) can be included in an extension.  Right now some of these are "strings" to allow for extension but they should really be enumerated types defined in an actual semantic extension (to avoid use of incorrect values).

Conclusion:
- Hold off on merging this - may not be the right answer to the problem of matching schemes to protocols.
- Consider adding an "auto" Security Scheme (would not have any parameters, so easy, and could go into 1.1) - make a PR, discuss on Wed in the TD call? @JKRhb will make a PR.
- Could recommend that non-HTTP protocols use "auto".  Need to check that this is reasonable for CoAP and MQTT at least (but some of the schemes ARE appropriate for these protocols, e.g. PSK for CoAP).  @j1y3p4rk will think about this (and how to fix the first problem, scheme-protocol-matching)

sebastiankb:
from today's TD call, decided to close this PR since there is a new PR provided by @JKRhb 

https://github.com/w3c/wot-thing-description/pull/1421

JKRhb:
Reopened this PR as we will discuss again today // CC @mmccool @sebastiankb

sebastiankb:

PR #1421, opened 2022-03-09
feat: Add AutoSecurityScheme

JKRhb:
This PR reflects the latest discussion from #1394 and #1419 by introducing a new `AutoSecurityScheme` to the specification that could be used to indicate that security parameters are negotiated by the Consumer and the Thing (semi-)automatially. This does not solve #1394 yet but could be a useful addition.


sebastiankb:
from today's TD call:
- it should be evaluated is this relevant for TD 1.1 or should be defer it to TD 2.0
- this PR should be evaluated in one of the next Security TF call (@j1y3p4rk @mmccool @JKRhb @Citrullin)

mmccool:
- For TD2.0 we are looking at a complete revamp to use extensions for more specific schemes that will be specific to protocols, but for now we are only working on PRs meant to be TD 1.x compatible
- This PR is meant for TD 1.1 and it allow a mitigation for people worried about including advance security information in TDs; it is "new", so should be backward-compatible
- The "auto" scheme will be a built-in scheme like "combo" or "nosec" however, and will be available without an extension in TD2.0

index.html
```diff
@@ -1804,6 +1805,7 @@ <h2>Security Vocabulary Definitions</h2>
           <code>nosec</code> (i.e., <code>"scheme":
           "nosec"</code>), indicating there is no authentication or
           other mechanism required to access the resource.</p></section>
+<section><h3><code>AutoSecurityScheme</code></h3><p>An auto authentication security configuration identified by the term <code>auto</code> (i.e., <code>"scheme": "auto"</code>). This scheme can be used to indicate that the security parameters are going to be negotiated by the underlying protocols at runtime. This may require consumers to prompt users for credentials, e. g. if a username and a password are required for authorized access to a resource.</p></section>
```
mmccool, line 1808: I would remove the "This may prompt consumers..." etc. sentence.  

Credentials for "auto" should work the same way as in the other schemes, and we in general don't require consumers to issue prompts to users, since they may not even have UIs.  Generally credentials need to be pre-provisioned to consumers but that is a whole other discussion (let's not go down that rabbit hole here).

ontology/wotsec.ttl
```diff
@@ -99,6 +99,11 @@
     rdfs:subClassOf :SecurityScheme ;
     rdfs:comment "API key authentication security configuration identified by the term <code>apikey</code> (i.e., <code>\"scheme\": \"apikey\"</code>).  This scheme is to be used when the access token is opaque, for example when a key in a proprietary format is provided by a cloud service provider.  In this case the key may not be using a standard token format.  This scheme indicates that the key provided by the service provider needs to be supplied as part of service requests using the mechanism indicated by the <code>\"in\"</code> field."@en .
 
+:AutoSecurityScheme rdf:type rdfs:Class, owl:Class ;
+    rdfs:label "AutoSecurityScheme" ;
+    rdfs:subClassOf :SecurityScheme ;
+    rdfs:comment "An auto authentication security configuration identified by the term <code>auto</code> (i.e., <code>\"scheme\": \"auto\"</code>). This scheme can be used to indicate that the security parameters are going to be negotiated by the underlying protocols at runtime. This may require consumers to prompt users for credentials, e. g. if a username and a password are required for authorized access to a resource."@en .
```
mmccool, line 105: Likewise, remove "prompt" sentence.

index.html
```diff
@@ -8035,6 +8037,32 @@ <h1>JSON Schema for TD Instance Validation</h1>
...
+          "@type": {
+            "$ref": "#/definitions/type_declaration"
+          },
+          "description": {
+            "$ref": "#/definitions/description"
+          },
+          "descriptions": {
+            "$ref": "#/definitions/descriptions"
+          },
+          "proxy": {
```
mmccool, line 8052: do we want descriptions, proxy, etc?  To discuss in security call.  These are optional, however.  *probably* ok, if the TD wants to hide what they are doing and negotiate e.g. the proxy dynamically they can omit these.

validation/td-validation.ttl
```diff
@@ -1369,12 +1369,18 @@
     sh:closed false ;
     sh:order 2 .
 
+:AutoSecuritySchemeShape a sh:NodeShape ;
+    sh:targetClass wotsec:AutoSecurityScheme ;
+    skos:definition """An auto authentication security configuration identified by the term <code>auto</code> (i.e., <code>\"scheme\": \"auto\"</code>). This scheme can be used to indicate that the security parameters are going to be negotiated by the underlying protocols at runtime. This may require consumers to prompt users for credentials, e. g. if a username and a password are required for authorized access to a resource."""^^rdf:HTML ;
```
mmccool, line 1374: remove "prompt" sentence

validation/tm-json-schema-validation.json
```diff
@@ -1049,6 +1049,37 @@
...
+        },
+        "scheme": {
+          "type": "string",
+          "anyOf": [
+            {
+              "enum": [
+                "auto"
+              ]
+            },
+            {
```
mmccool, line 1075: not quite sure why this is here...

mmccool (changes requested):
Suggest removing sentence that says consumer may be required to prompt for credentials.  It's true in some cases (e.g. OAuth2 flows may require it) but is not specific to auto so calling it out here doesn't make sense.  For IoT devices in general we would expect pre-provisioned credentials.  If a consumer does not have a UI but a negotiation results in a UI requirement it would simply have to fail.

But in general I think not bringing up this issue here is the better choice.

validation/tm-json-schema-validation.json
JKRhb: I regenerated the TM schema, therefore it has been integrated alongside the other SecuritySchemes.

sebastiankb:

PR #1432, opened 2022-03-14
split assertion on tmref overwrite

### LINKED ISSUE #1431: tm-tmRef-overwrite double assertion should be split into two
The assertion with id `tm-tmRef-overwrite` and text `If it is intended to override an existing JSON name-value pair definition from tm:ref, the same JSON name MUST be used at the same level of the tm:ref declearation that provides a new value. This process MUST follow the JSON Merge Patch algorithm as defined in [RFC7396] where the content of the referenced definition is patched with the new provided JSON name-value pairs.` should be split into two assertions

egekorkan:
fixes #1431


sebastiankb (approved):

sebastiankb:
should be merged in tomorrows TestFest call

PR #1437, opened 2022-03-15
Default text direction example and assertion

### LINKED ISSUE #1414: No base direction metadata
(citing `Thing` but globally in the document)
https://w3c.github.io/wot-thing-description/#thing
https://w3c.github.io/wot-thing-description/#titles-descriptions-serialization-json

Natural language string values, notably `title` and `description` (and their language map friends `titles` and `descriptions`) do not provide for base direction metadata.

> As defined in [5.3.1.1 Thing](https://w3c.github.io/wot-thing-description/#thing), the base text direction used to display human-readable metadata can either be estimated using heuristics such as the first-strong rule or inferred from language information. In TD documents the default language is defined by a value assigned to \@language in the \@context, and this, along with a script subtag if necessary, can be used to determine a base text direction. However, when interpreting human-readable text, each human-readable string value MUST be processed independently. In other words, a [TD Processor](https://w3c.github.io/wot-thing-description/#dfn-td-processor) cannot carry forward changes in direction from one string to another, or infer direction for one string from another one elsewhere in the TD.

> *Note* Strings on the Web [[STRING-META](https://w3c.github.io/wot-thing-description/#bib-string-meta)] suggests both strong-first and language-based inferencing as means to determine the base text direction. Given that the Thing Description format is based on JSON-LD 1.1 [[json-ld11](https://w3c.github.io/wot-thing-description/#bib-json-ld11)], which currently lacks explicit direction metadata, these approaches are currently considered appropriate at the time of this publication. However, if JSON-LD 1.1 adopts support for explicit base direction metadata as recommended by [[STRING-META](https://w3c.github.io/wot-thing-description/#bib-string-meta)], the Thing Description format should be updated to take advantage of that feature.

JSON LD 1.1 has the `i18n` namespace and can be used to define a compound literal for strings with both language and direction metadata.

https://www.w3.org/TR/json-ld/#the-i18n-namespace
https://www.w3.org/TR/json-ld/#the-rdf-compoundliteral-class-and-the-rdf-language-and-rdf-direction-properties

egekorkan:
fixes #1414

See https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1437.html#example-8 for an easier review

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1437.html" title="Last updated on Apr 5, 2022, 11:05 PM UTC (86e6841)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1437/bafa33e...86e6841.html" title="Last updated on Apr 5, 2022, 11:05 PM UTC (86e6841)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3023,31 +3023,31 @@ In summary, this requires the following:
 		for more detailed information.
 		<pre class="example">
 			{
-					"@context": [
-							"https://www.w3.org/2022/wot/td/v1.1",
-							{ 
-								"@language": "ar-EG",
-							  "@direction": "rtl" 
-							}
-					],
-					"title": "شيء يخصني يقيس درجة الحرارة",
-					"description":"شيء يقيس درجة الحرارة و يظهر حالته",
-					...
-					"properties": {
-							"temp": {
-									"title": "درجة الحرارة",
-									"type": "boolean",
-									"forms": [...]
-							},
-							"status": {
-									"title": "حالة",
-									"type": "object",
-									...
-									"forms": [...]
-							}
-					},
-					...
-			}
+        "@context": [
+            "https://www.w3.org/2022/wot/td/v1.1",
+            { 
+              "@language": "ar-EG",
+              "@direction": "rtl" 
+            }
+        ],
+        "title": "شيء يخصني يقيس درجة الحرارة",
+        "description":"شيء يقيس درجة الحرارة و يظهر حالته",
+        ...
+        "properties": {
+            "temp": {
+                "title": "درجة الحرارة",
+                "type": "boolean",
+                "forms": [...]
+            },
+            "status": {
+                "title": "حالة",
+                "type": "object",
+                ...
+                "forms": [...]
+            }
+        },
+        ...
+    }
 			</pre>
   </p>
 
--- a/index.html
+++ b/index.html
@@ -2975,20 +2975,6 @@ In summary, this requires the following:
     In other words, a <a>TD Processor</a> cannot carry forward changes in direction from one string to another,
     or infer direction for one string from another one elsewhere in the TD.
   </p>
-
-  <p class="note">
-    Strings on the Web [[?STRING-META]] suggests 
-    both strong-first and language-based inferencing as means
-    to determine the base text direction.
-    Given that the Thing Description format is based on JSON-LD 1.1 
-    [[?json-ld11]], which currently lacks explicit direction metadata,
-    these approaches are currently considered appropriate 
-    at the time of this publication. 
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -3022,6 +3008,49 @@ <h3>Human-Readable Metadata</h3>
 }
 </pre>
 
+
+  <p>
+    Strings on the Web [[?STRING-META]] suggests 
+    both strong-first and language-based inferencing as means
```
aphillips, line 3014: `strong-first` should be `first-strong`

index.html
```diff
@@ -3022,6 +3008,49 @@ <h3>Human-Readable Metadata</h3>
 }
 </pre>
 
+
+  <p>
+    Strings on the Web [[?STRING-META]] suggests
```
aphillips, line 3013: We don't actually suggest both first-strong and inference from language tags. We actually recommend metadata falling back to first-strong. About language tags we say:

> This approach is only recommended as a workaround for situations that prevent the use of metadata.

Because you are using JSON-LD, you have metadata for the document as a whole and for individual string values (notably `title` and `description`). What you don't have is per-string direction in language maps (e.g. `titles` and `descriptions`) and you could recommend the use of the language tag for these.

I would thus tend to say:

```
Strings on the Web recommends the use of metadata to determine the <a>base 
direction</a> of string values. When metadata is not present, applications SHOULD 
use <a>first-strong detection</a>as a fallback. Language maps do not permit per-string
metadata: implementations MAY infer the base direction from the language tag (as
a better indicator than first-strong detection).
```



Note that the <a> links are found in `i18n-glossary` (you can add to your xrefs to pick up automagically)

index.template.html
```diff
@@ -1845,6 +1831,49 @@ <h3>Human-Readable Metadata</h3>
 }
 </pre>
 
+
+  <p>
+    Strings on the Web [[?STRING-META]] suggests 
+    both strong-first and language-based inferencing as means
+    to determine the base text direction.
```
aphillips, line 1838: same comments

aphillips (changes requested):

egekorkan:
@aphillips I have addressed your feedback. However, there is one small issue: we are using these definitions inside normative assertions but these definitions are informative in i18-glossary (which makes sense since it is just a glossary). Should I change something in the respec config?


**Important note for spec editors**: I have changed the respec config

aphillips (approved):
Looks good to me! Thanks!

sebastiankb:
many thanks!

PR #1442, opened 2022-03-20
add new context url to tm schema

egekorkan:
Our new context URL `"https://www.w3.org/2022/wot/td/v1.1"` was not allowed in TMs, fixing this.


danielpeintner (approved):

sebastiankb (approved):

sebastiankb:

Issue #1325, opened 2021-12-14
clarify whether vocabulary terms in the information model are case sensitive


mlagally:
@egekorkan 
> I don't think it makes sense to write something like this normatively, especially if it is only for the `op` values. The spec says:

>  Future versions of the standard may extend this list but operations types SHOULD NOT be arbitrarily set by servients and be restricted to the values in the table below.

I agree with you, that it should not be only for the op values, a clarification is required for all TD literals. 

> If we are going to put something about case sensitiveness, there is already some at https://w3c.github.io/wot-thing-description/#td-basic-types-mapping which already implies this.

> But we can add a non-assertion text at the end of this section saying: 
> This means that all terms are case sensitive.

This section is talking about the JSON binding, but IMHO a common normative clarification should be done in chapter 5.

sebastiankb:
from today's TD call:
- group decided to mention in the TD spec that all key-value pairs and the sample values provided in the information model is case sensitive
- best place would be in Section 5.3 the explains this and is valid for the full TD model

sebastiankb:
merged

PR #1444, opened 2022-03-21
add text about case sensitive

sebastiankb:
Solve #1325

### DIFF
--- a/index.html
+++ b/index.html
@@ -1131,6 +1131,11 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
             </span>
         </p>
 
+        <p>
+            Please note that all following vocabulary terms and their choice values (if exists) are case sensitive. 
+            This is also true for the serialization of the information model (Section <a href="#sec-td-serialization"></a>).
+        </p>
+
       <section id="sec-core-vocabulary-definition">
       <h2>Core Vocabulary Definitions</h2>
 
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1131,6 +1131,11 @@ <h2>Class Definitions</h2>
             </span>
         </p>
 
+        <p>
+            Please note that all following vocabulary terms and their choice values (if exists) are case sensitive.
```
danielpeintner, line 1135: Maybe we can simplify it.

> Please note that all vocabulary terms and values are case sensitive.

sebastiankb:
from today's TD call:
- not going to merge
- we should evaluate the proposed sentence from Daniel if it has some impact on specific applications (e.g., do not take care if case sensitive)

sebastiankb:
I just followed the proposal from @danielpeintner and updated the text

index.html
```diff
@@ -1131,6 +1131,11 @@ <h2>Class Definitions</h2>
             </span>
         </p>
 
+        <p>
+            Please note that all vocabulary terms and values are case sensitive.
+            This is also true for the serialization of the information model (Section <a href="#sec-td-serialization"></a>).
```
danielpeintner, line 1136: I am just wondering. You are referring to "information model" and linking to the representation format. Is this intended?

sebastiankb:

index.html
danielpeintner: I mean... either the text is wrong or the link is wrong

Issue #1406, opened 2022-02-23
typo in text of dataResponse


sebastiankb:
>Defines the data schema of the Event response messages sent **be** the ...

--> 

> Defines the data schema of the Event response messages sent **by** the ...

JKRhb:
I accidentally linked the wrong issue in #1447, where this issue has been fixed. I will therefore close this one manually :)

PR #1447, opened 2022-03-23
chore: re-render index.html

JKRhb:
In #1421, there are currently a lot of unrelated changes as the rendered files seem not up to date. This PR simply re-renders the document as well as SVG and context files. This cleans up other PRs (after rebasing them).

Maybe the document on the main branch could be updated automatically after each PR to avoid this in the future?


sebastiankb:
Thanks for this PR. It seems that the complete content of dataResponse is removed with this PR. I think this is a mistake.

JKRhb:
Oh, you are right! I guess there is an error in an ontology file?

JKRhb:
@sebastiankb Thanks once again for your comment, I found the reason for the issue (a missing line in `merge.js`). The PR should now be ready :)

JKRhb:
I now included a fix for #1406 in this PR

sebastiankb:

Issue #1412, opened 2022-03-01
Languages of service-doc?


aphillips:
Example 27
https://w3c.github.io/wot-thing-description/#example-link-to-developer-documentation

> To point to a developer documentation of a [Thing](https://w3c.github.io/wot-thing-description/#dfn-thing) the value service-doc can be used:

```
"links": [{
    "rel": "service-doc",
    "href": "https://example.com/howTo",
    "type": "application/pdf"
}]
```

The example shows how to link to service documentation. Perhaps this should include an affordance for indicating the language(s) of the documentation and/or ways to indicate different language versions of the documentation?

egekorkan:
That is a valid concern, generally devices come with:
1. A single pdf (or other document formats) with multiple languages: This would be a single link with an array of languages
2. Different pdf for different languages: This would be an array of links with a language tag for each
3. Multiple pds: Combination of the two above

egekorkan:
Call of 02.03.2022:
- We would need an annotation that can be a single string or an array. There is an HTML attribute that can provide this

sebastiankb:
From today's TD call:
we should evaluate the following aspects:
- we should check if the web linking (rfc8288) already provides such a hint in links definition
- how html is handle this if links points to a pdf documents as example (it seems that there is tag about this)
- how is the relation of the underlying protocol like HTTP (language negation)

aphillips:
@sebastiankb 

The HTML attribute is `hreflang` and can be used on `link` and `a` elements in HTML. You might include an optional field or fields `hreflang` into the link structure. The `hreflang` can indicate the language of any target: they don't have to be HTML pages, any old mimetype will do.

RFC8288 supports `hreflang`. Here's one quote:

>    The "hreflang" attribute, when present, is a hint indicating what the language of the result of dereferencing the link should be.  Note that this is only a hint; for example, it does not override the Content-Language header field of a HTTP response obtained by actually following the link.  Multiple hreflang attributes on a single link-value indicate that multiple languages are available from the indicated resource.

The attribute doesn't participate in or affect HTTP; there is no guarantee that the tag is at all accurate. Language negotiation could still take place in addition to/in spite of the attribute.

egekorkan:
Closing after #1449

PR #1449, opened 2022-03-28
adds hreflang attribute to links

sebastiankb:
solves #1412


index.html
```diff
@@ -1956,7 +1956,8 @@ <h2>Hypermedia Controls Vocabulary Definitions</h2>
...
-                    {Height}x{Width} (e.g., "16x16", "16x16 32x32").</td><td>optional</td><td><a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a></td></tr></tbody></table> <p>Link relations can be used to describe relations such as to other Things (e.g., a Switch Thing controls a Lamp Thing), to a specific kind of Thing Models (e.g., a Thing Description
+                    {Height}x{Width} (e.g., "16x16", "16x16 32x32").</td><td>optional</td><td><a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a></td></tr>
+<tr class="rfc2119-table-assertion" id="td-vocab-hreflang--Link"><td><code>hreflang</code></td><td>The hreflang attribute specifies the language for a linked document. The value of this attribute should conform to [[RFC5646]].</td><td>optional</td><td><a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#string"><code>string</code></a></td></tr></tbody></table> <p>Link relations can be used to describe relations such as to other Things (e.g., a Switch Thing controls a Lamp Thing), to a specific kind of Thing Models (e.g., a Thing Description
```
aphillips, line 1960: BCP47 is a better reference than RFC5646, since it is durable (5646 replaced 4646 replaced 3066 replaced 1766...)

"should conform to" is a little obscure? I would probably say "should be a valid language tag [[BCP47]]". Note that the term "valid" has a specific meaning in BCP47.

ontology/hctl.html
```diff
@@ -110,6 +110,7 @@ <h2>Introduction</h2>
                 name of a previous definition given in a 
                 <code>schemaDefinitions</code> map must be used.</span><table class="def"><tbody><tr><td>Domain includes</td><td><code><a href="#AdditionalExpectedResponse">hctl:AdditionalExpectedResponse</a></code></td></tr></tbody></table></section>
 <section id="hasAnchor"><h4>hasAnchor</h4><p>IRI: <code>https://www.w3.org/2019/wot/hypermedia#hasAnchor</code></p><span>By default, the context, or anchor, of a link conveyed in the Link header field is the URL of the representation it is associated with, as defined in RFC7231, Section 3.1.4.1, and is serialized as a URI.</span><table class="def"><tbody><tr><td>Domain includes</td><td><code><a href="#Link">hctl:Link</a></code></td></tr></tbody></table></section>
+<section id="hasHreflang"><h4>hasHreflang</h4><p>IRI: <code>https://www.w3.org/2019/wot/hypermedia#hasHreflang</code></p><span>The hreflang attribute specifies the language for a linked document. The value of this attribute should conform to [[RFC5646]].</span><table class="def"><tbody><tr><td>Domain includes</td><td><code><a href="#Link">hctl:Link</a></code></td></tr></tbody></table></section>
```
aphillips, line 113: same comment here as above.

Nit: perhaps `the language for a linked document` => `the language of a linked document`?

ontology/hctl.ttl
```diff
@@ -95,6 +95,11 @@ For example, for HTTP and Events, it indicates which of several available mechan
     rdfs:label "hasSizes" ;
     schema:domainIncludes :Link .
 
+:hasHreflang rdf:type owl:ObjectProperty ;
+    rdfs:comment "The hreflang attribute specifies the language for a linked document. The value of this attribute should conform to [[RFC5646]]." ;
```
aphillips, line 99: ibid.

validation/td-validation.ttl
```diff
@@ -1009,6 +1009,13 @@
         sh:maxCount 1 ;
         sh:order 5 ;
     ] ;
+    sh:property [
+        sh:path hctl:hasHreflang ;
+        skos:definition """The hreflang attribute specifies the language for a linked document. The value of this attribute should conform to [[RFC5646]]."""^^rdf:HTML ;
```
aphillips, line 1014: ibid

aphillips (changes requested):
Minor comments about how to reference bcp47 and some editorial nits.

validation/td-validation.ttl
```diff
@@ -1009,6 +1009,13 @@
         sh:maxCount 1 ;
         sh:order 5 ;
     ] ;
+    sh:property [
+        sh:path hctl:hasHreflang ;
+        skos:definition """The hreflang attribute specifies the language of a linked document. The value of this should be a valid language tag [[BCP47]]."""^^rdf:HTML ;
```
aphillips, line 1014: etc.

aphillips (approved):
I'm good with the text as is, but would suggest changing 'should' to 'must' as noted.

egekorkan:
> But this would break the compatibility to RFC8288. Does a multi-language tag exists?

Good point. The RFC says:

> Multiple hreflang attributes on a single link value indicate that multiple languages are available from the indicated resource.

However, some [digging up](https://webmasters.stackexchange.com/questions/98172/can-multiple-hreflang-in-a-page-point-to-the-same-url) made me understand that this means multiple link with their own hreflang pointing to the same link means that. Thus, my change request is not valid. To say that two languages are available in on link, one would have a TD like the following:

```json
{
"links": [{
    "rel": "service-doc",
    "href": "https://example.com/manual",
    "hreflang":"de"
    "type": "application/pdf"
},
{
    "rel": "service-doc",
    "href": "https://example.com/manual",
    "hreflang":"nl"
    "type": "application/pdf"
}]
}
```

egekorkan (approved):

aphillips:
I was reviewing [this I-D](https://datatracker.ietf.org/doc/draft-ietf-httpapi-linkset/) which solves this problem by making `hreflang` an array. This saves repeating all of the other link information. Compatibility with 8288 can be achieved by deserializing the JSON as a list of links or by deserializing the array as multiple `hreflang` entries. RFC8288 says that `hreflang` can be repeated (which can't be done in JSON):

> Multiple hreflang attributes on a single link-value indicate that multiple languages are available from the indicated resource.

@sebastiankb there is a tag for "multiple languages" (it's `mul`), but that's not really very useful, since it doesn't tell you which languages those are. If your Thing has a manual for North America in English/Spanish/French and one for western Europe in (say) English/French/German/Spanish/Italian, both would be labeled `mul`.

sebastiankb:

PR #1452, opened 2022-03-29
fix: allow `uri` value only for `in` field of `APIKeySecurityScheme`

### LINKED ISSUE #1440: TD Schema does not allow for SecurityScheme property "in" to have the value "uri"
Currently, the TD Schema does not include the value `"uri"` as a possible value for the `"in"` property of SecuritySchemes
See: [1](https://github.com/w3c/wot-thing-description/blob/bafa33e47c84bb879e1f5a102649cc3bea1ec3ca/validation/td-json-schema-validation.json#L873), [2](https://github.com/w3c/wot-thing-description/blob/bafa33e47c84bb879e1f5a102649cc3bea1ec3ca/validation/td-json-schema-validation.json#L918), [3](https://github.com/w3c/wot-thing-description/blob/bafa33e47c84bb879e1f5a102649cc3bea1ec3ca/validation/td-json-schema-validation.json#L956), [4](https://github.com/w3c/wot-thing-description/blob/bafa33e47c84bb879e1f5a102649cc3bea1ec3ca/validation/td-json-schema-validation.json#L1003). These will cause any TD that uses this feature to not validate properly.

JKRhb:
This PR adds a separate `in` definition for the `APIKeySecurityScheme` that includes the newly `uri` value but removes it from the `in` field of the other `SecurityScheme`s.

I am not entirely sure if this is the best way to handle this problem – it would probably more elegant if you could simply extend the "regular" `in` field with another value but it seemed to me as if this is not possible.

Resolves #1440 when merged.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-thing-description/pull/1452.html" title="Last updated on Apr 27, 2022, 2:38 PM UTC (75976c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1452/404767a...JKRhb:75976c4.html" title="Last updated on Apr 27, 2022, 2:38 PM UTC (75976c4)">Diff</a>


JKRhb:
Hmm, it seems as if there is a problem with the JSON-LD context as you currently can't define two SecurityScheme fields with the same name at the moment. Therefore, this PR might require some more restructuring.

egekorkan:
Regarding the final result (index.html), it seems almost fine but `in` for APIKey should either have a default value or be mandatory no?

sebastiankb:
there is a comment and merge conflicts. @JKRhb would be great if you have a look

JKRhb:
> there is a comment and merge conflicts. @JKRhb would be great if you have a look

Thank you for the ping and for your comment, @egekorkan! I resolved the conflict now and regenerated the document, noticing a number of strange issues that need to be addressed before merging. These issues include the disappeared default value and vocabulary term. For the sake of backwards compatibility, the default value should probably remain `query`, right?

egekorkan:
@JKRhb there are conflicts again :( Regarding the default value, you are correct, https://www.w3.org/TR/wot-thing-description/#sec-default-values also has query

JKRhb:
I think the PR should now finally be ready :) Could you have another look, @egekorkan, @sebastiankb?

PR #1454, opened 2022-03-31
Security Consideration Update

mmccool:
Some minor fixes:
- fix typo in section reference (missing quote)
- add local ACE-OAuth ref (was not in Specref; however, as noted in ednote, will likely be replaced with a real IETF RFC by the time we go to CR, at which point we can remove the local ref and/or update it to point at the actual RFC)

### DIFF
--- a/index.html
+++ b/index.html
@@ -51,6 +51,19 @@
                 }
               ]
             , localBiblio: {
+                "ACE-OAuth": {
+                  title: "Authentication and Authorization for Constrained Environments (ACE) using the OAuth 2.0 Framework (ACE-OAuth)"
+		, authors: [
+                    "Ludwig Seitz"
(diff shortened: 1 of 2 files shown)


JKRhb:
I just noticed that there is an exisiting `[ACE]` citation/bibliography entry referencing an older version of the draft. That one should probably be replaced with the new one, I suppose?

JKRhb:
(The old draft is referenced [here](https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1454/b1d3bee...mmccool:0ca185e.html#adding-security-schemes).)

JKRhb (approved):
Thank you, @mmccool, changes look good to me! :)

sebastiankb (approved):

PR #1458, opened 2022-04-05
Constraining context value according to new algorithm

### LINKED ISSUE #1420: Restricting the context url in JSON Schema
Rather easy but urgent decision needs to be made before the testfest/plugfest of next week:

Should we allow only the new context URL? Currently at https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json#L67 three are allowed, I think only the latest should be allowed?

egekorkan:
fixes #1420

I had to introduce new terms to handle this properly. I have added comments to the schema since it is not obvious at all. I have done all the combinations I can think of. One is available at https://www.jsonschemavalidator.net/s/wmGwzNWP

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -61,21 +61,26 @@
           "sse"
         ]
       },
-      "thing-context-w3c-uri": {
+      "thing-context-td-uri-v1": {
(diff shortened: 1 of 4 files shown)


sebastiankb:

PR #1466, opened 2022-04-13
use "@type" instead of rdf:type

ethieblin:
This is necessary for JSON-LD framing of the TD.
The security triples are not framed back correctly with "rdf:type" instead of "@type" in the context.
The input data in these examples is a set of security RDF triples which have been transformed into JSON-LD with rdflib (python library for rdf).
The same output is obtained after uploading and retrieving the triples from a SPARQL endpoint.
```
<urn:uuid:55f01138-5c96-4b3d-a5d0-81319a2db677> a td:Thing .
<urn:uuid:55f01138-5c96-4b3d-a5d0-81319a2db677> td:definesSecurityScheme _:b0 .
<urn:uuid:55f01138-5c96-4b3d-a5d0-81319a2db677> td:hasSecurityConfiguration> <https://json-ld.org/playground/nosec_sc> .
_:b0 a wotsec:NoSecurityScheme .
_:b0 td:hasInstanceConfiguration <https://json-ld.org/playground/nosec_sc> .
```

Framing with the original context gives [playground with original context](https://tinyurl.com/34w9529x)
```json
"securityDefinitions": {
    "nosec_sc": {
      "@type": "nosec"
    }
  }
```
Framing with the new context gives [playground with this MR context](https://tinyurl.com/3v2axww4)
```json
"securityDefinitions": {
    "nosec_sc": {
      "scheme": "nosec"
    }
  }
```

### DIFF
--- a/context/td-context-1.1.jsonld
+++ b/context/td-context-1.1.jsonld
@@ -222,8 +222,7 @@
           "@id": "wotsec:oneOf"
         },
         "scheme": {
-          "@id": "rdf:type",
-          "@type": "@vocab"
+          "@id": "@type"
         },
         "description": {
           "@id": "td:description"
@@ -311,8 +310,7 @@
           "@id": "wotsec:oneOf"
         },
         "scheme": {
-          "@id": "rdf:type",
-          "@type": "@vocab"
+          "@id": "@type"
         },
         "description": {
           "@id": "td:description"
--- a/context/wot-security-context.jsonld
+++ b/context/wot-security-context.jsonld
@@ -54,8 +54,7 @@
       "@id": "wotsec:oneOf"
     },
     "scheme": {
-      "@id": "rdf:type",
-      "@type": "@vocab"
+      "@id": "@type"
     },
     "description" : {
       "@id" : "td:description"


ethieblin:
@wiresio fyi

sebastiankb:
from today's TD call:
- it seems that the render script needs to be updated to have the change permanent
- the td-context-1.1.jsonld is generated by the render script and should not change manually
- @ethieblin can you update the script as well?

ethieblin:
@sebastiankb I manually changed [context/wot-security-context.jsonld](https://github.com/w3c/wot-thing-description/pull/1466/files#diff-a6f3215191a57600aaec3473c2220f1eea8640f66089ff8d5bd27045fc5be614) then ran the script merge.js to get the td-context-1.1.jsonld changes.
I will check that the other scripts of render.sh are fine with this change.

ethieblin:
I updated the toRDF.js file so that the ttl file do not change (but in the other PR, I can move the commit or merge the two PR into one if you prefer): https://github.com/w3c/wot-thing-description/pull/1468/commits/5824fe3a6a881772bcb0d8a6cfda90c3c88bc0b8

sebastiankb:

PR #1468, opened 2022-04-13
feat(context): use "@type" instead of "rdf:type"

ethieblin:
Similarly to https://github.com/w3c/wot-thing-description/pull/1466

The input data of these examples was generated by transforming a TD into RDF and the RDF into json-ld with rdflib
(same result when uploading/downloading the RDF triples from a SPARQL endpoint)

`"type"` does not round trip on the TD properties type

[playground with original context](https://tinyurl.com/2n4xe7yc)
```
"properties": {
    "airpressure": {
      "@type": "number",
      }
 }
 ```
[playground with new context](https://tinyurl.com/4xxd6y6p)

```
"properties": {
    "airpressure": {
      "type": "number",
      }
 }
 ```

### DIFF
--- a/context/json-schema-context.jsonld
+++ b/context/json-schema-context.jsonld
@@ -87,8 +87,7 @@
       "@container": "@set"
     },
     "type": {
-      "@id": "rdf:type",
-      "@type": "@vocab"
+      "@id": "@type"
     },
     "title" : {
       "@id" : "td:title"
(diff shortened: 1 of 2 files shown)


ethieblin:
@wiresio fyi

sebastiankb:
same situation: see https://github.com/w3c/wot-thing-description/pull/1466#issuecomment-1111101423

sebastiankb:
same situation: see https://github.com/w3c/wot-thing-description/pull/1466#issuecomment-1111101423

sebastiankb (approved):

PR #1470, opened 2022-04-20
Create assertion-sync.yml

egekorkan:
This will need an update later but I wanted to put it here already.

This generates the template.csv every day if there is a change


sebastiankb:

sebastiankb:

Issue #1461, opened 2022-04-08
Allow `SecurityScheme` extenstions in JSON Schema


relu91:
**context**: *In node-wot we are trying to communicate with [tuya](https://www.tuya.com/) devices. See https://github.com/eclipse/thingweb.node-wot/pull/735*

As [Section 7.2](https://w3c.github.io/wot-thing-description/#adding-security-schemes) regulates it should be possibile to add new security schemas with proper context extensions. However, the JSON schema strictly checks for only the set of SecuritySchemas defined inside the Thing Description specification.

PR #951, opened 2020-08-14
Updating rendering script 

vcharpenay:
Rendering is now more closely aligned with the SHACL schema language and several mistakes in `validation/td-validation.ttl` have been fixed. Documentation of rendering is available as comments in `render.sh`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/951.html" title="Last updated on Sep 19, 2020, 12:13 AM UTC (93c074f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/951/32ba69e...93c074f.html" title="Last updated on Sep 19, 2020, 12:13 AM UTC (93c074f)">Diff</a>

PR #1474, opened 2022-04-24
feat: allow definition of additional security schemes

JKRhb:
This PR is a potential fix for https://github.com/w3c/wot-thing-description/issues/1461. It allows the definition of "generic" security schemes that don't fall into the pre-defined categories. This allows, for example, to add the `ACESecurityScheme` from the appendix to a TD (see [here](https://www.jsonschemavalidator.net/s/Y4SoCSOy) for a "demonstration").

There is no check included that makes sure that the new security schemes correspond with a context extension yet. Maybe such a check could also be added in a follow-up PR, though, as the general possibility of adding additional schemes probably has a higher priority IMHO.


validation/td-json-schema-validation.json
```diff
@@ -786,6 +786,42 @@
...
+                "basic",
+                "digest",
+                "bearer",
+                "psk",
+                "oauth2"
+              ]
+            }
+          }
+        },
+        "required": [
```
egekorkan, line 821: How about making `@type` required. Otherwise people can put anything in scheme with no way for others to understand it.

egekorkan (changes requested):

validation/td-json-schema-validation.json
JKRhb: Sure, that would be a possibility :) Then I guess we would need to adjust [Example 49](https://w3c.github.io/wot-thing-description/#example-49).

egekorkan:
> Sure, that would be a possibility :) Then I guess we would need to adjust [Example 49](https://w3c.github.io/wot-thing-description/#example-49).

I decided it's better to put this comment here instead of the review comment since it's a general design decision kind of problem.

Well that was the case I was thinking about. That example is theoritically and practically correct. However, I don't know of a way to check the presence of a context field in case an additional scheme is used. We can put an assertion and then put a JS script in Playground to check that but not everyone uses playground. 

Here are the solutions I have in mind:
- Do nothing more to this PR and merge: This allows someone who has a custom security scheme to put `scheme:bread` which could be an actual mechanism needed to communicate with the Thing. This would have negative impact on interoperability.
- Require `@type`: Like in the linked example above, this is not needed and this would also make it possible to have `@type : bread` . It would maybe help raise awareness...
- Find a way to check context values together with scheme values or `@type` : This would be 90% correct but this would not find errors in example 49, which is actually incorrect given the example urls that do.not have ace security information. Even though I don't know how to do this with JSON Schema, it might be possible
- Do full semantic validation: That would be the ideal case but also unlikely that people to this validation, so it'll be useless.

JKRhb:
Maybe a workaround could be that additional `scheme`s must have a prefix like `ace:`. This could probably be checked for easily. An additional assertion in the document could then specify that additional security schemes must refer to context extensions?

index.html
```diff
@@ -8343,6 +8343,42 @@ <h1>JSON Schema for TD Instance Validation</h1>
...
+            "$ref": "#/definitions/description"
+          },
+          "descriptions": {
+            "$ref": "#/definitions/descriptions"
+          },
+          "proxy": {
+            "$ref": "#/definitions/anyUri"
+          },
+          "scheme": {
+            "type": "string",
```
JKRhb, line 8362: > Maybe a workaround could be that additional schemes must have a prefix like ace:. 

I suppose this could look like the following:

```suggestion
            "type": "string",
            "pattern": ".*:.*",
```

JKRhb: Experimenting in the JSON-LD playground, I noticed, though, that compact IRIs in places other than the @type do not actually get expanded (by default?).

egekorkan: Regarding the pattern check, it is also a not bad option but it would still allow people to do `"scheme":"bakery:bread"`

Regarding `@type` , [example 43](https://w3c.github.io/wot-thing-description/#example-43) has all the possible usage of external vocabularies.

mmccool:
To be consistent with TD1.0, and with the example, we *should* allow extensions.  The question is how to constrain them so they are meaningful.  Requiring "extended" schemes need a context file and use a prefix would be a useful.

HOWEVER, I'm not sure we want to add new normative statements at this point.  We DO want to fix bugs, though: and allowing only a fixed set of schemes in one place and then saying you can use extensions in another place is such a "bug".  We need to use "e.g." for the set of schemes, not "one of", at the very least.  Cleaning up/constraining the syntax for extensions etc., adding a "generic" scheme, etc. may have to wait until TD 2.0.

mmccool:
So I've confirmed the "BUG" exists, and propose (due to timing) that we do only the minimum to solve this bug: "any of" should be changed back to "e.g." (like it was in TD 1.0).  Yes, it's far from perfect, but....

egekorkan:
We cannot change the `enum` in the scheme to `examples` since each value of the enum has its specific associated structure. What @JKRhb does is not adding a new scheme called genericScheme. It is only changing the JSON Schema file (which is informative) and allows the scheme to have any value

JKRhb:
I now found the PR that replaced the "e.g." with "one of": https://github.com/w3c/wot-thing-description/pull/951 The PR also changed `scheme` from mandatory to optional, which is also another bug I noticed only now.

relu91 (approved):

JKRhb:
Thank you for your feedback, @relu91 :) Should we also include the mandatory prefix in the `scheme` field for additional security schemes, though? This would mitigate the risk of typos in the "pre-defined" security schemes.

Regarding the bug I found in https://github.com/w3c/wot-thing-description/pull/1474#issuecomment-1109171687, I would open another issue for a follow-up PR as it seems that there is something wrong with the render script and/or the ontology.

index.html
```diff
@@ -8343,6 +8343,42 @@ <h1>JSON Schema for TD Instance Validation</h1>
           }
         ]
       },
+      "genericSecurityScheme": {
```
JKRhb, line 8346: This could also be changed to `additionalSecurityScheme`, indicating that it is not one of the pre-defined ones.

JKRhb: I changed the name to `additionalSecurityScheme` now as I find it to be a bit clearer.

index.html
JKRhb: I now added a (slightly adjusted) pattern check as well as a comment (as you described in #1482) and the example from the specification. While I still think that requiring a @type would be useful, I suppose that it is not covered by the current assertions and would also cause potential incompatibility with TD 1.0. Therefore, I would suggest not adding this to the schema but keeping it in mind for TD 2.0 as an additional assertion.

JKRhb:
I now made some adjustments to the PR that should address the suggestions from @egekorkan and also fix the TD document itself :) Looking forward to your feedback.

egekorkan (approved):

sebastiankb (approved):

PR #1475, opened 2022-04-25
Create Automatic Issue Labeling Action

egekorkan:
This PR triggers a very small job that adds `needs-triage` label to each issue that is opened. We can then use this to filter issues that we did not have time to discuss in a meeting or for an editor to categorize.


sebastiankb:

Issue #931, opened 2020-07-22
Content Type for Thing Model (IANA registration needed)


sebastiankb:
In today's TD call we were thinking about own content type for TDT. That would help to distinguish between TD and TDT when both are exchanged or requested.

sebastiankb:
I will check with IANA and ask for the registration steps.

sebastiankb:
A first draft for IANA (simple c&p from TD and replace TD with TM): 

application/tm+json Media Type Registration

Type name:
    application

Subtype name:
    tm+json

Required parameters:
    None

Optional parameters:
    None

Encoding considerations:
    See RFC 6839, section 3.1.

Security considerations:
    See RFC 8259, section 12.

    Since a WoT Thing Model is intended to be a pure data exchange format for Thing metadata, the serialization SHOULD NOT be passed through a code execution mechanism such as JavaScript's eval() function to be parsed. An (invalid) document may contain code that, when executed, could lead to unexpected side effects compromising the security of a system.

    WoT Thing Model can be evaluated with a JSON-LD 1.1 processor, which typically follows links to remote contexts (i.e., TD context extensions, see W3C WoT Thing Description, section 7) automatically, resulting in the transfer of files without the explicit request of the Consumer for each one. If remote contexts are served by third parties, it may allow them to gather usage patterns or similar information leading to privacy concerns. While implementations on resource-constrained devices are expected to perform raw JSON processing (as opposed to JSON-LD processing), implementations in general SHOULD statically cache vetted versions of their supported context extensions and not to follow links to remote contexts. Supported context extensions can be managed through a secure software update mechanism instead.

    Context Extensions (see W3C WoT Thing Description, section 7) that are loaded from the Web over non-secure connections, such as HTTP, run the risk of being altered by an attacker such that they may modify the TD Information Model in a way that could compromise security. For this reason, Consumer again SHOULD vet and cache remote contexts before allowing the system to use it.

    Given that JSON-LD processing usually includes the substitution of long IRIs [RFC3987] with short terms, WoT Thing Models may expand considerably when processed using a JSON-LD 1.1 processor and, in the worst case, the resulting data might consume all of the recipient's resources. Consumers SHOULD treat any TD metadata with due skepticism.

Interoperability considerations:
    See RFC 8259.

    Rules for processing both conforming and non-conforming content are defined in this specification.
Published specification:
    https://www.w3.org/TR/wot-thing-description11/

Applications that use this media type:
[...]

PR #1476, opened 2022-04-27
Add section about IANA Thing Model content type registration

sebastiankb:
solves #931

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1476.html" title="Last updated on May 12, 2022, 7:40 AM UTC (76398ed)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1476/a76e087...76398ed.html" title="Last updated on May 12, 2022, 7:40 AM UTC (76398ed)">Diff</a>


sebastiankb:
@plehegar if this looks ok for you I would start to submit the official registration request at IANA.

plehegar (approved):

index.html
```diff
@@ -7277,8 +7277,78 @@ <h2><code>application/td+json</code> Media Type Registration</h2>
...
+         </dd>
+          <dt>Applications that use this media type:</dt>
+          <dd>
+            All participating entities in the W3C Web of Things, that is,
+            <a>Things</a>, <a>Consumers</a>, and Intermediaries as defined in the
+            <a href="https://www.w3.org/TR/wot-architecture11/">W3C 
+        Web of Things (WoT) Architecture 1.1</a> document.
+          </dd>
+          <dt>Fragment identifier considerations:</dt>
+          <dd>See <a href="https://tools.ietf.org/html/rfc6839#section-3.1">RFC&nbsp;6839, section 3.1</a>.
```
mmccool, line 7325: JSON Pointer is [RFC6901](https://w3c.github.io/wot-thing-description/#bib-rfc6901)].
RFC 6839 is just general guidance on how tags work but says that "if there is an extension the fragment identifier can be defined" but the parent is otherwise inherited, but JSON itself (the parent) does not define a fragment identifier.

I think we should *explicitly* state that JSON Pointer (the RFC above) is the default fragment identifier for TMs (AND TDs...)

index.html
```diff
@@ -7277,8 +7277,78 @@ <h2><code>application/td+json</code> Media Type Registration</h2>
...
+          <dt>Subtype name:</dt>
+          <dd>tm+json</dd>
+          <dt>Required parameters:</dt>
+          <dd>None</dd>
+          <dt>Optional parameters:</dt>
+          <dd>None</dd>
+          <dt>Encoding considerations:</dt>
+          <dd>See <a href="https://tools.ietf.org/html/rfc6839#section-3.1">RFC&nbsp;6839, section 3.1</a>.
+          </dd>
+          <dt>Security considerations:</dt>
```
mmccool, line 7296: ok, this looks good, I don't know why I saw the old text when I looked last

index.html
```diff
@@ -7277,8 +7277,78 @@ <h2><code>application/td+json</code> Media Type Registration</h2>
...
+          <dt>Fragment identifier considerations:</dt>
+          <dd>See <a href="https://tools.ietf.org/html/rfc6839#section-3.1">RFC&nbsp;6839, section 3.1</a>.
+          </dd>
+          <dt>Additional information:</dt>
+          <dd>
+            <dl>
+              <dt>Magic number(s):</dt>
+              <dd>Not Applicable</dd>
+              <dt>File extension(s):</dt>
+              <dd>.jsontm, .tm.json, .tm.jsonld</dd>
```
mmccool, line 7333: ideally we would say .tm.jsonld is preferred

mmccool:
Never mind my above comments about security, it seems I was looking at an old version somehow.  I did a real review and it looks good, *except* I think we should make JSON Pointer the official fragment identifier for both TD and TM (otherwise it inherits from JSON, which does not in fact have one, so then TD and TM would not have one)

mmccool (changes requested):

sebastiankb:
from today's TD call:
- @sebastiankb will include @mmccool input
- after getting approval from @mmccool I'm going to merge

index.html
sebastiankb: done

index.html
sebastiankb: done

mmccool (approved):

PR #1480, opened 2022-04-27
default values for read/writeOnly only in PropertyAffordance definitions 

### LINKED ISSUE #1417: Default values of `readOnly` and `writeOnly` for non-property affordances
Should `readOnly` and `writeOnly` have default values when they are used in any data schema or only in Property affordances? Currently we have the following:
image

This means that actually an action input also has `readOnly` and `writeOnly` when the TD is expanded to include defaults. I find this really strange and think that the defaults should be limited to only property affordances

sebastiankb:
fixes #1417

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1480.html" title="Last updated on May 4, 2022, 4:00 PM UTC (3e0a2c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1480/e302121...3e0a2c7.html" title="Last updated on May 4, 2022, 4:00 PM UTC (3e0a2c7)">Diff</a>


egekorkan (approved):

sebastiankb:
from today's TD call:
- not merge the PR yet
- we should check if there is some backwards compatibility problems of existing implementations such as from the last TestFest

sebastiankb:

PR #1484, opened 2022-04-29
chore: fix filename bug in assertion sync workflow

JKRhb:
I noticed that the assertion sync is currently failing due to an incorrect file name. I think this PR should fix the issue.

### DIFF
--- a/.github/workflows/assertion-sync.yml
+++ b/.github/workflows/assertion-sync.yml
@@ -27,7 +27,7 @@ jobs:
       # This is done so that it can be compared later on
         run: |
           mkdir .temp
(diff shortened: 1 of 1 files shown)


egekorkan:
I had tested this in my fork, weird that it is failing but the fix is correct

egekorkan (approved):

sebastiankb (approved):

sebastiankb:

JKRhb:
Hmm. In my fork, the Workflow is still failing, even with the file name change: https://github.com/JKRhb/wot-thing-description/runs/6292962829?check_suite_focus=true 

Should the file path actually be `testing/inputs/results-template.csv`?

Issue #1453, opened 2022-03-31
Consistent way of linking / integrating JSON Schemas into the document


danielpeintner:
- however we should think about to add the JSON Schema of the TM within the spec document in the next update

_Originally posted by @sebastiankb in https://github.com/w3c/wot-thing-description/issues/1430#issuecomment-1083279896_

danielpeintner:
Note: As mentioned in the TD call I think we should either
* link both JSON Schema files in the document
  OR
* include both files

Mixing does not make sense to me.

egekorkan:
I would prefer to link both of them in this case. My arguments are:
1. Smaller specification document
2. Sections after the schemas need less scrolling
3. People who read it are sent to the GitHub version which is:
  - more or equally up to date 
  - easier to copy-paste
  - its link is more obvious. We had some people asking where the schemas are
4. If we have another schema in the future, it might start getting more annoying to put more and more schemas.

PR #1487, opened 2022-05-02
JSON Schema fixes 

egekorkan:
In this PR I am doing the following:
- changing `iri-reference` to `uri-reference` in the json schema. fixes https://github.com/w3c/wot-thing-description/issues/956
- remove the json schema from the index.html . fixes https://github.com/w3c/wot-thing-description/issues/1453 . This will also avoid myriad of merge conflicts we get due to that.


index.template.html
```diff
@@ -6392,8 +6392,10 @@ <h3>Webhook Event Example</h3>
   <section id="json-schema-for-validation" class="appendix informative">
   <h1>JSON Schema for TD Instance Validation</h1>
```
JKRhb, line 6393: Should we also include a link to the TM schema here/in a separate section?

egekorkan: Might be a good idea. We have it already at https://w3c.github.io/wot-thing-description/#thing-model-concept but maybe a section somewhere that puts all these kind of informative resources would be nice. We now have JSON Schema based TD and TM validation, SHACL based validation and Typescript Types (at scripting API repo)

index.html
```diff
@@ -1209,16 +1209,16 @@ <h2>Core Vocabulary Definitions</h2>
                 names are used in a <code>security</code>
                 name-value pair.</td><td>mandatory</td><td><a>Map</a> of <a href="#securityscheme"><code>SecurityScheme</code></a></td></tr>
 <tr class="rfc2119-table-assertion" id="td-vocab-profile--Thing"><td><code>profile</code></td><td>Indicates the WoT Profile mechanisms followed by this Thing Description and the corresponding Thing implementation.</td><td>optional</td><td><a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#anyURI"><code>anyURI</code></a> or <a>Array</a> of <a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#anyURI"><code>anyURI</code></a></td></tr>
+<tr class="rfc2119-table-assertion" id="td-vocab-schemaDefinitions--Thing"><td><code>schemaDefinitions</code></td><td>Set of named data schemas.
```
mlagally, line 1212: unrelated change.

egekorkan: > unrelated change.

Yes but not my fault. This keeps happening due to auto generation script

sebastiankb:

Issue #1410, opened 2022-03-01
Interaction of RFC3986 and JSON Pointer


aphillips:
11.3.2 Extension and Import
https://w3c.github.io/wot-thing-description/#thing-model-extension-import

> The tm:ref value MUST follow the pattern
>
> *   file location as URI [[RFC3986](https://w3c.github.io/wot-thing-description/#bib-rfc3986)](Section 4.1)), followed by
> *    \# character, and followed by
> *    JSON Pointer [[RFC6901](https://w3c.github.io/wot-thing-description/#bib-rfc6901)] definition.

JSON Pointer allows basically any Unicode character (excepting the path separator). The `tm:ref` definition allows the pointer to appear effectively as a fragment identifier glued onto the file location's URI. URIs on the wire are not code point sequences, though. Should there be a mention of the need to encode as UTF-8 for the wire? I think the intention of the URI reference is more like what is specified by the [URL Spec](https://url.spec.whatwg.org/) rather than octets 

(All of the examples are ASCII, so it's non-obvious, and JSON doesn't need to percent-encode the values)

egekorkan:
Call of 02.03:
- We should discuss whether it is OK to have spaces instead of URL encoding (ege guesses that they should be URL encoded)

sebastiankb:
I'm not sure we have to do any action here. [TD must be encoded in UTF-8](https://w3c.github.io/wot-thing-description/#td-json-open_utf-8) and this should be also valid for the content of the `tm:ref`. Do I see that correctly?

sebastiankb:
from todays TD call, we would like also discuss this issue again with @aphillips.

sebastiankb:
from today's TD call:
- it seems to introduce a note that describes the URI situation and how it can encoded on the wire
- it should make clear that a double encoding should be avoided
- @sebastiankb will provide a PR; @aphillips will review

sebastiankb:
PR merged

PR #1500, opened 2022-05-09
avoid double encoding for tm:re

sebastiankb:
solves #1410

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1500.html" title="Last updated on May 18, 2022, 3:57 PM UTC (75fd278)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1500/78cfc50...75fd278.html" title="Last updated on May 18, 2022, 3:57 PM UTC (75fd278)">Diff</a>

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -244,31 +244,45 @@ a[href].internalDFN {
 </style>
 
 <script>
-    document.addEventListener('DOMContentLoaded', () => {
-        // let ReSpec do syntax highlighting first
-        document.respec.ready.then(() => {
-            document.querySelectorAll('.with-default').forEach(n => {
-                let wo = n.querySelector('pre:nth-of-type(1)');
-                let w = n.querySelector('pre:nth-of-type(2)');
-                let cbox = n.querySelector('.with-default-control input[type="checkbox"]');
+	function defaultCheckboxListener() {
+		document.querySelectorAll('.with-default').forEach(n => {
+			let wo = n.querySelector('pre:nth-of-type(1)');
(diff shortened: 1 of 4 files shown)


aphillips:
I see how you got here, but I'm not sure that the addition of the note will be that clear. Note that double-encoding would encode the percent sign (the string wouldn't include any non-ASCII characters--the escaping would process ASCII characters reserved in URL). Perhaps instead:

> Portions of the `tm:ref` value might contain non-ASCII characters that require URL ("percent") encoding before use. Before applying escapes to a `tm:ref` value, implementations should check that the value is not already encoded.

sebastiankb:
@aphillips thanks for your feedback. Text is updated based on you proposal.

index.template.html
```diff
@@ -5141,6 +5141,13 @@ <h3>Extension and Import</h3>
     </span>
     </p>
 
+    <p class="note">
+        Portions of the <code>tm:ref</code> value might contain non-ASCII characters that require URL ("percent") 
+        encoding before use. Before applying escapes to a <code>tm:ref</code> value, implementations should check that the value is not already
```
egekorkan, line 5146: I think we need an RFC assertion here while making sure that the assertion is understandable on its own (implies changes to the sentence)

egekorkan (changes requested):

sebastiankb:

sebastiankb:
from today's TD call, decided to merge.

@egekorkan also agree to keep it as a note

Issue #1488, opened 2022-05-03
at risk labels for CR


sebastiankb:
we need to highlight the places and features that are `at risk` based on the latest testing result

PR #1501, opened 2022-05-09
WIP: Add at-risk hints in the text

sebastiankb:
solves #1488

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -2112,7 +2112,9 @@ In summary, this requires the following:
     instance of <code>SecurityScheme</code>, MUST be serialized as a JSON object.
   </span></p>
 
-  <p><span class="rfc2119-assertion" id="td-security-schemes">
+  <p><span class="at-risk">The following feature is at
+    risk: </span>
+      <span class="rfc2119-assertion" id="td-security-schemes">
     All name-value pairs of an instance of one of the <a>Subclasses</a> of
     <code>SecurityScheme</code>,
     where the name is a <a>Vocabulary Term</a> included in the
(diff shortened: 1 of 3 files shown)


index.template.html
```diff
@@ -2112,7 +2112,9 @@ <h3><code>securityDefinitions</code> and <code>security</code></h3>
     instance of <code>SecurityScheme</code>, MUST be serialized as a JSON object.
   </span></p>
 
-  <p><span class="rfc2119-assertion" id="td-security-schemes">
+  <p><span class="at-risk">The following feature is at
+    risk: </span>
+      <span class="rfc2119-assertion" id="td-security-schemes">
```
JKRhb, line 2117: This feature has been a part of TD version 1.0 (see https://www.w3.org/TR/wot-thing-description/#security-serialization-json), can it still be marked as at risk?

(I have to admit that I do not fully understand what this assertion is actually mandating, though. Maybe it could be rephrased if it is worth keeping?)

sebastiankb: Thanks for the hint. I removed the at-risk label again.

sebastiankb:
from today's TD call:
- decided not to merge
- there is a css file that should be maintained atrisk.css
- @mmccool will check the script if the corresponding assertions are auto generated and will create a PR with the updated atrisk.css file

egekorkan:
@mmccool I see https://w3c.github.io/wot-thing-description/#combosecurityscheme and others highlighted properly and from the testfest I remember that you have automated everything. This PR can be closed without merging?

Issue #1502, opened 2022-05-10
Beautify infromation model drawings for CR


sebastiankb:
So far, the drawing for Figure 1-4 are auto generated. For CR we should create a more attractive drawing similar to [TD 1.0](https://www.w3.org/TR/wot-thing-description/#td-core-model).

PR #1504, opened 2022-05-17
Beautify visualizations

fatadel:
Fixes https://github.com/w3c/wot-thing-description/issues/1502


danielpeintner:
Note: I think we should tweak the script to no *override* the images after the next `npm render`.
A configuration flag would be nice.

danielpeintner:
Note2: I think based on https://github.com/w3c/wot-thing-description/pull/1491/ we need the `sync` keyword.

sebastiankb:
PR https://github.com/w3c/wot-thing-description/pull/1507 replace this one

PR #308, opened 2018-11-26
Added panasonic.html to testing/implementations

ToruKawaguchi:

PR #1505, opened 2022-05-18
Updated Implementation Report and Results Files - May 2022

mmccool:
This is not really "done" but I would suggest merging it as it updates the report data significantly.  Note however that we need to update the manual assertions, including for the older test inputs.

- Based on updated test inputs in [wot-testing PR #308](https://github.com/w3c/wot-testing/pull/308)
- Added TinyIoT, Logilab, and Intel-WoT-HA (home assistant) results
- Added ECHONET Bridge and UNIBO Farm from 09.2021 plugfest
- Added intel-proxy.csv (manual assertions only)
- Older input files from TD 1.0 testing, if they were not updated for TD 1.1, to check compatibility - Huawei, Panasonic, Smart Things, OCF, etc.
- Removed obsolete implementation descriptions (but... might need to retrieve them for TD 1.0 compatibility testing).


sebastiankb:

sebastiankb:

PR #1522, opened 2022-06-03
June 2022 Implementation Report - Update 1

mmccool:
Note: merging early so others can see recent changes.  Will create a new PR for further updates.

Draft implementation report for June 2022 Testfest.

Aim was to finalize update by end of testfest to prepare for CR transition, including updates to CSS files, at-risk markup, etc. 

Please see testing/output.log.  At the end there are a few assertions noted that do not have RFC-2119 keywords that need to be fixed; see issue #1523


sebastiankb:
it seems that the atrisk.css is missing

PR #1528, opened 2022-06-08
New assertions into assertion.csv

egekorkan:
1. I have updated the script to take extra-asserts.html into account
2. I have added new assertions to extra-asserts.html so that the new terms are taken into account

I will be merging it instantly but feel free to revert it. We need it asap for testfest


egekorkan:
Also some assertions were obsolete, so I have removed them from depends.csv and categories.csv

mmccool:
Beat me to it.  I updated the implementatoin report also (and merged it, so template.csv is also updated now).

Issue #1520, opened 2022-05-31
ReSpec Error - Reference "[ACE-OAuth]" not found.


danielpeintner:
we use the spec reference [[ACE-OAuth](https://w3c.github.io/wot-thing-description/#bib-ace-oauth)] which is not available in [specref](https://www.specref.org/?q=ACE-OAuth)
see for example https://w3c.github.io/wot-thing-description/#adding-security-schemes

![grafik](https://user-images.githubusercontent.com/11281461/171180783-53167ee1-7f9f-427a-88d1-b0f49c6cb563.png)

@mmccool what is the correct reference?

danielpeintner:
The problem is that if we publish CR and we have the following reference https://w3c.github.io/wot-thing-description/#bib-ace-oauth looking like this, it is not good.

![grafik](https://user-images.githubusercontent.com/11281461/171368174-d684d05d-ffbe-431e-9f2b-323eb633321a.png)

sebastiankb:
in my point of view, the should be only a replacement from [ACE-OAuth] to [RFC9200], right?

danielpeintner:
The problem right now is that RFC9200 is not known yet, see [specref](https://www.specref.org/?q=RFC9200).

egekorkan:
Call of 15.06:
- We have an editor's note about this:
![image](https://user-images.githubusercontent.com/20195407/173866286-ef6b63f1-c0a9-4159-a490-ef483dffdfd0.png)

- Specref is sometimes slow to adopt such new specs, we will add it to localBiblio for now.

mmccool:
Note the above PR also updates a second reference I found (where we give an example of how to use an extension).

mmccool:
RFC9200 is still in "Auth48" state as of today: https://www.rfc-editor.org/auth48/rfc9200

Once it is published and is added to ReSpec we will want to update this references.

sebastiankb:
[ACE-OAuth] is replaced by [RFC9200] in the spec: 

![image](https://user-images.githubusercontent.com/13832739/183378884-ed6487c5-059d-45fc-9a58-a8eb17d52166.png)

I will close this issue

PR #1544, opened 2022-06-16
Add RFC9200 reference

mmccool:
This is to replace the ACE-OAuth reference.  Note however that while the ACE OAuth spec has passed review, as of this moment RFC9200 has NOT been published: it is still in the queue.  Please see the status at https://www.rfc-editor.org/info/rfc9200.  I put in the "expected" URL in the reference; right now though it will not work, although if you google for it you can find a draft here: https://ftp.ripe.net/rfc/authors/rfc9200.pdf

As soon as this RFC is both published and in specref the local definition should be removed.  I expect this will happen prior to PR.

Let's leave the issue (#1520) open until this is fully resolved (e.g. we are using specref for this citation).

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1544.html" title="Last updated on Jun 16, 2022, 4:51 PM UTC (6c4cf86)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1544/1d10392...mmccool:6c4cf86.html" title="Last updated on Jun 16, 2022, 4:51 PM UTC (6c4cf86)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -176,6 +176,21 @@
                 , publisher: "W3C"
                 }
               }
+              , "RFC9200": {
+                  "href": "https://www.rfc-editor.org/rfc/rfc9200",
+                  "title": "Authentication and Authorization for Constrained Environments (ACE) Using the OAuth 2.0 Framework (ACE-OAuth)",
+                  "authors": [
+                    "L. Seitz",
+                    "G. Selander",
+                    "E. Wahlstroem", 
+                    "S. Erdtman",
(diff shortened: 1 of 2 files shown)


JKRhb (approved):
LGTM :) Thanks, @mmccool!

egekorkan:
We had agreed on this exact solution in the call, merging

PR #1549, opened 2022-06-22
Add husky pre-commit for rendering

relu91:
This PR adds a convenient git hook to automatically render the documentation and the TmSchema at every commit. The idea is to reduce human errors when contributing to the spec. The script is an improved version of what we already have in https://github.com/w3c/wot-binding-templates/tree/main/.husky. It works also with a dirty stage, which means that you can partially stage the changes and the hook will generate the documentation accordingly. 

**Note**: to correctly initialize the hook you have to call `npm install` at least once in the root folder of the repository.


sebastiankb (approved):

mmccool:
TD call June 22: we agreed that this can be marked as non-substantive to resolve the IPR issue.  After that, some more reviewers (Ege? Daniel?) would probably be a good idea.

sebastiankb:
from today's TD call: since Ege still on vacation lets postpone the decision about this PR to next week

danielpeintner:
@relu91 this PR solves https://github.com/w3c/wot-thing-description/issues/883 right?

egekorkan (approved):

sebastiankb:

PR #1552, opened 2022-06-23
Add informative text and an example for relative tm:ref

### LINKED ISSUE #1546: Explicitly allow/mention relative `tm:ref` imports?
The TD specification currently only gives examples for `tm:ref` that import definitions from other documents. However, as the empty string for the URI part indicates a same-document reference (if I am not mistaken), I think it would be nice if the specification explicitly mentioned that relative imports are also legal values and gave an example for such an import. Otherwise, you could get the impression from the current examples that you can use `tm:ref` for imports from external documents only.

An example for a relative `tm:ref` import could look something like this:

```json
{
  "@context": ["https://www.w3.org/2022/wot/td/v1.1"], 
  "@type" : "tm:ThingModel",
  "title": "Multi Sensor Model",
  "properties": {
    "genericTemperature": {
      "type": "number",
      "unit": "C"
    },
    "innerTemperature": {
      "tm:ref": "#/properties/genericTemperature",
      "title": "The inner temperature",
      "minimum": 10
    },
    "outerTemperature": {
      "tm:ref": "#/properties/genericTemperature",
      "title": "The outer temperature",
      "description": "The outer temperature is measured in Kelvin",
      "unit": "K"
    }
  },
  "tm:required": [
    "#/properties/innerTemperature",
    "#/properties/outerTemperature"
  ]
}
```

JKRhb:
As discussed in #1546, this PR adds a few informative sentences regarding relative `tm:ref` imports, as well as an additional example (which could probably be simplified).

Resolves #1546

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-thing-description/pull/1552.html" title="Last updated on Jun 23, 2022, 8:50 AM UTC (32011f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1552/780da07...JKRhb:32011f0.html" title="Last updated on Jun 23, 2022, 8:50 AM UTC (32011f0)">Diff</a>


index.template.html
```diff
@@ -4723,7 +4725,42 @@ <h3>Extension and Import</h3>
...
+                "tm:ref": "#/properties/genericTemperature",
+                "title": "The inner temperature",
+                "minimum": 10
+            },
+            "outerTemperature": {
+                "tm:ref": "#/properties/genericTemperature",
+                "title": "The outer temperature",
+                "description": "The outer temperature is measured in Kelvin",
+                "unit": "K"
+            }
```
thjaeckle, line 4756: The JSON intendation of the properties should be one level deeper here, improving readbility.

index.template.html
```diff
@@ -4723,7 +4725,42 @@ <h3>Extension and Import</h3>
...
+            },
+            "tm:required": [
+            "#/properties/innerTemperature",
+            "#/properties/outerTemperature"
```
thjaeckle, line 4760: Here JSON intendation would also improve readability.

thjaeckle (commented):
The JSON intendation of the example could be improved.
Apart from that: Looks good to me.

index.template.html
JKRhb: Thank you! Should be fixed now :)

sebastiankb (approved):

sebastiankb:
Thanks, looks good to me too. There could be more text about the example, however, in my view it's simple and kind of self-explanatory.

Issue #1548, opened 2022-06-21
Oracle review comments on CR draft - Part 1


mlagally:
# Oracle review comments (Part 1)

## RFC2119 assertions

There are multiple requriements in normative sections which contain an optional clause (should, may, recommended) in lower case. Several of them are important assertions and should be marked up with RFC2119 markup.

Examples include the language around TD processors and TD producers in *5.3.1.1 Thing*, content of some tables in chapter *5.3*, e.g. *5.3.3.9 OAuth2SecurityScheme* and several clauses in *8. Behavioral Assertions*.

## Hints
In several sections, the text talks about hints, e.g. *5.3.4.1 Link*: 

type: "Target attribute providing a hint indicating what the media type [RFC2046] of the result of dereferencing the link should be."

To achieve interoperability this is not sufficient. it should be an assertion that only media type values defined by IANA *MUST* be used.
 

### 5.3 Class Definitions
The section should make it a mandatory requirement that vocabulary **terms** MUST be case sensitive.

#### 5.3.1.1 Thing
- security: "Set of security definition names, chosen from those defined in securityDefinitions. These must all be satisfied for access to resources." What does "all be satisfied" mean? Bettwe reword to "security mechanisms supported by the thing"

##### 5.3.1.2 InteractionAffordance
- uriVariables: ObjectSchema and ArraySchema are forbidden, however there's an assertion required.
- forms: forms are mandatory, need to clarify if 0-element forms are permitted.

##### 5.3.1.3 PropertyAffordance
- invalid combinations of op values: some combinations don't make sense and should be prohibited. Examples include observe+unobserve.
- observable: RFC2119 *SHOULD* assertion

##### 5.3.1.4 ActionAffordance
- combinations of operations do not make sense and should not be permitted.

- An operation *cancelAllActions* would make sense as an emergency stop feature.

##### 5.3.1.5 EventAffordance
- data: is currently optional, should be mandatory since an event subscription without a data flow does not make sense.
- combinations of operations do not make sense and should not be permitted.

##### 5.3.1.6 VersionInfo

- "It is recommended that the values within instances and model of the VersionInfo Class follow the semantic versioning pattern ..." should be an RFC2119 assertion.

#### 5.3.2 Data Schema Vocabulary Definitions

- The following should be an RFC2119 assertion: "it is noted that data schema definitions within Thing Description instances are not limited to this defined subset and MAY use additional terms found in JSON Schema". 
[...]

PR #1558, opened 2022-06-29
addresses CR 1.1 candidate review from Oracle  

sebastiankb:
see #1548

- [ ] RFC2119 assertions
- [ ] hint terms
- [x] vocabulary terms and values MUST be case sensitive

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1558.html" title="Last updated on Aug 17, 2022, 3:47 PM UTC (3e2ba00)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1558/103f786...3e2ba00.html" title="Last updated on Aug 17, 2022, 3:47 PM UTC (3e2ba00)">Diff</a>


index.html
```diff
@@ -1168,7 +1168,10 @@ <h2>Class Definitions</h2>
         </p>
 
         <p>
-            Please note that all vocabulary terms and values are case sensitive.
+            <span class="rfc2119-assertion" id="td-processor">
+                Please note that all vocabulary terms and values MUST be case sensitive.
+            </span>
+            
             This is also true for the serialization of the information model (Section <a href="#sec-td-serialization"></a>).
```
mlagally, line 1175: This looks like another normative requirement. Suggest to either include it in the <span> section above or define another normative <span>

index.html
```diff
@@ -7013,7 +7016,7 @@ <h5>Script Injection</h5>
 	      when generating HTML: with care. 
 	      <span class="rfc2119-assertion" id="sec-inj-sanitize">
 		      Strings sourced from TDs MUST either be sanitized using a carefully vetted HTML
```
mlagally, line 7018: There was a problem with the markup tooling that prevents having two assertions in the same <span>. Please check with @mmccool  on how to handle that.

index.template.html
```diff
@@ -1168,7 +1168,10 @@ <h2>Class Definitions</h2>
         </p>
 
         <p>
-            Please note that all vocabulary terms and values are case sensitive.
+            <span class="rfc2119-assertion" id="td-processor">
+                Please note that all vocabulary terms and values MUST be case sensitive.
```
mlagally, line 1172: Does this really hold for values?

Are 10m different from 10M ?

index.template.html
```diff
@@ -1168,7 +1168,10 @@ <h2>Class Definitions</h2>
...
+            </span>
+            
             This is also true for the serialization of the information model (Section <a href="#sec-td-serialization"></a>).
         </p>
```
mlagally, line 1176: see comment above

index.template.html
danielpeintner: @mlagally I think the point you are raising goes beyond what the TD can offer. You are asking for a *real* canonicalization process.
I guess no-one would argue that besides case-sensitive charterers we would also need to match the number "1" with "1.0" etc. Hence, also values need to match "exactly"...

egekorkan:
I will continue working on this PR

egekorkan:
@mlagally , I have addressed your comments that I think are valid and need fixing. There are some that I do not agree with, which I have pointed out in the related issue. Since you have found some normatively important fixes, this PR has normative changes. I think none are very important for a discussion.

index.template.html
sebastiankb: I think the case sensitive requirement is different to the canonicalization one.  Canonicalization will be a topic for 2.0.

index.html
egekorkan: This comment seems to be outdated. This is now a normative requirement. Resolving

egekorkan:
@sebastiankb I am going to resolve the merge conflicts and merge the main here so that we can see the new changes that happened with some assertions

egekorkan:
@sebastiankb I think that this PR is ready. Let's remove WIP label, merge it and then continue work on separate PRs

sebastiankb:

PR #1562, opened 2022-07-06
remove metadata from example ids

### LINKED ISSUE #1216: URLs and IDs should not have embedded metadata

egekorkan:
fixes #1216

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1562.html" title="Last updated on Jul 6, 2022, 12:35 PM UTC (330365a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1562/52747aa...330365a.html" title="Last updated on Jul 6, 2022, 12:35 PM UTC (330365a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -500,7 +500,7 @@ a[href].internalDFN {
 <aside class="example" id="simple-thing-description-sample" title="Thing Description sample">
     <pre>{
     "@context": "https://www.w3.org/2022/wot/td/v1.1",
-    "id": "urn:dev:ops:32473-WoTLamp-1234",
+    "id": "urn:uuid:0804d572-cce8-422a-bb7c-4412fcd56f06",
     "title": "MyLampThing",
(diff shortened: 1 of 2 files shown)


egekorkan:
I have generated them with https://www.uuidgenerator.net/version4 and also put different ids to each example

PR #1564, opened 2022-07-06
explain contentType usage

### LINKED ISSUE #1531: Explaining contentType usage
I have realized that we are not explaining what messages are described by `contentType`. It can be understood if details are read all over the place, but it would be better to list the following possibilities somewhere:

- Request Response are the same: `contentType`  in the form applies to both
- Request Response different: `contentType`  in the form applies to request, for the Response one needs `response` term
- Request not present: `contentType`  in the form applies only to response (i.e. there is only `output` in an action)
- Response not present: `contentType`  in the form applies only request (i.e. there is only `input` in an action)

Note: Request is what Consumer sends to Thing, Response is what Thing sends to Consumer

egekorkan:
fixes #1531

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1564.html" title="Last updated on Nov 2, 2022, 12:36 PM UTC (f1cb64b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1564/48caa9d...f1cb64b.html" title="Last updated on Nov 2, 2022, 12:36 PM UTC (f1cb64b)">Diff</a>


egekorkan:
Also, I am not super happy about how this looks like. Maybe a matrix might look better. Any opinions?

danielpeintner:
> Also, I am not super happy about how this looks like. Maybe a matrix might look better. Any opinions?

Maybe some kind of a _tree_  ? Not sure either.
Text/tables are definitely easier to maintain than figures...  so maybe it is fine as it is

egekorkan:
That is also true. By the way, what I meant with the matrix is something like the following. The cells can be the actual text or links to some paragraphs that come later.image

egekorkan:
Call of 13.07: I will try the second matrix format as well and we can decide on one next time

egekorkan:
So I have started a table as an alternative but it is not possible to represent all the cases mentioned above, such as 1 input, 1 output, different contentType and same contentType . Even though that listing the cases is not elegant, it gets the job done :)

egekorkan:
I had a separate chat with @danielpeintner and he recommended a decision tree. I did one hand drawing on this direction, definitely better than the matrix idea

image

danielpeintner:
I agree, it is better but still kind of complex 🙈

egekorkan:
I think we can wait if people have other ideas about this. I think this can be done after CR since it is simply explaining something we already have

egekorkan:
Below are the three options so far:
- Alternative1: Simple Table with Text
image
- Alternative2: Matrix. This is not complete
image
- Alternative3: Decision Tree. This is just a hand drawing. I can provide an SVG drawn in inkscape if we agree on this
image

sebastiankb:
from today's TD call: @zolkis maybe you also a good idea of a representation of this combination

egekorkan:
@ektrah I have done the single contentType use case, please have a look

ektrah:
Looks quite good so far. Two comments:

* How do I determine which case applies? (Could https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1564.html#table-operation-types be extended with columns indicating if input/output is present/not present in each operation?)
* I think the content types specified in headers (indicating the payload format or the preference for a payload format) should always take precedence over the content type in the form.

egekorkan:
Thanks for the feedback (also sorry for taking too long to do this)!

I did not understand the first point but we can discuss it in the meeting. 

Regarding the other one, they should always match (There is an assertion that says Every form in a WoT Thing Description MUST accurately describe requests (including request headers, if present) accepted by the [Thing](https://w3c.github.io/wot-thing-description/#dfn-thing) in an interaction.). 
In case they do not, the header should be definitely used for decoding. If we encourage this without the previous clause, the `contentType` in the form for responses with a protocol that has header support becomes irrelevant. However, even in HTTP it is not mandatory to have this information. So I guess that we recommend that the Consumer should look into the header first, if it cannot find one using the `contentType` of the form makes sense (which is what you said).

egekorkan:
Call of 26.10:
- The changes would help implementers but this is a rather big (informative) change. So we should not merge it during this stage.
- We can revisit it in a couple weeks and decide to where to put it. We can put it in the appendix as a way for developers to look up to it.

egekorkan:
By the way, this can be useful for Scripting API as well at https://w3c.github.io/wot-scripting-api/#dfn-parse-interaction-response or in each operation of the consumed and exposed interface

egekorkan:
I am done with this PR. However, there is a small clause about the examples in the paragraph before the tables. I propose to add these examples in a separate PR, this is already getting too big

egekorkan:
Call of 02.11:
- We should get more feedback from developers since this is a complicated case.
- The group is OK to merge. @relu91 will provide a review separately, maybe in the upcoming examples PR
- It is good to have it in the appendix.

Issue #1064, opened 2021-03-03
Need Fragment Identifier IANA registration for td-json


mmccool:
Apparently, technically, to use JSON Pointer to refer to part of a TD, the fragment identifier needs to be registered for the media type.  There is in fact no fragment identifier rule defined for application/json and https://www.ietf.org/rfc/rfc6839.txt explicitly says that for derived types like application/td+json, the rule is inherited from application/json (so, if we don’t specify a rule, there isn’t one).

The “body” security scheme is used internally in a TD and not with a URL so doesn’t need this.  However it would be handy for links from a TD to parts of another TD, as in the proposed geolocation information model.

Of course we should do the same thing for TMs.

mmccool:
We also need to clean up definitions for relative URLs and partial URLs (eg only fragment identifier part in URL that refers to current document)

sebastiankb:
I am not 100% sure I understand your request. Is there a problem that JSON pointer processors are not able to parse TDs because the media type is not known?

mmccool:
@sebastian, formally, to use a fragment identifier scheme in a URL this has to be defined in the IANA content type, or in a parent type.  However, neither application/json nor application/jsonld define one.  This is not a big deal, it's just a technicality.

PR #1571, opened 2022-07-08
fix tm:required example

### LINKED ISSUE #1563: Wrong format for tm:required in examples
The TD spec mentions JSON Pointer as the type of `tm:required` . However, the examples contain:
```json
{
"tm:required": [
        "#/properties/status",
        "#/actions/toggle"
    ]
}
```

Use of `#` is actually not a JSON Pointer as per [RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901):
image

egekorkan:
fixes #1563

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1571.html" title="Last updated on Jul 12, 2022, 1:27 PM UTC (9885a1d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1571/d4daa90...9885a1d.html" title="Last updated on Jul 12, 2022, 1:27 PM UTC (9885a1d)">Diff</a>


danielpeintner (changes requested):
I think there are some more left

https://github.com/w3c/wot-thing-description/blob/3ccbb14aa818bcfae7e53949786b26102b2a61eb/index.html#L6028-L6031

egekorkan:
@danielpeintner I have provided fixes for your comment

danielpeintner (approved):

JKRhb:
> I mean, even the RFC is giving some warnings that it is not necessarily the case. I think that we cannot just say JSON Pointer since it can have multiple representations.

That is a good point. Would this imply that we should not define a fragment identifier for `application/td+json` due to incompatibilities (resolving #1064)? Or is this question independent of this issue?

Another aspect: Removing the `#` for JSON pointers would imply that the relative `tm:ref` introduced in #1552 need to be changed as well, right? This would probably require changes to the respective assertion (in a follow-up PR).

PR #1587, opened 2022-07-14
Add hreflang to schema

egekorkan:
Tests see: https://www.jsonschemavalidator.net/s/qwTU2Kxb

to remove fail, go to the last line and change the hreflang value to `de` or any other bcp47 tag

We should add this to descriptions and titles but it might be breaking some implementations that were looking at the schema but not reading the spec properly

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -152,6 +152,10 @@
         }
       ]
     },
+    "bcp47_string":{
+      "type":"string",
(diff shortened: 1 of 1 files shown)


egekorkan:
I had computer problems and did everything via browser, this change should be copied over to tm schema

validation/td-json-schema-validation.json
```diff
@@ -152,6 +152,10 @@
         }
       ]
     },
+    "bcp47_string":{
+      "type":"string",
+      "pattern": "^(((([A-Za-z]{2,3}(-([A-Za-z]{3}(-[A-Za-z]{3}){0,2}))?)|[A-Za-z]{4}|[A-Za-z]{5,8})(-([A-Za-z]{4}))?(-([A-Za-z]{2}|[0-9]{3}))?(-([A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3}))*(-([0-9A-WY-Za-wy-z](-[A-Za-z0-9]{2,8})+))*(-(x(-[A-Za-z0-9]{1,8})+))?)|(x(-[A-Za-z0-9]{1,8})+)|((en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang)))$"
```
sebastiankb, line 157: did you take this definition from an official specification

sebastiankb (approved):
PR looks ok. there is just a comment about the origin of the pattern.

validation/td-json-schema-validation.json
egekorkan: I took it from https://github.com/project-open-data/project-open-data.github.io/blob/master/schema/1_0_final/single_entry.json#L223 and validated it against some picks from https://tools.ietf.org/search/bcp47#appendix-A

I could not find a copy pastable pattern at the IETF spec

sebastiankb:

PR #1636, opened 2022-07-29
Update manual.csv

egekorkan:
Includes newest assertions of playground. Merging quickly to sync


egekorkan:
The diff is this large since it adds a new column but it is actually just removing rows

egekorkan:
@ashimura @mmccool I also have IPR problems now. Not sure what is going on

egekorkan:
But I can mark it like that since it is nothing normative

Issue #1594, opened 2022-07-15
Using `tm:required` for non-affordance members


egekorkan:
Currently, the use of `tm:required` is limited to affordances with the following assertion in the spec:

> The JSON Pointers of tm:required MUST resolve to an entire interaction affordance Map definition. 

This means that it is not possible to indicate that some members like top level forms, version, securityDefinitions etc. cannot be made mandatory. Should we relax this assertion?

sebastiankb:
from today's TD call:
- default assumptions of TM, everything is mandatory and should be followed by TD instances
- it turns out, that we would more need a TD "tm:optional" then "tm:required" (this would also solve the issue #1581)
- we should ask developer about this idea (e.g., @thjaeckle, @mjkoster )

thjaeckle:
As I already suggested in #1400 I absolutely support that as it feels more "natural" for me as developer to explicitly declare optional affordances in the "interface" (as I understand the ThingModel).

egekorkan:
That is good news (in a way). @thjaeckle do you think you can provide implementation examples until next week's testfest?

thjaeckle:
@egekorkan "implementation" as in TM examples with a suggested "tm:optional" section?

egekorkan:
Yes :)

thjaeckle:
Then sure - I can prepare something.. just point me to the place where to put those ;)

egekorkan:
I have spoken with two implementors of TM features, namely @thjaeckle (contributor to Ditto) and @fexpal (contributor to Editdor) and they would be ok with this change. For them, most (or nearly all) use cases use TMs like an interface where every affordance is implemented anyways. 

TMs with `tm:optional` should go to https://github.com/w3c/wot-testing/tree/main/data/input_2022/TD under the corresponding implementation

thjaeckle:
There is one problem I just noticed with `tm:optional`:
When using with TM extension, a TM B (e.g. being a "lamp" TM) extending a TM A (being a "switchable" TM with an `tm:optional` action `"switch-on-for-duration"`) can't define that for the TM B this action is mandatory and not optional

So to make the example clear:

TM A ("switchable") defines that action "switch-on-for-duration" is optional:
```json
{
  "@context": [
    "https://www.w3.org/2022/wot/td/v1.1",
    {
      "time": "http://www.w3.org/2006/time#"
    }
  ],
  "@type": "tm:ThingModel",
  "title": "Switchable",
  "version": {
[...]

PR #1640, opened 2022-08-02
init tm:optional

sebastiankb:
solve #1594

PR that is presented in tomorrows TD call. 

It is used to replace `tm:required`. This feature will result to an "at risk" declaration in the CR phase.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1640.html" title="Last updated on Aug 3, 2022, 7:47 AM UTC (725dbfe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1640/29f1a28...725dbfe.html" title="Last updated on Aug 3, 2022, 7:47 AM UTC (725dbfe)">Diff</a>


sebastiankb:
from today's TD call:
- there are pros and cons (see minutes)
- we should explain well why this changes was applied (especially in the change log)
- decided to merge
- change needs also be adopted by the Discovery TF (also  see https://github.com/w3c/wot-discovery/issues/388)

PR #1642, opened 2022-08-03
SHACLE Type fixes

sebastiankb:
it seems there are some mistakes defined in the validation file coming from PR #1574. This result to strange rendering behavoir in PR #1638 and #1639 (many places have "any type" defined instead of the specific datatype).

This PR should fix this issue and the render script should work correct again.

Issue #1438, opened 2022-03-16
ComboSecurityScheme: oneOf and allOf can be strings according to vocab table


FadySalama:
The vocab table under [ComboSecurityScheme](https://w3c.github.io/wot-thing-description/#combosecurityscheme) states that `"oneOf"` and `"allOf"` can be either of type `string` or `Array of string` but their descriptions state specifically that both should be arrays containing at least 2 strings. This might cause confusion

egekorkan:
This specific one should say that it must be an array

egekorkan:
The td-validation.ttl is correct (at least we think so) since it has `minCount 2` at https://github.com/w3c/wot-thing-description/blob/5080f00af839c6dae1c567b6a6f8a4599db4b248/validation/td-validation.ttl#L2171.

We need to understand how `oneOf` and `enum` in DataSchema is done, since they have the correct type:
![image](https://user-images.githubusercontent.com/20195407/173872084-d219be5a-96d9-4bc6-94a9-e30fbe9c88ec.png)

mmccool:
Agree with closing issue if the problem has been resolved.

egekorkan:
Can be closed as indicated.

PR #1645, opened 2022-08-03
fix type of `additionalResponses` and ComboSec `oneOf` and `allOf`

### LINKED ISSUE #1624: Type of additionalResponses
I am not sure if this is a spec generation bug but the `additionalResponses` seems to be `AdditionalExpectedResponse or Array of AdditionalExpectedResponse` . However, the JSON Schema only accepts arrays and the name also implies that. There is also the following  text in 5.3.4.2.2 (that is added my me):
> In this case the response name-value pair is still used for the primary response but additionalResponses may also be provided, whose value is an array of AdditionalExpectedResponse objects.

And another text that is an assertion:
> If the content type of an additional expected response differs from the content type of the form, the Form instance MUST include an entry in the array associated with the name additionalResponses that includes a value for the name contentType.

relu91:
As the title says this PR fixes the type of terms `additionalResponses`, `oneOf` and `allOf`. Now they are correctly rendered as `Array of ...` instead of `... and Array of ...`. 

*Notice* that this PR should be compatible with #1642 since I manually discarded all the artifacts created by errors in the SHACL file. Therefore we should be able to stack the twos without any conflicts. 

Fixes #1624 and #1438

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-thing-description/pull/1645.html" title="Last updated on Aug 3, 2022, 9:46 AM UTC (0e5a6b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1645/29f1a28...relu91:0e5a6b1.html" title="Last updated on Aug 3, 2022, 9:46 AM UTC (0e5a6b1)">Diff</a>


sebastiankb:

Issue #1556, opened 2022-06-29
Flexible character for instanceName in Thing Model definition 


sebastiankb:
The [tm-compose-name-collision assertation](https://w3c.github.io/wot-thing-description/#tm-compose-name-collision) predefines the character `'_'` to identify affordances from different TM definitions and to avoid name collisions. I think, TD should be more flexible and should allow other character then `'_'`.

A simple solution is that there is no assertion anymore that forces to use  `'_'`. Instead we can point out in the TD spec that the usage of `instanceName` would also be enough and can also introduce the `'_'` as last character or an alternative character such as `'.'` . Example:

```
{
...
    "links": [
      {
        "rel": "tm:submodel",
        "href": "./Ventilation.tm.jsonld",
        "type": "application/tm+json",
        "instanceName": "ventilation_"
      },
      {
        "rel": "tm:submodel",
        "href": "./LED.tm.jsonld",
        "type": "application/tm+json",
        "instanceName": "led_"
      }
    ],
...
  }
```

Or

```
{
...
    "links": [
      {
        "rel": "tm:submodel",
        "href": "./Ventilation.tm.jsonld",
        "type": "application/tm+json",
        "instanceName": "ventilation."
      },
      {
        "rel": "tm:submodel",
        "href": "./LED.tm.jsonld",
        "type": "application/tm+json",
        "instanceName": "led."
      }
    ],
...
  }
```

This is a small change and would help to be more flexible. What do you think?

sebastiankb:
from today's TD call:
- no decision yet
- we like to collect more views on this topic and get more people that have this implementation experiences

sebastiankb:
> Does it need to be specified at all?

Good point. I would be ok to redefine the assertion to

"Thereby the generation process SHOULD avoid possible name collisions. " 

Followed with a recommendation statement:

"To achieve this, the following strategy is recommended: All sub/child interaction names begin with the instanceName value followed with a predefined character(s) (e.g., '_', '.', '-', etc.)  and ends with the interaction name of the sub/child [Thing Model](https://w3c.github.io/wot-thing-description/#dfn-thing-model)."

What do you think? If ok, I can prepare a PR for Wednesday.

thjaeckle:
@sebastiankb yes, I think that is a good suggestion 👍

PR #1639, opened 2022-08-01
Adjust policy-like security assertions

mmccool:
Resolves #1635 

- See discussion in that issue for how these were resolved.  
- The net effect will be to remove three "policy-like" assertions that were difficult/impossible to test in implementations, while still keeping strongly worded (but non-normative) text in the specification.  
- The fourth assertion (on secure update) was more of a pure testing problem, but folding it into an earlier assertion resolves the issue without making it less normative.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1639.html" title="Last updated on Aug 3, 2022, 2:49 PM UTC (180b927)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1639/e3f5271...mmccool:180b927.html" title="Last updated on Aug 3, 2022, 2:49 PM UTC (180b927)">Diff</a>

PR #1649, opened 2022-08-09
Allow flexible chartacter for self-contained TDs from TMs

sebastiankb:
solve #1556

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1649.html" title="Last updated on Aug 17, 2022, 8:43 AM UTC (93c5565)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1649/103f786...93c5565.html" title="Last updated on Aug 17, 2022, 8:43 AM UTC (93c5565)">Diff</a>

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -5190,10 +5190,24 @@ provides <code>dimmable</code> and <code>RGB</code> capabilities.
     A single TD can also be generated which contains the interaction definitions of the top level/parent
     <a>Thing Model</a> and all interaction definitions of all sub/child <a>Thing Models</a>. 
     <span class="rfc2119-assertion" id="tm-compose-name-collision">
-       To avoid name collisions of the sub/child interaction names SHOULD rename the JSON name to the <code>instanceName</code> followed with <code>'_'</code> and the interaction name of the sub/child <a>Thing Model</a>.
+        Thereby the generation process MUST avoid possible name collisions. 
     </span> 
-    The following example shows a generated (self-contained) <a>Thing Description</a> of the Smart Ventilator Thing Model.
+    <span class="rfc2119-assertion" id="tm-compose-name-collision-seperation">
+        To achieve this, the following strategy is RECOMMENDED: 
+        All sub/child interaction names 
+        <ul>
+            <li>begin with the corresponding <code>instanceName</code> value,</li>
+            <li>followed with a predefined character(s) (e.g., '_', '.', '-', etc.), and</li>
+            <li>ends with the interaction name of the sub/child <a>Thing Model</a>.</li>
+        </ul>
+    </span> 
+
+    E.g., the property <code>switch</code> from the TM with the instance name <code>ventilation</code> will be 
+    concatenated to <code>ventilation_switch</code> when '_' character is used as separator. 
 </p>
+
+<p>The following example shows a generated (self-contained) <a>Thing Description</a> of the 
+    Smart Ventilator Thing Model. </p>
 
 <pre class="example" title="Self-contained TD of the Smart Ventilator TM">
     {
--- a/index.template.html
+++ b/index.template.html
@@ -5192,21 +5192,7 @@ provides <code>dimmable</code> and <code>RGB</code> capabilities.
     <span class="rfc2119-assertion" id="tm-compose-name-collision">
         Thereby the generation process MUST avoid possible name collisions. 
     </span> 
-    <span class="rfc2119-assertion" id="tm-compose-name-collision-seperation">
-        To achieve this, the following strategy is RECOMMENDED: 
-        All sub/child interaction names 
-        <ul>
-            <li>begin with the corresponding <code>instanceName</code> value,</li>
-            <li>followed with a predefined character (e.g., '_', '.', '-', etc.), and</li>
-            <li>ends with the interaction name of the sub/child <a>Thing Model</a>.</li>
-        </ul>
-    </span> 
-
-    E.g., the property <code>switch</code> from the TM with the instance name <code>ventilation</code> will be 
-    concatenated to <code>ventilation_switch</code> when '_' character is used as separator. 
-</p>
-
-<p>The following example shows a generated (self-contained) <a>Thing Description</a> of the 
+    The following example shows a potential generated (self-contained) <a>Thing Description</a> of the 
     Smart Ventilator Thing Model. </p>
 
 <pre class="example" title="Self-contained TD of the Smart Ventilator TM">
@@ -5241,7 +5227,7 @@ provides <code>dimmable</code> and <code>RGB</code> capabilities.
                     }
                 ]
             },
-            "ventilation_switch": {
+            "switch": {
                 "type": "boolean",
                 "description": "True=On; False=Off",
                 "forms": [
@@ -5250,7 +5236,7 @@ provides <code>dimmable</code> and <code>RGB</code> capabilities.
                     }
                 ]
             },
-            "ventilation_adjustRpm": {
+            "adjustRpm": {
                 "type": "number",
                 "minimum": 200,
                 "maximum": 1200,
@@ -5260,7 +5246,7 @@ provides <code>dimmable</code> and <code>RGB</code> capabilities.
                     }
                 ]
             },
-            "led_R": {
+            "R": {
                 "type": "number",
                 "description": "Red color",
                 "forms": [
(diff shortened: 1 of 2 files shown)


sebastiankb:
this PR provides also the rendered version of PR #1639

index.template.html
```diff
@@ -5190,11 +5190,25 @@ <h3>Composition</h3>
...
-       To avoid name collisions of the sub/child interaction names SHOULD rename the JSON name to the <code>instanceName</code> followed with <code>'_'</code> and the interaction name of the sub/child <a>Thing Model</a>.
+        Thereby the generation process MUST avoid possible name collisions. 
     </span> 
-    The following example shows a generated (self-contained) <a>Thing Description</a> of the Smart Ventilator Thing Model.
+    <span class="rfc2119-assertion" id="tm-compose-name-collision-seperation">
+        To achieve this, the following strategy is RECOMMENDED: 
+        All sub/child interaction names 
+        <ul>
+            <li>begin with the corresponding <code>instanceName</code> value,</li>
+            <li>followed with a predefined character(s) (e.g., '_', '.', '-', etc.), and</li>
```
thjaeckle, line 5200: I would write "with a predefined character" (without plural) - as this is only a recommendation, you can keep it simple and recommend to use only one character.

thjaeckle (approved):
Looks good to me apart from the (optional) "plural s" - which could get omitted IMO.

index.template.html
sebastiankb: done

thjaeckle:
> * there is the concern that a consumer may not identify anymore that it does not know where the property is coming from

Is this relevant to know as TD consumer?
I don't think that it should be as the consumer of the TD does not even need to know that on TM modeling level composition (via submodels) was used.
That's IMO the main benefit of the inlining - no links must be followed, the TD contains all of the affordances.

And if it must be determinable for the consumer of the TD which affordance is from a referenced submodel and which one not, this must IMHO be made e.g. an annotation inside the affordance and not just a "pattern" of an affordance name.

I think is is sufficient to just formulate that name collisions MUST be prevented.

What would be the difference between a recommendation for the naming strategy and using SHOULD?

Would this be the way then?
> The `instanceName` of a submodel SHOULD be used as prefix. 
> A delimiter SHOULD be used to separate `instanceName` and affordance name?

sebastiankb:
If the generated TD does not contain information about which TM it is based on (=no corresponding link entry provided), then it is perfectly fine for the consumers not to know where the interaction affordances come from, and they should accept affordances as they are offered in the TD.

However, if there is the knowledge shared, that TD is based on a TM (=link entry is provided) that uses the composition feature it would be useful for the consumer to understand which affordances is coming from, especially if there identical affordances names used with identical content.

E.g., if TM A and TM B are composed and they include an identical 'status' property definition and the generated TD uses own way of renaming, e.g.,  status1 and status2 then it is hard to understand by the consumer which status property it is belong to (TM A or B?).  

Maybe we can also distinguish between if the TM reference is provided in the TD or not. If provided then it SHOULD follow the guideline, otherwise is up to you.

> What would be the difference between a recommendation for the naming strategy and using SHOULD?

SHOULD is weaker than MUST. A definition can be found here https://www.rfc-editor.org/rfc/rfc2119

>Would this be the way then?

>    The instanceName of a submodel SHOULD be used as prefix.
>    A delimiter SHOULD be used to separate instanceName and affordance name?

Sounds also ok for me.

thjaeckle:
> However, if there is the knowledge shared, that TD is based on a TM (=link entry is provided) that uses the composition feature it would be useful for the consumer to understand which affordances is coming from, especially if there identical affordances names used with identical content.

Ok, then I understand where this comes from.
However - then you must also define that e.g. no affordance name in the TM which links to "submodels" is allowed to start with the `instanceName` of any referenced submodel - even if no naming conflict occurs, this then will lead to a TD where you still can't say where the affordances are coming from.

Example TM:
```json
{
...
  "@type": "tm:ThingModel",
  "links": [{
    "rel": "tm:submodel",
    "href": "./Lamp.tm.jsonld",
    "type": "application/tm+json",
    "instanceName": "fancy"
  }],
  "properties": {
    "fancy_something": {
       ...
    }
  }
}
```

This will lead to e.g. a TD:
```json
{
...
  "links": [{
    "rel": "type",
    "href": "...",
    "type": "application/tm+json"
  }],
  "properties": {
    "fancy_lampProp1": {
      ...
    },
    "fancy_something": {
       ...
    }
  }
}
```

At which point you again cannot determine which property affordance comes from the "own" TM and which ones from a submodel.

So I am wondering if this "identification via naming" is the right approach - or if this should even be determinable at all.

sebastiankb:
Good catch! 

I think what we can do here is to provide a hint that a consumer should also check where the interaction affordances are coming from by looking up the linked TMs. Identification by instanceName might not be sufficient.

sebastiankb:
I just came from an internal discussion and we are still quite unsure how to handle this issue smoothly.  

A straightforward solution would be for all affordances names, as defined in the origin sub-TMs, to be copied 1:1 into the self-contained TD (without a preceding prefix). This would keep the 'naturality' of the origin TM definition (take the name as defined by the author). 
This solution is fine as long as there is no name collusion. If there is a collusion then a self-contained TD cannot be created (similar in JAVA programming language).

To overcome this conflict, different strategies can be considered:

1. no self-contained TD can be generated: only hierarchy based TDs can be generated ([example](https://w3c.github.io/wot-thing-description/#example-61))
2. if you own the TMs: change all the affordances names that potential causes collisions in a TM composition
3. rename affordances in the TD instance that would causes name collusions, e.g., by the `instanceName` prefix approach
4. per default, all sub-TM affordances are renamed in the TD instance (e.g., as shown in this [example](https://w3c.github.io/wot-thing-description/#example-self-contained-td-of-the-smart-ventilator-tm)).  

Because of these different variations of strategies (with their advantages and disadvantages), we would suggest that no specific approach should be recommended or specified by the TD 1.1 spec. We would simple keep the assertion that name collusion should be avoided (also suggested [here](https://github.com/w3c/wot-thing-description/issues/1556#issuecomment-1171164127) by @thjaeckle).  That's it. The[ Example 62](https://w3c.github.io/wot-thing-description/#example-self-contained-td-of-the-smart-ventilator-tm) will be also updated with no instanceName prefix in the affordance names.

What do you think?

@thjaeckle @relu91

thjaeckle:
@sebastiankb that's fine with me.
Naming collisions are even not allowed in JSON - so normally you won't be able to even compose the TD with colliding e.g. property names - as your used JSON library should already prevent that.

egekorkan:
> @sebastiankb that's fine with me. Naming collisions are even not allowed in JSON - so normally you won't be able to even compose the TD with colliding e.g. property names - as your used JSON library should already prevent that.

Also, the TD spec asserts this too, in case you serialize to other formats that allow collisions:
-  Actions collected in the JSON-object based actions member MUST have unique JSON names. (same for properties and events too)

relu91:
I would support this approach too. Also, people can fall back to the non-self-contained TD when collisions occur. I would say that for the time being is a good compromise. 

For the future, I've found interesting the proposal by @ashimura to leverage JSON-LD prefix features. For example:
```js
{
// ... other prop
"links": [
        {
            "rel": "item",
            "href": "http://my.tms.com/Ventilation.td.jsonld",
            "type": "application/td+json"
        },
        {
            "rel": "item",
            "href": "http://my.tms.com/LED.td.jsonld",
            "type": "application/td+json"
        },
        {
            "rel": "type",
            "href": "http://my.tms.com/SmartVentilator.tm.jsonld",
            "type": "application/tm+json"
        }
    ]
    // ...
}
```
TD generated with JSON-LD prefixes
```js
{
"@context" : [
  // other TD contexts
  { 
      "ventilation": "http://my.tms.com/Ventilation.td.jsonld#",
      "led": "http://my.tms.com/LED.td.jsonld#",
      "smart": "http://my.tms.com/SmartVentilator.tm.jsonld#"
  }
 "properties" : {
    "ventilation:switch" : { /* ... */ },
    "led:R": { /* ... */ }
 }    
// ... other prop
```

I found this solution compelling, but I think we all agree that it bends the JSON-LD semantic maybe too much. Something worth discussing for TD 2.0.

sebastiankb:
from today's TD call:
- decided to merge
- @relu91 this is an interesting approach which would now also work with this PR. However, we should check if this is really working out since we are not defining a @context map with TMs. 
- if it is working we should consider this for TM 2.0

Issue #1612, opened 2022-07-24
[OracleReview] 5.3.2.7 StringSchema


egekorkan:
- [ ] contentEncoding: please add link to RFC
- [ ] contentMediaType: please add link to RFC
- [ ] RFC2119 assertion: "Since this value is not a fragment identifier, and is not relative to the root of the TD but to whichever data schemas the security scheme is bound to, this value SHOULD not start with #;"

sebastiankb:
from today's TD all, group like to close this issue
@mlagally please can you check of this is ok and your issue was addressed

egekorkan:
Call of 05.04: The issue is addressed in #1650. Closing

PR #1650, opened 2022-08-09
update references for contentEncoding  and contentMediaType + extand assertation with another sentance

sebastiankb:
solves #1612

@mmccool please can you check the extended security `body` assertion if this is ok for you.


egekorkan (approved):

sebastiankb:
from today's TD call:
-  decided to merge
- @mlagally it is intended to not to assign it as an assertion since is just an informative text.

Issue #1617, opened 2022-07-24
 [OracleReview] 6.3.4 securityDefinitions and security


egekorkan:
- missing normative assertions:

"A named security configuration must be given in the securityDefinitions map. That definition must be activated by including its JSON name in the security member, which can and should be of type string when only one definition is activated."

egekorkan:
I am a bit confused about this issue. @mlagally do you mean that we need this new text or that there is a text like this and we should add the proper assertions?

In any case, we do not force arrays to turn into strings if there is only one content in them

mmccool:
- The first sentence is basically saying the securityDefinitions Map must not be empty.  It's already mandatory, but I don't know if we say anywhere that Maps must not be empty.  If we do it should be a general statement for all Maps and then we don't really need this sentence (as an assertion; could still state it informatively for clarity).
- First part of the second sentence (note this sentence should be broken up into two separate statements) is about relating the values used in "security" with the named objects in securityDefinitions.  We do need to limit the values used to those defined in securityDefinition objects.
- Second part of the second sentence relates to my intention to deprecate using arrays as values for "security" and using "combo" instead (it is clearer).   Technically both single strings and arrays are still allowed for backward compatibility.

Not quite sure if the above are not handled elsewhere in other assertions.  I need to double-check; we don't want duplicate assertions about the same requirement if we can help it.

mlagally:
@egekorkan I suggest this existing text should be vboken down into two normative assertions.
The first sentence requires at least one security definition. 
The second sentence is a requirement on the value of the security member.
Both should be MUST assertions.
@mmccool I think the current text is very specific and has more constraints than just forbidding empty maps.

sebastiankb:
just for information: those two sentences already exists in TD 1.0

sebastiankb:
from today's TD call: 
- @mmccool will check if this issue in the security TF and will look if there is already assertions existing that address those two statments

egekorkan:
I have checked our JSON Schema. We are not allowing an empty map for `securityDefinitions` but we currently allow an empty array for `security`. I will fix it there but a text is needed to make this more assertive

sebastiankb:
I double check this issue and it seems that the sentence is now rephrased in Section 6.3.4. In my view the issue is solved.

sebastiankb:
from today's TD call, decided to close this issue

PR #1653, opened 2022-08-10
MinLength for security array

egekorkan:
This addresses [one part of ](https://github.com/w3c/wot-thing-description/issues/1617#issuecomment-1210257397)

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -32,7 +32,8 @@
           "type": "array",
           "items": {
             "type": "string"
-          }
+          },
+          "minItems":1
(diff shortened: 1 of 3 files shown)


egekorkan:
I have not updated the tm schema intentionally since the array can be empty there

sebastiankb (approved):

sebastiankb:
from today's TD call:
- PR is ok
- we should check the script that takes over the definition to the TM Schema definition
- @egekorkan will check later and will merge this PR

egekorkan:
@relu91 indeed I had to add some extra logic to the generator

PR #1654, opened 2022-08-10
[schema] remove trailing space

egekorkan:
This was actually causing our schema to be invalid against the JSON Schema Meta Schema (a schema to validate schemas) and a popular python implementation (https://github.com/python-jsonschema/jsonschema) was not validating any TDs at all. Since we are using draft 7, the use of items is only possible with an object as a value.

This however means that we should also not allow this in our array schema? I am a bit confused since I see different behavior in different JSON Schema validators. I will check with JSON Schema community to see what can be really happening

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -2,7 +2,7 @@
   "title": "Thing Description",
   "version": "1.1-10-June-2022",
   "description": "JSON Schema for validating TD instances against the TD information model. TD instances can be with or without terms that have default values",
-  "$schema ": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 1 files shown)


sebastiankb:
from today's TD call:
- decided not to merge
- we will wait until JSON Schema authors will response to Ege's email. Maybe there are some comments which would be good to know

egekorkan:
Our mistake was actually quite silly, a trailing space in `$schema` meaning that some tools would pick a JSON Schema draft version that they "like" and if they like 2019-09, this term becomes invalid. Thanks @Julian from JSON Schema Community for spotting this!

egekorkan:
merging since it was a silly mistake in the end...

PR #1662, opened 2022-08-10
Fix a trailing space in the a JSON Schema's $schema

Julian:
This was ignoring it previously! As `$schema ` is unfortunately different from `$schema`.


egekorkan:
I will take over the changes, thank you!

PR #1676, opened 2022-08-17
Tweak "out-of-band" in td-security-no-secrets assertion

### LINKED ISSUE #1669: Tweak "out of band" wording
sharing secrets out of band is a consequence of not putting them in the TD... tweak wording of td-security-no-secrets assertion

mmccool:
Resolves #1669 

New wording of assertion:

For all security schemes, any keys, passwords, or other sensitive information directly providing access MUST NOT be stored in the TD and should instead be shared and stored out-of-band via other mechanisms. 

Changes:
- Removed "private" in front of "keys".  This has a specific meaning in asymmetric cryptography and in fact private keys are NEVER shared.  But symmetric keys may be.  Anyway, use of "private" will cause unnecessary confusion.
- Rephrased to say "instead" for out-of-band sharing so it is clearer that this is what should be done rather than putting this information in the TD.

Don't agree with splitting this into two assertions since then we will just have two redundant assertions saying the same thing.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-thing-description/pull/1676.html" title="Last updated on Aug 17, 2022, 5:12 PM UTC (ed61348)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1676/883f222...mmccool:ed61348.html" title="Last updated on Aug 17, 2022, 5:12 PM UTC (ed61348)">Diff</a>


egekorkan (approved):

sebastiankb:

PR #1677, opened 2022-08-17
Define what it means to satisfy a security scheme

### LINKED ISSUE #1667: Define what "satisfying" a securityDefinition means
to avoid adding an extra assertion to the table defining "security", add some *informative* text to the opening section of Security Schemes to state what "satisfying" a security scheme means (it means the client needs to provide all necessary credentials, correctly respond to any challenges, etc.)

mmccool:
Resolves #1667 

Add following informative text to the "Security Scheme" section to define what it means to "satisfy" a security scheme.  This *intentionally* does not talk about details (e.g. credentials, protocols, etc), as these will be defined by each scheme.

> Each security scheme object used in a TD defines a set of requirements to be met before access can be granted. We say a security scheme is satisfied when all its requirements are met. In some cases requirements from multiple security schemes will have to be met before access can be granted.


egekorkan (approved):

relu91 (approved):

sebastiankb:
from today's TD call, decided to merge.
@mmccool will resolve the conflict and will merge this PR then

PR #1712, opened 2022-10-05
Add table numbers and captions using new respec option

egekorkan:
This is only a draft at the moment since some tables are generated by STTL.js . I need to see if I can change sttl.js or do some post editing of the index.html in a systematic way.

### DIFF
--- a/index.html
+++ b/index.html
@@ -825,7 +825,8 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
         as follows:
     </p>
   
-    <table class="def">
+    <table class="def numbered">
+    <caption>Namespaces used in TDs</caption>
     <thead>
         <tr>
             <th>Vocabulary</th>
@@ -856,7 +857,8 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
     </p>
 
 
-    <table class="def">
+    <table class="def numbered">
+        <caption>Namespaces used in TMs</caption>
         <thead>
             <tr>
                 <th>Vocabulary</th>
@@ -1565,7 +1567,8 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
 	  <p><span>The following table contains content types which MAY use data schema to describe the 
         structure of their payloads.</span></p>
 	  <p>
-         <table class="def">
+         <table class="def numbered">
+            <caption>Content types that can use a Data Schema</caption>
             <thead>
                <tr>
                   <th>Format</th>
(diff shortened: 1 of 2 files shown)


egekorkan:
The reason why I am doing this is to make it easier for people to reference to tables

egekorkan:
So I have managed to change some other files to have numbers for all the tables. However, a lot of them have the same caption since they all use the same template in the templates.sparql

egekorkan:
So I have managed to change some other files to have numbers for all the tables. The caption of some look the same at the moment.

I have also deleted the part of the rendering that includes json schema into the spec.

templates.sparql
```diff
@@ -50,7 +50,8 @@ template :classes(?ns) {
 } order by ?rank
 
 template :fields(?class ?ns) {
-    "<table class=\"def\">"
+    "<table class=\"def numbered\">"
+        "<caption>Vocabulary Terms</caption>"
```
egekorkan, line 54: This needs some more work to make it unique

egekorkan:
Maybe @relu91 or @ethieblin can help here regarding how to pick the correct string for the captions. At the [comment above](https://github.com/w3c/wot-thing-description/pull/1712#discussion_r1012267587), I have left it at a working stage. I have tried the following:
- `"<caption>Vocabulary Terms at the <a href=\""?class"\">Current Level</a></caption>"` but this puts a string with `<` and `>` around it. 
- `"<caption>Vocabulary Terms at the <a href=\""str(?class)"\">Current Level</a></caption>"` this does not seem to work. I have tried strbefore as well

Both render it as the following:
image

I am using this inside a format, i.e.:
```
        format {
        "<caption>Vocabulary Terms at the <a href=\""?class"\">Current Level</a></caption>"
        # "<caption>Vocabulary Terms at the <a href=\""str(?class)"\">Current Level</a></caption>"
        }
```

templates.sparql
egekorkan: Worst case we just say "at the Current Level"

relu91:
I think you have to use `concat` -> `concat("<caption>Vocabulary Terms at the <a href=\"", ?class,"\">Current Level</a></caption>" )`

templates.sparql
egekorkan: this caption is needed in order to generate the numbers

egekorkan:
I have done some improvements but there is some weird behavior that I have documented in the comments

sebastiankb:
from today's TD call, not ready to merge.

@relu91 will have a look as well

PR #1683, opened 2022-08-19
Change to tm optional in validation

egekorkan:

PR #1715, opened 2022-10-05
Add model and prohibit instance in tm schema

### LINKED ISSUE #1713: Prohibit instance from TM Schema
There is the assertion at the spec saying:

> Due to the definition of Thing Model the term instance MUST be omitted within the version container.

However, we do not prohibit this in the TM JSON Schema and we should.

egekorkan:
fixes #1713

### DIFF
--- a/validation/tm-json-schema-validation.json
+++ b/validation/tm-json-schema-validation.json
@@ -2,7 +2,7 @@
   "title": "Thing Model",
   "version": "1.1-10-June-2022",
   "description": "JSON Schema for validating Thing Models. This is automatically generated from the WoT TD Schema.",
-  "$schema ": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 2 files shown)


egekorkan:
I have taken over the suggestions and changes from #1683 . That one can be closed or merged first

egekorkan:
I have also added regexes to check that `tm:optional` uses the correct syntax defined by:
> The JSON Pointers of tm:optional MUST resolve to an entire interaction affordance Map definition.

We cannot check this entirely with a JSON Schema, i.e. if there is no property `myAmazingTemp`, it is still possible to use `/properties/myAmazingTemp` but at least it is not possible to do `/properties/myAmazingTemp/type`. This will be done in playground at least.

I have tested the regexes with an online tool and put links to in the comments.

sebastiankb (approved):

sebastiankb:

PR #1725, opened 2022-10-14
Oct 2022 impl report updates

mmccool:
- update with latest results and playground tests - resolved 6 at-risk assertions (Basic security scheme, thing-model assertions)
- some reversions too, e.g. tm-placeholder-usecase and context-default-language; see diffs

PR #1726, opened 2022-10-18
Enumerate At-Risk Items

mmccool:
- Add content to sotd to explicitly list at-risk items (linking to assertions representing each feature)
- The set of at-risk features is not based on the atrisk.csv in the current repo, but on the [pending PR updating the atrisk items with the latest results](https://github.com/w3c/wot-thing-description/pull/1725)
- Since many of the at-risk assertions are bunched up in the S&P Considerations sections, mention in opening paras of these sections how these will be dealt with if not completed by PR (will be converted to informative statements, not deleted)
- This text is consistent to what I am also proposing for Discovery and Architecture
- I have put this into the "main" top-level index file due to other pending PRs, to avoid having to re-sync a copy in publication/ver11.  Note that for Architecture and Discovery though I only added this to the CR version.  We can delete these comments after the copy for the CR transition is taken from the top-level version if necessary.

### DIFF
--- a/index.html
+++ b/index.html
@@ -434,6 +434,27 @@ a[href].internalDFN {
       <li>A <a href="#td-writeall-consumer">behavioral assertion</a> for <code>writeallproperties</code> that allows rejection of incomplete writes.</li>
     </ul>
 -->
+    <p>At the time of CR transition, due to insufficient implementation
+    experience the following features are at risk:</p>
+    <ul>
+    <li><a href="#sec-body-name-json-pointer">Indicating location of security information in body of payload using JSON pointers.</a></li>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -6876,7 +6897,7 @@ <h2>Derivation of Thing Description Instances</h2>
 	"title": "Smart Pump",
 	"id": "urn:example:123-321-123-321",
 	"description": "Smart Pump live plant and simulator",
-	"version": {"instance": "1.0.0", "model": "1.0.0" },
+	"version" : {"instance": "1.0.0", "model": "1.0.0" },
```
mmccool, line 6900:
```suggestion
	"version": {"instance": "1.0.0", "model": "1.0.0" },
```

PR #1736, opened 2022-11-01
CR exit criteria

mmccool:
- Add paragraph specifying CR exit criteria to sotd section in publication/ver11/5-cr
- List ALL statements at risk in sotd (overlooked some Thing Model ones, and added S&P ones)
- Updated IR template - tooltips
- Rerun index.html -(ReSpec export) -> static.html -(htmltidy)-> Overview.html processing

I also rebased before running this and seem to have picked up some edits to the change log.  Please check that the result is correct, it looks like just formatting changes.  Note that "static.html" and "Overview.html" get overwritten by the above processing so any edits should be made to "index.html".  The old version did not have the links filled in which should have happened with the ReSpec export above so I think "static.html" might have been mistakenly edited.

Update: Sebastian has checked and the change log looks to be ok.

### DIFF
--- a/publication/ver11/5-cr/Overview.html
+++ b/publication/ver11/5-cr/Overview.html
@@ -639,8 +639,8 @@ var[data-type]:hover::after,var[data-type]:hover::before{opacity:1}
       ]
     }
   ],
-  "publishISODate": "2022-10-21T00:00:00.000Z",
-  "generatedSubtitle": "W3C Candidate Recommendation Snapshot 21 October 2022"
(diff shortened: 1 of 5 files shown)


sebastiankb (approved):

mmccool:
Sorry, just added a few more things to this PR.  Apologize for lumping somewhat unrelated things together, but I don't think any of these are controversial and wanted to avoid conflicts:
- Listed ALL at-risk assertions in sotd section (this is another CR transition requirement; if we leave any out, it may delay the transition)
- Updated IR report template to add tooltips to table headers

PR #1737, opened 2022-11-02
Add explanation about observeproperty relationship

### LINKED ISSUE #888: [question][behavior][testing] Should writing to an observable property trigger new values being pushed
From the Consumer perspective, if the Consumer writes to a property while also observing it, should it definitely get a notification of the value it wrote? For a Thing implemented by node-wot, it is the case but can this be a guaranteed behavior? 
This is an interesting case for testing since at the moment we cannot test event affordances of a Thing automatically since the Consumer does not know how to trigger them. However, the observability of a writable property can be tested if the previous question is valid. 
There were short discussions on whether writing and then reading should return the written value and it was decided that it cannot be guaranteed since a process can change the property value in the meantime. However, I think that this question is different.

P.S. This issue can be moved to wot-testing as well.

egekorkan:
fixes #888

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1737.html" title="Last updated on Nov 2, 2022, 1:54 PM UTC (b43f0af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1737/6e2d51a...b43f0af.html" title="Last updated on Nov 2, 2022, 1:54 PM UTC (b43f0af)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2525,6 +2525,10 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
             </tbody>
             </table>
         </p>
+        <p class="ednote" title="writeproperty and observeproperty relationship">
+            Writing to a property does not necessarily mean that a new value will be sent to the Consumer observing the property. 
+            It depends on the protocol and implementation.
+        </p>
         <p class="ednote" title="Data Schemas of meta operations">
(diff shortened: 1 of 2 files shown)


sebastiankb:

PR #1752, opened 2022-11-25
Update Implementation Report (Indexing)

mmccool:
This PR makes some changes to how the implementation report and CSV files are organized:
- template.csv now has an "Index" column and an "Assertion" column
- the assertions in the CSV file and in the report are assigned indices in order by where they appear in the text, and are sorted by these indices in both places
- The indices are integers starting at 1 except for child assertions, which are fractions that start at x.01 where x is the index of the parent assertion.  This is so when sorted they appear directly after their parent assertion.  These fractional numbers are also assigned in the order in which they appear in the text
- "Extra assertions" defined in "extra_assertions.html" are assigned numbers after the last one in the main text (if parent assertions, of which I don't think there are any) or fractional indices after the indices of the parent assertion in the main text.  The fact that child assertions are (sometimes) defined in a separate file was my main motivation for using fractional indices for them.
- The testing/template.csv file now has an extra column, "Assertion", which has the text of the assertion.  This has been stripped of HTML markup, has had newlines turned into whitespaces, and also has " characters escaped to avoid CSV errors.  Note that to escape a " in CSV you have to double it, like this: ""; my tool does this automatically (and yes, some assertions do have quotes in them...).  It seems to format ok in github and I checked it with a CSV linting tool, but let me know if you see any CSV errors.
- The atrisk.csv and atrisk.css files have the same content as before but have been reordered.  So there are a lot of changes noted in the diff but the actual set of atrisk assertions has not changed.

NOTE: These changes do not require any updates to existing test results.  The ordering may be different but my tool only looks at the "ID" and "Status" fields of inputs.  The index and re-ordering is only for convenience for capturing inputs.  New results can use the new template without any difficulty, although we'll have to check if the playground can also use CSV files with the new columns e.g. for manual results. 

NOTE 2: These changes were made at the request of the Architecture and Profile task forces but I'm applying them to TD for consistency, and because it's the more difficult test case (has child, table, default assertions, etc all of which made things more complicated).


mmccool:
Note also the source of the implementation report (and so the assertion text) is publication/ver11/5-cr/Overview.html, not index.html.  I will wait until after the CR transition is complete, and then will start pulling input from index.html (at the top level) instead in prep for the PR transition.

sebastiankb (approved):

Issue #1742, opened 2022-11-14
Fix assertion extraction tool to have gaps between terms


egekorkan:
It can be seen in many places but one example is https://github.com/w3c/wot-thing-description/blob/main/testing/assertions.csv#L81

The keyword that was wrapped in `<code>` does not have space before and after, which makes it difficult to read

egekorkan:
@mmccool can provide a regex for fixing this

PR #1754, opened 2022-12-11
Sync manual csv with playground and new titles

egekorkan:
The header has changed from Description to Assertion (in template.csv and assertions.csv) so I am adapting to it. Otherwise, no major addition or removal of individual assertions


mmccool:
Your new files have the slight problem that when an assertion goes over more than one line, you have combined them without a space, so words are getting run together.  You should convert newlines to spaces.  It's also a good idea to get rid of leading spaces, trailing spaces, convert tabs to spaces, and convert multiple spaces to one space (all junk I had to do when I generated template.csv...)

mmccool:
Note that my tooling only cares about ID and Status and ignores the other columns.

mmccool:
I will pull the regexes from my implementation so you can clean this up...

egekorkan:
The regex should go here: https://github.com/w3c/wot-thing-description/blob/main/testing/extractFile.js

Also once you put that, this issue will be hopefully fixed: https://github.com/w3c/wot-thing-description/issues/1742

mmccool:
See PR #1756; not sure which one you want to merge first (might get conflicts in assertions.csv but these can be corrected by regenerating the file).

egekorkan:
> See PR #1756; not sure which one you want to merge first (might get conflicts in assertions.csv but these can be corrected by regenerating the file).

@mmccool thank you for the other PR. The other one needs to merged first, I will come back to it afterwards.

egekorkan:
I have checked after the testfest and there is no new changes other than this here

PR #1755, opened 2022-12-12
Implementation Report Update - Dec 2022

mmccool:
Include new results from Dec 2022 Testfest.


k-toumura:
[Rendered version (c324c90)](https://cdn.statically.io/gh/w3c/wot-thing-description/c324c905e77e7da59cb940bc21c11caafb280c5c/testing/report11.html)
Number of at-risk assertions: 25.

mmccool:
Number of at-risk assertions: 22.

k-toumura:
[Rendered version (c6e3dde)](https://cdn.statically.io/gh/w3c/wot-thing-description/c6e3dde3b5cbbaae78aab70bc98e09d59e6575af/testing/report11.html)

mmccool:
As of the final update, there are now only 18 at-risk assertions.  However, the playground does not seem to be picking up the client flow implemented by wot-experimental, which should resolve two more assertions, at least.  So once we resolve that issue I will make another PR to update the report.  However, we plan to merge this so it can count as the "Testfest result".

PR #1756, opened 2022-12-13
Fix Assertion Text in extractFile.js script

mmccool:
Improve the regex to clean up assertion text extract from the spec.  Also regenerates testing/assertions.csv; some assertions were missing some spaces, some had tabs, etc.


egekorkan (approved):
Changes are only in the descriptions and no influence on the ids. This can be definitely merged.

mmccool:
Will leave it up to you when to merge.  I will have to copy to the other repos which may be doing something similar... will ask in the testfest meeting tomorrow.

egekorkan:
@mmccool let's merge now.

PR #1758, opened 2022-12-21
Updating at risk assertions in the main body

egekorkan:
This syncs the index.html with https://github.com/w3c/wot-thing-description/blob/main/testing/atrisk.csv


egekorkan:
This should be done after CR transition

PR #1772, opened 2023-02-08
Add table numbers and captions using new respec option v.2

relu91:
This PR contains the changes of #1712 with an additional commit to fix the renderer. The history was a little messed up so I pointed directly to the main branch.


danielpeintner:
_No review, just a comment/question._

I can see the table numbering where there wasn't  anything before.
e.g., in [4. Namespaces](https://w3c.github.io/wot-thing-description/#namespaces) 👍 

grafik

Unfortunately I cannot click on it nor is there an ID generated so that I can point people to a specific table like
`#id-of-my-table-about-X`

Do you think this is possible also since it would be very useful.

danielpeintner:
Note: There is an id, see
https://pr-preview.s3.amazonaws.com/relu91/wot-thing-description/pull/1772.html#table-namespaces-used-in-tds

Unfortunately one needs to look into the HTML source code since the Table header is not clickable like for examples or so.

relu91:
@danielpeintner unfortunately there is no built-in way to make the `Table n°` clickable: respec controls the [generation](https://github.com/w3c/respec/blob/bb474bfa23725105b891ef54d2bd64af159551d2/src/core/tables.js#L64) of the caption label. However, in 180da57 I made the whole caption a link. What do you think? I'd be happier if I could add the link only in the `Table n°` but this way might still work. 

**Note**: the other table captions need to be updated too.

danielpeintner:
There does not seem to be any link for *some* tables (Table 1 and 2)?
https://pr-preview.s3.amazonaws.com/relu91/wot-thing-description/pull/1772.html#table-namespaces-used-in-tds

I see that for Table 3 there is a link...
see https://pr-preview.s3.amazonaws.com/relu91/wot-thing-description/pull/1772.html#table-vocabulary-terms-in-thing-level

BTW, there is an open feature request: https://github.com/w3c/respec/issues/4222 and we just need to wait...

relu91:
> EDIT: I think changing just some was intentional, right @relu91 ?

Yes it was, I just didn't want to change all if you don't decide to go in that direction

> BTW, there is an open feature request: https://github.com/w3c/respec/issues/4222 and maybe we just need to wait...

Ah cool, somehow I missed it. Let's decide int today's call.

egekorkan:
Call of 22.02: We will wait for respec to have the automatic links otherwise the manual part can be done.

relu91:
PR in flight: https://github.com/w3c/respec/pull/4392

relu91:
> PR in flight: [w3c/respec#4392](https://github.com/w3c/respec/pull/4392)

PR merged, release should be done soon.

egekorkan:
@relu91 given that the release has also happened, what should we do with this PR?

egekorkan:
Call of 22.03: No objections, merging since it is an informative change.

PR #1790, opened 2023-04-05
Editorial fixes for issue 1735

### LINKED ISSUE #1735: Minor editorial changes for PR
From comment [#1722](https://github.com/w3c/wot-thing-description/issues/1722#issuecomment-1289099039):

However, I found the following minor issues and would suggest we fix them:
* missing "a" for "as JSON object" in some of the assertions, e.g., td-multi-languages_descriptions
    * should add "a" and make it "as a JSON object"
* missing values between commas (",") at td-processor	Syntax
    * should add actual values between commas
* ambiguous "included" at td-security-combo-exclusive-oneof-or-allof	
    * should clarify Exactly one of either oneOf or allOf MUST be included in what
* ambiguous usage of "token" and "authorization", e.g., at td-security-oauth2-client-flow, td-security-oauth2-client-flow-no-auth, td-security-oauth2-code-flow and td-security-oauth2-device-flow
    * would be better to explicitly state "the token element" and "the authorization element", or clarify those are specific keywords meaning the Thing Description elements (e.g., using specific style like bold italic).
* ambiguous usage of "form level" and "Thing level" at td-security-overrides
    * would e better to explicitly sate "the level of the form element" and "the level of the Thing element", or like the above point, better to use some specific style (e.g., bold italic) for "form" and "Thing" within this assertion

egekorkan:
fixes #1735 

This changes the assertion text but not the meaning of the assertions so these are just clarifications.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1790.html" title="Last updated on Apr 26, 2023, 3:10 PM UTC (af89a4b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1790/31de580...af89a4b.html" title="Last updated on Apr 26, 2023, 3:10 PM UTC (af89a4b)">Diff</a>


egekorkan:
Call of 05.04:
- We should not use italics
- We should add terminology for 
  - Thing level
  - Form level
  - Affordance level
  - Data Schema level
  - and say generally "... level" means inside that object

egekorkan:
I have incorporated the feedback. There are further "changes" to assertions but these are simply adding a link to the definitions that  are added in the terminology.

index.template.html
```diff
@@ -851,6 +851,18 @@ <h2>Terminology</h2>
         The default representation format for <a>Thing Descriptions</a> is JSON-based as defined by
         this specification.
     </dd>
+    <dt>
+        <dfn id="dfn-td-level" class="lint-ignore">Levels of a TD</dfn> (including
+        <dfn id="dfn-td-thing-level">Thing Level</dfn>, <dfn id="dfn-td-affordance-level" class="lint-ignore">Affordance Level</dfn>,
+        <dfn id="dfn-td-thing-level" class="lint-ignore">Data Schema Level</dfn>, <dfn id="dfn-td-thing-level" class="lint-ignore">Forms Level</dfn>)
```
egekorkan, line 857:
```suggestion
        <dfn id="dfn-td-data-schema-level" class="lint-ignore">Data Schema Level</dfn>, <dfn id="dfn-td-forms-level" class="lint-ignore">Forms Level</dfn>)
```

egekorkan:
Call of 26.04: We will revisit with more participants.

egekorkan:
Call of 17.05: Terminology contains an addition but that is informative. The rest are small changes.

PR #1791, opened 2023-04-05
Add references to TD 1.0

### LINKED ISSUE #1770: Adding link to 1.0 REC
With the publication of the CR, we have lost a convenient link to the 1.0 publication, i.e. https://www.w3.org/TR/wot-thing-description/ points to TD 1.1 and this is apparently correct. So I propose to simply adding a "Previous Recommendation" link like JSON-LD 1.1 https://www.w3.org/TR/json-ld11/ . It would be the dated URL.

egekorkan:
I have added it to respec config and to some places in the text by making sure to use [wot-thing-description10] . Respec config says to use `prevRecURI: "https://www.w3.org/TR/2014/example-20140327/",` but this generates a link with name "Last Recommendation" which I find confusing. You can check the previous commits to see it.

fixes #1770

[JSON LD 1.1](https://www.w3.org/TR/json-ld11/) manages to do it but I get different behavior

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1791.html" title="Last updated on Apr 5, 2023, 12:30 PM UTC (29ede36)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1791/31de580...29ede36.html" title="Last updated on Apr 5, 2023, 12:30 PM UTC (29ede36)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -186,6 +186,15 @@
             },
             otherLinks: [
                 {
+                    key: "Previous Recommendation",
+                    data: [
+                        {
+                            value: "https://www.w3.org/TR/2020/REC-wot-thing-description-20200409/",
+                            href: "https://www.w3.org/TR/2020/REC-wot-thing-description-20200409/"
+                        }
+                    ]
+                },
+                {
                     key: "Contributors",
                     data: [
                     {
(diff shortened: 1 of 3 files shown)


egekorkan:
Call of 05.04: There is a change to the assertion text but no effect on implementations. Merging

PR #1802, opened 2023-04-18
Fix html checker errors

### LINKED ISSUE #1314: [Editorial] HTML checkers shows errors
see https://validator.w3.org/nu/?doc=https%3A%2F%2Fw3c.github.io%2Fwot-thing-description%2F

* no end tag
* end tag without start tag
* ...

egekorkan:
fixes #1314 

Currently, we have 60 warnings/errors/info . A new check by uploading the index.html has now 14 which are only info.

There is one change to an assertion text that breaks a list into normal text but does not change the meaning, so no problem with the publication process.

Please hide white spaces when reviewing.

The main things I am fixing that needs attention from future editors:
- Do not put a list inside a paragraph. See [here](https://stackoverflow.com/questions/5681481/should-ol-ul-be-inside-p-or-outside).
- Do no put a pre (examples) inside a paragraph

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1802.html" title="Last updated on Apr 27, 2023, 8:36 PM UTC (b27176f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1802/75a92bd...b27176f.html" title="Last updated on Apr 27, 2023, 8:36 PM UTC (b27176f)">Diff</a>


egekorkan:
Call of 26.04:
- We will wait for further reviews.
- Automated checks should be used later on
- We should use HTML tidy instead.
- @egekorkan will fix the conflicts

egekorkan:
I have tried using HTML Tidy and it is pretty good. However, it creates such a big diff that we cannot review it at this stage... Please see https://diffy.org/diff/17470c91a1032 to understand why.

For the next charter, we can make it mandatory to use but I think I would still prefer something more generic since we have other files than HTMLs to maintain like our JS, JSON, TTL files. One thing would be to apply eslint and prettier on those files since they are development-oriented and then run HTML tidy on the output HTML since that is what will be used before publication. I think that tooling including specwork tooling should be a priority next charter.

danielpeintner:
I noticed that we have one more warning...

image

egekorkan:
> I noticed that we have one more warning...
> 

@danielpeintner I have introduced that definition but forgot to use it. It is now used.

egekorkan:
Call of 17.05: We can merge. For understanding the diff, htmldiff should be used instead of git diff.

PR #1809, opened 2023-04-29
Regenerate TM JSON Schema

JKRhb:
This (trivial) PR regenerates the TM JSON Schema and incorporates the updated version number from #1786.


egekorkan:
Merging without meeting since it is trivial, no implication to implementations and can be important for test inputs.

PR #1812, opened 2023-05-03
Update manual.csv

egekorkan:
Some manual assertions were still there


mmccool (approved):
will also need updates in report

PR #1816, opened 2023-05-10
May 2023 IR - manual classification updated

mmccool:
- manual.csv was updated
- this PR moves entries from one table to another in the report, but does not change any statuses


egekorkan (approved):
The diff might look big but it is just moving things around

PR #1823, opened 2023-05-17
Remove manually added rfc classes

### LINKED ISSUE #1806: Align usage of RFC2119 class
We seem to have a mixed use the RFC class.

If a span has the class, any capitalized MUST, SHOULD, etc. automatically gets a <em class="rfc2119">SHOULD</em> wrapping. However, we seem to add it manually _sometimes_ like at https://github.com/w3c/wot-thing-description/blob/main/index.html#L2345

We should simply never add it manually and let respec do it for us.

egekorkan:
fixes #1806

I have also did respec exports of the HTML (editors draft and this PR): https://diffy.org/diff/2f9d0076a9a42

My observations: 
- We were manually adding a title attribute which is not added by respec
- All capitalized forms are recognized, including MUST NOT etc.

I would still classify this as an editorial change. Just need a quick confirmation from @mmccool that his tool is not dependent on these manual additions. I guess it is not since it was not used everywhere, just some random locations.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1320,8 +1320,7 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
           "td-context-ns-thing-mandatory">
 
           The <code>@context</code>
-          name-value pair <em class="rfc2119" title=
-          "MUST">MUST</em> contain the anyURI
+          name-value pair MUST contain the anyURI
           <code>https://www.w3.org/2022/wot/td/v1.1</code>
           in order to identify the document as a TD 1.1 which would allow <a>Consumers</a> to use the newly introduced terms.
           </span>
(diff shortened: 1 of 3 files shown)


egekorkan:
Call of 17.05:
- We wait for @mmccool opinion but can be merged if he is ok
- We should check if other specs in W3C do something similar
- The diff linked above shows that respec was adding a second class unnecessarily. It is a bug of respec but it is our misuse in the first place.
- We should check our other WoT specs to see if they do the same mistake.

mmccool:
Yeah, my tooling does not need this extra markup, it looks for the capitalized words themselves.  My only concern is if the "rfc2119" class changes formatting but is only used in some places then the document would be formatted inconsistently.  I think Architecture uses them all over the place but should probably be removed as well.

In short, I am ok with this PR.

mmccool (approved):

PR #1827, opened 2023-05-18
Resolve security-context-secure-fetch

mmccool:
- Modifies at-risk assertion "security-context-secure-fetch" due to insufficient implementations (0/2)
- Downgrade to informative statement by converting "SHOULD" to "should" and commenting out assertion markup
- See https://github.com/w3c/wot-testing/tree/main/events/2023.03.Online


egekorkan (approved):

egekorkan:
Call of 24.05: We have decided to merge this PR to remove the assertion and comment out the RFC spans. The implementations that do it can still do it.

PR #1828, opened 2023-05-18
Resolve td-security-oauth2-device-flow

mmccool:
- Removes at-risk assertion "td-security-oauth2-device-flow" due to insufficient implementations (0/2)
- Also removes (comments out, actually) all other informative content (including table column) related to the OAuth2 "device" flow, and reference to RFC defining it
- Ontologies updated, then index.html re-rendered
- Validation files also updated.  Content commented out where possible, otherwise removed (e.g. no comments are supported in JSON files, so "device" was removed in both the TD and TM JSON Schemas).
- See https://github.com/w3c/wot-testing/tree/main/events/2023.03.Online


egekorkan (approved):

egekorkan:
Call of 24.05: We have decided to merge this PR to remove the assertion and comment out the RFC spans. There have been no implementations of this.

PR #1837, opened 2023-05-31
CR Publication Prep

egekorkan:


egekorkan:
Call of 07.06:
- New feature note should be added

egekorkan:
> @egekorkan Did you run HTML tidy also? I did this in the past for CR publication and it caught a lot of HTML errors (fixes for which I then merged back into the main doc, but more may have been introduced since).

Yes! :) I have fixed quite some more beforehand as well

PR #1843, opened 2023-07-04
Fix well known operation types only

lu-zero:
The assertion contains two conflicting statements, keep only the one
that matches the concept expressed few lines above:

```
    The list of possible operation types of a form is fixed.
```

***
<a href="https://pr-preview.s3.amazonaws.com/luminem/wot-thing-description/pull/1843.html" title="Last updated on Jul 12, 2023, 3:20 PM UTC (2f6ced0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1843/a09239f...luminem:2f6ced0.html" title="Last updated on Jul 12, 2023, 3:20 PM UTC (2f6ced0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2298,8 +2298,7 @@ is provided in Appendix <a href="#json-schema-for-validation"></a>.
           Future versions of the standard may extend this list but
           <span class="rfc2119-assertion" id=
           "well-known-operation-types-only">operations types
-          SHOULD NOT be
-          arbitrarily set by servients and be restricted to the values in the
+          MUST be restricted to the values in the
           table below.</span></p>
 
         <div id="table-operation-types">
--- a/validation/td-validation.ttl
+++ b/validation/td-validation.ttl
@@ -762,8 +762,7 @@
           Future versions of the standard may extend this list but
           <span class="rfc2119-assertion" id=
           "well-known-operation-types-only">operations types
-          SHOULD NOT be
-          arbitrarily set by servients and be restricted to the values in the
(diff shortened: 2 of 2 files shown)


relu91 (approved):
I think the change is right, I can't think of reasons why keeping the assertion as it is. Just note that this is not editorial, even if I consider it as a bug fix.

egekorkan:
Sadly we cannot change any assertions at this moment other than typos etc. I think that there is no conflicting statement about this. The list of `op` values are fixed

lu-zero:
At least one person was unsure on how to interpret the `SHALL NOT` and asked me about it 

We stated above and in the second sentence in the assertion that you cannot make up additional operations, but yet we use the relaxed bound in the first statement.

We can defer to 2.0 the clarification, but...

egekorkan:
We will discuss this in tomorrow's meeting. 

The technical change is fine since we are already restrictive about op values in our schema (see [here](https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json#L381)). This can be considered a bug fix but we need to understand if this would be flagged as a normative change. There is no known use case to allow flexibility in the value of `op`. Before this PR, I did not realize that MUST NOT and SHALL NOT are different but with the current wording we allow flexibility.

sebastiankb (approved):

ashimura:
have got approval from @plehegar , so let's merge this :)

egekorkan:
Call of 20.09: We can merge it

PR #1852, opened 2023-07-12
Update README with rendering instructions

egekorkan:
No changes to the spec text


egekorkan:
Call of 12.07: The group is ok to merge since it is just instructions and not specification text

PR #1856, opened 2023-07-19
Start documenting NAMESPACES

### LINKED ISSUE #1855: Documenting Namespace Related Information
The need to main a couple of namespaces for different purposes. This also implies the resources behind them need to be maintained and how they are linked to each too. So far, this seems to have happened in a not-so-transparent fashion where I cannot find which file in our repo a URL like `https://www.w3.org/2022/wot/td/v1.1` points to. In general, we should document these since we now need URLs for TD and TM JSON Schemas, TM Ontology (see https://github.com/w3c/wot-thing-description/issues/1807) meanwhile linking to correct W3C guidelines such as https://www.w3.org/2005/07/13-nsuri to understand what URLs we should create.

egekorkan:
fixes #1855

### DIFF
--- a/NAMESPACES.md
+++ b/NAMESPACES.md
@@ -18,7 +18,8 @@ For choosing new namespaces, please refer to <https://www.w3.org/2005/07/13-nsur
 | `https://www.w3.org/2019/wot/security` (same as above) | `http://w3c.github.io/wot-thing-description/ontology/wotsec.html` | `text/html`|
 | `https://www.w3.org/2019/wot/hypermedia` | `http://w3c.github.io/wot-thing-description/ontology/hypermedia.ttl` | `text/turtle`
 | `https://www.w3.org/2019/wot/hypermedia` (same as above)| `http://w3c.github.io/wot-thing-description/ontology/hyperm.html` | `text/html` |
-| `https://www.w3.org/2022/wot/td/v1.1` | TBD | `application/ld+json`|
+| `https://www.w3.org/2022/wot/td/v1.1` | `https://w3c.github.io/wot-thing-description/context/td-context-1.1.jsonld` | `application/ld+json`|
+| `https://www.w3.org/2022/wot/tm` | `w3c.github.io/wot-thing-description/ontology/tm.ttl` | `text/turtle`|
 
 ## JSON Schemas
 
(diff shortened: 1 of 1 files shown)


egekorkan:
Regarding URLs for JSON Schemas:
Call of 19.07: 
- @relu91 we should not use content negotiation since it is not a different representation of the same content. It would be better to keep the same URL design with the years to have consistency. For TD 2.0 we can change to `/ns` (@egekorkan and @ashimura agree).
- We should have resolution on how to handle URIs such as: "We agree to use URIs with date for TD 1.1 and 1.0. For TD 2.0 we will reevaluate and possibly change to using /ns approach"
- Concrete URI Mapping is being propose below as brainstorming:

Resources to have new URIs for:
- TM JSON Schema: https://github.com/w3c/wot-thing-description/blob/TD11-PR/validation/tm-json-schema-validation.json
- TD JSON Schema https://github.com/w3c/wot-thing-description/blob/TD11-PR/validation/td-json-schema-validation.json

`https://www.w3.org/2022/wot/td/v1.1` and `https://www.w3.org/2022/wot/tm` are URIs to the context files (`https://www.w3.org/2019/wot/td/v1 for 1.0)

- `https://www.w3.org/2022/wot/td-schema/v1.1`
- `https://www.w3.org/2022/wot/tm-schema/v1.1` tm schema (obviously 1.1)
- `https://www.w3.org/2019/wot/td-schema/v1`

For TD 2.0, we can do something like this:
- `https://www.w3.org/ns/wot/td-schema` and `https://www.w3.org/ns/wot/td-context`

Note: For the JSON Schemas, since we are referring to the tagged version which has a `$id`, we need to change that too.

egekorkan:
I have added the TTL endpoint for TM and also the JSON Schemas however there is a rather important problem to solve. When we are pointing to github.io links, which are synced to the master branch, the media type is automatically provided. When we want to tag a branch in GitHub, the links we can provide are raw git content (e.g. `https://raw.githubusercontent.com/w3c/wot-thing-description/TD11-PR/validation/td-json-schema-validation.json`). In that case, GitHub explicitly does not provide a media type. Some solutions that come to mind:
- Put the relevant content under the publication folder, which includes adding new documents to https://github.com/w3c/wot-thing-description/tree/TD11-PR/publication/rec (TD 1.0 REC). I am not sure if this is allowed at this point but since there will be zero changes to the spec text, it might be fine. We need to of course do this for all resources in TD 1.0 and TD 1.1. This will result in URLs like `https://w3c.github.io/wot-thing-description/publication/rec/index.html`
- Generate URLs via a service like statically. Example: `https://cdn.statically.io/gh/w3c/wot-thing-description/TD11-PR/validation/tm-json-schema-validation.json`
- For TD 1.1 REC, generate a folder with all that resources (like publication) and revert all the documents to 1.0 stage. I do not like this option since that means that we cannot touch those document for all upcoming TD versions.

Ignoring the content type and serving the content should not be an option since (semantic web) tools depend on it afaik. My vote would be to go for option 1.

egekorkan:
Call of 26.07: We will merge this and work on the document to understand how to proceed.

PR #1858, opened 2023-07-25
Remove mention of atrisk section

egekorkan:
this was an oversight due to some sparql rules not getting updated. Somehow it was not part of the atrisk.css :/


sebastiankb (approved):

JKRhb:
As it seems, the bug is caused by an at-risk note that is still present in the .ttl file:

https://github.com/w3c/wot-thing-description/blob/7bf1c533c1ffcb4da7aeb1b55aa1b6e2f4855dc0/validation/td-validation.ttl#L1525

I guess it would probably make sense to remove that one instead/additionally?

egekorkan:
> As it seems, the bug is caused by an at-risk note that is still present in the .ttl file:
> 
> https://github.com/w3c/wot-thing-description/blob/7bf1c533c1ffcb4da7aeb1b55aa1b6e2f4855dc0/validation/td-validation.ttl#L1525
> 
> I guess it would probably make sense to remove that one instead/additionally?

Ah yes that is the real source of the problem! I wanted to do it asap yesterday and maybe went for a dirtier fix ^^

egekorkan:
Done. It also had 0 changes on the index.html

egekorkan:
TD Call of 26.07: The TF agrees that this PR can be merged after getting approval from PLH.

chrisn (approved):
Looks good, thank you!

ashimura:
Thanks a lot, @chrisn !

ashimura:
This PR has been approved by @plehegar , so let's merge this :)

egekorkan:
Call of 20.09: Merging

PR #1861, opened 2023-07-31
Desired namespaces for 1.0 and 1.1 resources

egekorkan:
This assumes that we:
- Add resources to folder at 1.0 rec publication: `https://github.com/w3c/wot-thing-description/tree/main/publication/rec`
- Create a new folder at 1.1 rec publication: `https://github.com/w3c/wot-thing-description/tree/main/publication/rec11`

### DIFF
--- a/NAMESPACES.md
+++ b/NAMESPACES.md
@@ -1,4 +1,4 @@
-# Namespaces and Resources
+# [Draft] Namespaces and Resources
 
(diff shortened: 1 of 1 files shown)


egekorkan:
Call of 02.08:
- It would be better to not have such long URLs
- It would be better to arrange our GitHub actions such that a tag results in a separate gh pages deployment.
- This assumes usage of the repository as a long-term storage, it should have only live files. Tagged branches or mirroring is better.

PR #1862, opened 2023-07-31
Add tentative rec10 resources

egekorkan:
This adds the JSON Schema and the TD context that are taken from https://github.com/w3c/wot-thing-description/blob/REC1.0


egekorkan:
Call of 02.08:
- The group is ok to merge but we will look at the big picture
- 2.0 work should organize this kind of content much better
- Together with the other PRs, we should explain the overall mechanism

egekorkan:
Call of 27.09: Merging. This folder is only tentatively used until they are moved to a stable folder at https://github.com/w3c/wot-resources

PR #1863, opened 2023-07-31
Add rec11 resources

egekorkan:
This adds the following to the publication/ver11/7-rec folder
- TD and TM JSON Schemas
- TM Ontology
- TD Context File

### DIFF
--- a/publication/ver11/7-rec/context/td-context-1.1.jsonld
+++ b/publication/ver11/7-rec/context/td-context-1.1.jsonld
@@ -1,0 +1,1403 @@
+{
+  "@context": {
+    "td": "https://www.w3.org/2019/wot/td#",
+    "jsonschema": "https://www.w3.org/2019/wot/json-schema#",
+    "wotsec": "https://www.w3.org/2019/wot/security#",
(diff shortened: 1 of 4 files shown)


danielpeintner:
> Regarding JSON Schemas: Should we remove the date from the version field? What do you think @relu91 ?

FYI: In the scripting API we used the date in the version field to create according TS definititions and label them accordingly.
see https://www.npmjs.com/package/wot-thing-description-types and https://www.npmjs.com/package/wot-thing-model-types

egekorkan:
Call of 23.08: The publication folder is only a tentative area and it should not be used for this purpose. A separate repo is desired whose name is not important but can be github.com/w3c/wot-resources . It will not be visible to anyone since we will redirect anyways. It can be used by multiple WoT specs and it should have a more restricted access than the usual wot repositories. Chairs and Team Contact can only merge PRs and no one can make a commit to main branch.

egekorkan:
Call of 20.09: The WG does not have a consensus on where to put these resources. Keeping it open until then.

egekorkan:
Call of 27.09: Merging. They will be moved to wot-resources folder later on

PR #1866, opened 2023-08-02
TD html ontology fix 

### LINKED ISSUE #1860: Publish `base` as html rendered representation?
*** **Please note the following comment has no impact on the TD specification itself** ***
 
Terms have a unique IRI. E.g., `observable` has the IRI https://www.w3.org/2019/wot/td#isObservable
When you click on this link, the browser shows the specific meaning of this term.

This seems to work well but it seems there is no IRI html representation for the term `base`. Based on the [TD ontology ](https://github.com/w3c/wot-thing-description/blob/c1f925b4b1b3318ed0b6c51d661460127b4c2ea2/ontology/td.ttl#L37) it should be https://www.w3.org/2019/wot/td#baseURI

It seems just a mistake and we should check the render script why the [html version](https://github.com/w3c/wot-thing-description/blob/main/ontology/td.html) `base` term is not onboard.

sebastiankb:
fixes #1860


sebastiankb:
@mahdanoura

mahdanoura:
I checked the ontology, the baseURI is defined as an AnnotationProperty, which means an rdfs:domain can't be assigned to it.

wiresio:
Maybe @ethieblin can also have a look.

ethieblin:
To be sure, you manually modified the td.html file ?
Because if I am not mistaken, all the <prefix>.html files are automatically generated using the render.sh script at the route of the td repository.

It is ok for me if we have a guarantee that this script will not be used in the future, otherwise your changes may be lost.

relu91:
> To be sure, you manually modified the td.html file ?
Because if I am not mistaken, all the .html files are automatically generated using the render.sh script at the route of the td repository.

Yes, that is correct. In my understanding, `td.ttl` should have been modified. However, the baseURI property is correctly declared [there](https://github.com/w3c/wot-thing-description/blob/main/ontology/td.ttl#L37). There must be a problem in the render script. Let me double-check.

mahdanoura:
I checked the `templates.sparql` file in the render script, it is listed to render classes, Object Properties and Data Properties, so if you also want the Annotation properties to be rendered it has to be modified there.

relu91:
> I checked the ontology, the baseURI is defined as an owl:AnnotationProperty, which means an rdfs:domain can't be assigned to it.

Yes, exactly. Maybe a proper solution to this issue is to add a section with all the `owl:AnnotationProperties` ? we have buch of those that are not really rendered:

```ttl
:supportContact rdf:type owl:AnnotationProperty ;
    rdfs:label "supportContact" ;
    rdfs:comment "Provides information about the TD maintainer as URI scheme (e.g., mailto [[RFC6068]], tel [[RFC3966]], https[[RFC9112]])."@en ;
    rdfs:seeAlso schema:contactPoint ;
    rdfs:isDefinedBy <https://www.w3.org/2019/wot/td> .

:followsProfile rdf:type owl:AnnotationProperty ;
    rdfs:label "followsProfile" ;
    rdfs:comment "Indicates the WoT Profile mechanisms followed by this Thing Description and the corresponding Thing implementation."@en ;
    rdfs:isDefinedBy <https://www.w3.org/2019/wot/td> .

:baseURI rdf:type owl:AnnotationProperty ;
    rdfs:label "baseURI" ;
    rdfs:comment "Define the base URI that is used for all relative URI references throughout a TD document. In TD instances, all relative URIs are resolved relative to the base URI using the algorithm defined in [RFC3986]. base does not affect the URIs used in @context and the IRIs used within Linked Data [LINKED-DATA] graphs that are relevant when semantic processing is applied to TD instances."@en ;
    rdfs:isDefinedBy <https://www.w3.org/2019/wot/td> .

:versionInfo rdf:type owl:AnnotationProperty ;
    rdfs:label "versionInfo" ;
    rdfs:comment "Provides version information."@en ;
    rdfs:seeAlso schema:version ;
    rdfs:isDefinedBy <https://www.w3.org/2019/wot/td> .

:instance rdf:type owl:AnnotationProperty ;
    rdfs:label "instance" ;
    rdfs:comment "Provides a version identicator of this TD instance."@en ;
    rdfs:isDefinedBy <https://www.w3.org/2019/wot/td> .

:model rdf:type owl:AnnotationProperty ;
    rdfs:label "model" ;
    rdfs:comment "Provides a version indicator of the underlying TM."@en ;
    rdfs:isDefinedBy <https://www.w3.org/2019/wot/td> .
```

And maybe another question, are those really just `owl:AnnotationProperty`? At least for baseURI, my gut feeling tells me that it could also be seen as an ObjectProperty.

relu91:
> I checked the `templates.sparql` file in the render script, it is listed to render classes, Object Properties and Data Properties, so if you also want the Annotation properties to be rendered it has to be modified there.

yes, same time :) check my comment above.

mahdanoura:
> Yes, exactly. Maybe a proper solution to this issue is to add a section with all the `owl:AnnotationProperties` ? we have buch of those that are not really rendered:
> 
I think rendering those properties normally doesn't make too much sense as they really serve as metadata, but in this case as it includes important information such as the baseURI, instance, and model references throughout the TD, and TM document could make sense. 

> And maybe another question, are those really just `owl:AnnotationProperty`? At least for baseURI, my gut feeling tells me that it could also be seen as an Datatype Property.

I was surprised to see the baseURI considered as an AnnotationProperty, in my opinion it makes more sense to consider it as a Datatype Property.

relu91:
> I was surprised to see the baseURI considered as an AnnotationProperty, in my opinion it makes more sense to consider it as a Datatype Property.

Ok, so I would say that a proper solution for #1860 is to change it to Datatype Property. Then we can discuss later about the others (I would say that we can translate all to Datatype Properties, but I want to know more about the reasons why they were designed as Annotation properties).

relu91 (changes requested):
See the comment above.

sebastiankb:
> To be sure, you manually modified the td.html file ?

yes, just as a quick solution. If we can solve this with the render script this would a better solution.

egekorkan:
Call of 20.09:
- @relu91 will do the correct solution by changing this PR but make sure that we do not change the index.html. He will also change it for the other missing keywords as shown in https://github.com/w3c/wot-thing-description/pull/1866#issuecomment-1661871586

egekorkan:
#1881 makes this unnecessary

PR #1867, opened 2023-08-08
Documenting TD resources next to the spec

egekorkan:
This documents the additional resources next to the TD spec that we maintain. Next commits will detail the usage of those documents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,9 +133,33 @@ items are listed in `testing/inputs/atrisk.csv`.  If at-risk items are
 updated, to update the at-risk highlighting the implementation report
 needs to be generated first, and then the rendering.
 
-
 ## Resources used together with the TD Specification
 
-- Ontology files
-- Context files
-- JSON Schemas
+In addition to the TD specification, the Working Group maintains additional resources that are relevant for implementers.
+These are categorized below and their use cases explained.
+
+### Vocabularies and Ontologies
+
+The present specification introduces the TD Information Model as a set of constraints over different vocabularies.
+The current list of vocabularies are found under [the namespaces section](https://w3c.github.io/wot-thing-description/#namespaces).
+
+By default, if a user agent does not perform any content negotiation, a human-readable HTML documentation is returned instead of the RDF document. 
+To negotiate content, clients must include the HTTP header `Accept: text/turtle` in their request.
+
+The working documents are found under the [ontology](./ontology) folder.
+
+### Context files
+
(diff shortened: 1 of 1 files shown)


egekorkan:
@mahdanoura and @relu91 could you have a look and possibly extend the uses of ontologies and context files?

relu91 (approved):
I had a look at the changes and I think that overall it is a good description of the current resources managed by us. One thing that we can mention is that in the past we use to have also these alignment ontologies: [schema](https://github.com/w3c/wot-thing-description/blob/main/ontology/td-sdo.ttl) and [sosa](https://github.com/w3c/wot-thing-description/blob/main/ontology/td-sosa.ttl). But I am not even sure if they are up to date with the current status.

README.md
```diff
@@ -132,3 +132,34 @@ that highlights at-risk items in the generated `index.html`.  The at-risk
...
+
+By default, if a user agent does not perform any content negotiation, a human-readable HTML documentation is returned instead of the RDF document. 
+To negotiate content, clients must include the HTTP header `Accept: text/turtle` in their request.
+
+The working documents are found under the [ontology](./ontology) folder.
+
+### Context files
+
+In order to have reference to the TD ontology within JSON-LD documents, including TDs, we provide a JSON-LD context file.
+The working document can be found at [context/td-context-1.1.jsonld](./context/td-context-1.1.jsonld) and the version-specific context file can be retrieved from the `@context` value of a TD 1.0 or TD 1.1 instance.
```
relu91, line 154: I think we can and a simple sentence to explain the context file is actually a merge of the other files contained in the folder. 
```suggestion
The working document can be found at [context/td-context-1.1.jsonld](./context/td-context-1.1.jsonld) and the version-specific context file can be retrieved from the `@context` value of a TD 1.0 or TD 1.1 instance. **Note** that  [context/td-context-1.1.jsonld](./context/td-context-1.1.jsonld) is the result of the rending process and it should not be edited manually. In practice, the rending process merges [context/td-context.jsonld](./context/td-context.jsonld), [context/json-schema-context.jsonld](./context/json-schema-context.jsonld), [context/wot-security-context.jsonld](./context/wot-security-context.jsonld), [context/hypermedia-context.jsonld](./context/hypermedia-context.jsonld).  
```

relu91 (changes requested):
On second thought I think we can add these two lines.

egekorkan:
@relu91 I have applied your suggestion but I had to fix the typo of `rending` to `rendering`

relu91 (approved):

PR #1870, opened 2023-08-08
Fix schema links to static ones

egekorkan:
We should fix these notes. My proposal is to merge this and https://github.com/w3c/wot-thing-description/pull/1861 together and decide on where to serve the content from. It doesn't matter if the schemas come from a new repo, a GitHub release or the publication folder but these links should appear in the REC.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1870.html" title="Last updated on Aug 8, 2023, 10:43 PM UTC (6e88bb5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1870/c1f925b...6e88bb5.html" title="Last updated on Aug 8, 2023, 10:43 PM UTC (6e88bb5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -466,6 +466,7 @@ a[href].internalDFN {
         specification and do not have at least two implementations at that
         time will either be removed or converted into informative
         statements, as appropriate.</p>
+
   </section>
 
   <section id="introduction" class="informative">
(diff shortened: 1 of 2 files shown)


relu91 (changes requested):
Don't we to update the `$id` [here](https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json#L6) too? 

**Update**: ah probably is it something that we have to change when we decide on the mechanism in #1861

sebastiankb (approved):

egekorkan:
> Don't we to update the `$id` [here](https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json#L6) too?
> 
> **Update**: ah probably is it something that we have to change when we decide on the mechanism in #1861

So I would try to keep the PRs focused for now. This one changes the index.html and once we agree on the link, we can change the resource behind it easily (where the `$id` will be present)

egekorkan:
Merging as communicated at https://lists.w3.org/Archives/Member/member-wot-wg/2023Aug/0003.html

PR #1890, opened 2023-10-02
Rec11 Prep v2

egekorkan:
Still in draft stage but NOT choosing draft PR to generate preview links.

Using this statically link for checks which needs to be updated for each commit:
https://cdn.statically.io/gh/w3c/wot-thing-description/ededdb49b596e402fc70a39709cef603b3a3178e/publication/ver11/7-rec/Overview.html

[Link Checker Rules](https://validator.w3.org/checklink?uri=https%3A%2F%2Fcdn.statically.io%2Fgh%2Fw3c%2Fwot-thing-description%2Fededdb49b596e402fc70a39709cef603b3a3178e%2Fpublication%2Fver11%2F7-rec%2FOverview.html&hide_type=all&depth=&check=Check): We need to fix the resources repository and redirect URLs, otherwise these URLs will always show errors
[Pubrules](https://www.w3.org/pubrules/?url=https%3A%2F%2Fcdn.statically.io%2Fgh%2Fw3c%2Fwot-thing-description%2Fededdb49b596e402fc70a39709cef603b3a3178e%2Fpublication%2Fver11%2F7-rec%2FOverview.html&profile=REC-Echidna&validation=simple-validation&informativeOnly=false&echidnaReady=true&patentPolicy=pp2020): Other than "do link, HTML and CSS checking, it says "If it is the intention to incorporate new features in future updates of the Recommendation, please make sure to identify the document as intending to allow new features." and I am not sure about what to say here. We are not planning or even able to add features to RECs.
[HTML Check](https://validator.w3.org/nu/?doc=https%3A%2F%2Fcdn.statically.io%2Fgh%2Fw3c%2Fwot-thing-description%2Fededdb49b596e402fc70a39709cef603b3a3178e%2Fpublication%2Fver11%2F7-rec%2FOverview.html): Fine
[CSS Checks](https://jigsaw.w3.org/css-validator/validator?uri=https%3A%2F%2Fcdn.statically.io%2Fgh%2Fw3c%2Fwot-thing-description%2Fededdb49b596e402fc70a39709cef603b3a3178e%2Fpublication%2Fver11%2F7-rec%2FOverview.html&profile=css3svg&usermedium=all&warning=1&vextwarning=&lang=en): Only a warning about the W3C logo

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1890.html" title="Last updated on Oct 4, 2023, 3:13 PM UTC (cb6bc67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1890/b34f28a...cb6bc67.html" title="Last updated on Oct 4, 2023, 3:13 PM UTC (cb6bc67)">Diff</a>


index.html
```diff
@@ -59,7 +59,7 @@
                 }, {
                     name: "Matthias Kovatsch",
                     company:    "Huawei",
-                    companyURL: "https://www.huawei.com"
+                    companyURL: "https://www.huawei.com/en/?ic_medium=direct&ic_source=surlent"
```
danielpeintner, line 62: I am wondering why we need to add this redirect/updated URLs ?

index.template.html
```diff
@@ -59,7 +59,7 @@
                 }, {
                     name: "Matthias Kovatsch",
                     company:    "Huawei",
```
egekorkan, line 61: We should add something like "Until DATETIME" to indicate when he stopped working there. Similar for Victor Charpenay.

egekorkan:
Call of 04.10: 
- There will be a follow up PR to fix affiliation of Matthias Kovatsch and Victor Charpenay
- We will not add `<p>Future updates to this Recommendation may incorporate <a href="https://www.w3.org/2023/Process-20230612/#allow-new-features">new features</a>.</p>` since that implies doing an amendment to the specification to add a feature. As agreed in our charter, we will start working on TD 2.0.

index.html
danielpeintner: I see but I think it is not useful to have such a change

egekorkan: In the end, this specific one is removed since we just put "while at Huawei". Do you think we should not change any redirected URL?

danielpeintner: In those case I think we should keep the URLs as they were.. even adding "/en" is wrong since people from China will also see the English version even though they should be forwarded to the Chinese one..
The same applies to the Intel link.

mmccool: I think we want the simple URLs.  The more complicated one is more likely to change and then be broken...

PR #1892, opened 2023-10-04
Move rec 1.0 and 1.1 resources to separate folders

egekorkan:
As discussed in main call, these resources should not be in the same folder with the publication HTML files


egekorkan:
TD Call 04.10: We can merge, no objections.

PR #1899, opened 2023-10-18
Fix "Intel" usage to "Intel Corp."

egekorkan:
Already agreed in main call


egekorkan:
TD Call of 18.10: We agreed to merge this simple editorial fix

Issue #1898, opened 2023-10-18
Unnecessary triple quotes around comments


mahdanoura:
The last half of the `ontology.ttl `file has triple quotes for some of the classes in the comments, e.g.

`rdfs:comment """Operation type of forms used to read a property value"""@en ;`

PR #1901, opened 2023-10-18
unnecessary triple quotes around comments

mahdanoura:
#1898

### DIFF
--- a/ontology/td.ttl
+++ b/ontology/td.ttl
@@ -265,90 +265,90 @@
 
 :readProperty rdf:type :OperationType ;
     rdfs:label "readProperty" ;
-    rdfs:comment """Operation type of forms used to read a property value"""@en ;
+    rdfs:comment "Operation type of forms used to read a property value"@en ;
     rdfs:isDefinedBy <https://www.w3.org/2019/wot/td> .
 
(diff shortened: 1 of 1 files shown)


danielpeintner:
Out of curiosity I _checked_ when to use triple quotes and when not to do so. Apparently one has to use triple quotes for "newlines"... maybe also in other cases. Having said that, it is not wrong per se to have them _always_. It might be even safer to do so... 

Anyhow, no objection to merge as is.

mahdanoura:
> Out of curiosity I _checked_ when to use triple quotes and when not to do so. Apparently one has to use triple quotes for "newlines"... maybe also in other cases. Having said that, it is not wrong per se to have them _always_. It might be even safer to do so...
> 
> Anyhow, no objection to merge as is.

It's not incorrect to do so, but as I write they are unncessary. It also causes the usual SyntaxError: invalid syntax, when trying to parse the ontology. I was trying a SCHACL library and came across this.

egekorkan:
Call of 18.10: This does not result in any changes in the index.html (when run locally). The TF is fine to merge and then @egekorkan will copy over to the wot-resources

PR #1903, opened 2023-10-18
Delete NAMESPACES.md

egekorkan:
The file content is moved to wot-resources with PR https://github.com/w3c/wot-resources/pull/6 . Once that PR is merged, we can merge this one as well


egekorkan:
Call of 25.10: Merging, also agreed on last week.

PR #1900, opened 2023-10-18
Modeling string or array of string in the TTL files

mahdanoura:
#1884

PR #1904, opened 2023-10-19
Copy over last changes to ontology documents

egekorkan:
#1900 and #1901 have changed these resources that we need to copy over. Also mirroring at https://github.com/w3c/wot-resources/pull/8


egekorkan:
Call of 25.10: Merging since no objections

PR #1909, opened 2023-10-25
Inconsistencies in wotsec ontology

### LINKED ISSUE #1908: Inconsistencies in wotsec ontology
The PoPSecurityScheme class in the wotsec.ttl is commented out, but still used for some DataProperties. This makes the figure in https://w3c.github.io/wot-thing-description/#td-security-model show wrong information.

mahdanoura:
fixes #1908


egekorkan:
I have tried it locally and there are some things to note:
- The figure doesn't change since we are using static pngs at the moment.
- There is a change in the HTML version.

egekorkan:
Call of 25.10: We will merge this PR. The rest of the figure related issues will be done later.

PR #1911, opened 2023-10-25
Figure Update Process Documentation

egekorkan:
This simply adds this step since it was not clear that there is this task to be done before publication.

### DIFF
--- a/publication/README.md
+++ b/publication/README.md
@@ -43,4 +43,12 @@ checkers can use.
 
 ## Manual Link Corrections
 
--  http://www.etsi.org/deliver/etsi_ts/103200_103299/103264/02.01.01_60/ts_103264v020101p.pdf should be changed to https://www.etsi.org/deliver/etsi_ts/103200_103299/103264/02.01.01_60/ts_103264v020101p.pdf  (making it https. It is wrong in specref but will be changed hopefully by https://github.com/tobie/specref/pull/757)
+Some redirects come from [specref](https://www.specref.org/) not being up to date. In those cases, you need to manually update the final Overview.html.
+
+-  None
+
+## Converting Automatically Generated SVGs to PNGs
+
+All SVG files under [visualization](../visualization) are generated automatically when running the `npm run render` command.
+Before publication, PNGs version of the figures should be generated.
+A tool like [Inkscape](https://inkscape.org/) can be used to do that.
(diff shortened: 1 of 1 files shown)


publication/README.md
```diff
@@ -43,4 +43,12 @@ checkers can use.
...
--  http://www.etsi.org/deliver/etsi_ts/103200_103299/103264/02.01.01_60/ts_103264v020101p.pdf should be changed to https://www.etsi.org/deliver/etsi_ts/103200_103299/103264/02.01.01_60/ts_103264v020101p.pdf  (making it https. It is wrong in specref but will be changed hopefully by https://github.com/tobie/specref/pull/757)
+Some redirects come from [specref](https://www.specref.org/) not being up to date. In those cases, you need to manually update the final Overview.html.
+
+-  None
+
+## Converting Automatically Generated SVGs to PNGs
+
+All SVG files under [visualization](../visualization) are generated automatically when running the `npm run render` command.
+Before publication, PNGs version of the figures should be generated.
+A tool like [Inkscape](https://inkscape.org/) can be used to do that.
```
danielpeintner, line 54: Does it make sense to mention _why_ we need the additional step to "manually" generate the PNGs?
Like
* organize the overall structure (better)
* add/remove parts
* ???

egekorkan: @danielpeintner what do you think now?

danielpeintner: Prose looks good now.

In an ideal world we should just need to reorder... but I do recall there is more happening. @mahdanoura didn't you mention something yesterday?

Just by looking at 
* https://w3c.github.io/wot-thing-description/visualization/td.png
  vs.
* https://w3c.github.io/wot-thing-description/visualization/td.svg

I detected some differences already:
* ActionAffordance in PNG misses synchronous
* in PNG there are ~5 pointers to DataSchema, in SVG many more (maybe just combined?)
* in PNG there are ~2 pointers to MultiLanguage, in SVG more (maybe just combined?)
* InteractionAffordance class has title and description. In SVG the type is of "langString" instead of "string"
* type langString in Thing class also..
* ... maybe there is more 🙈

mahdanoura: exactly what I was trying to point out yesterday. Those differences are because the shacl shape types are not consistent with the ontology dataproperty types, in the ontology they are defined with e.g., xsd:string and suddenly in the shacl langString. This shows the necessity of the change that is required in the toolchain, different people are designing the ontologies and maintaining the shacl shapes differently.

egekorkan: The figures are fixed in #1913 . The toolchain issue will be revisited in TD 2.0. Resolving

egekorkan:
Call of 02.11: @lu-zero and @mjkoster are fine with this documentation. It is good to give directives on this.

PR #1912, opened 2023-10-25
TM Regex fix

### LINKED ISSUE #1906: TM Schema regex error
In our TM Schema, we have a regex at https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json#L1852 . This is sadly failing in golang and rust and subsequently the TM Schema cannot be used by many golang based JSON Schema validators (not sure about Rust yet). An alternative is available at https://regex101.com/r/qSBNcw/1 but that fails for Rust. Ideally we should fix this asap.

Note: `/` is not reserved character for regex but we are escaping it but I do not remember why

egekorkan:
fixes #1906 (discussion is there)

### DIFF
--- a/validation/tm-json-schema-validation.json
+++ b/validation/tm-json-schema-validation.json
@@ -1849,8 +1849,8 @@
           {
             "not": {
               "type": "string",
-              "pattern": "(/)(.*\\1){2}",
-              "$comment": "regex tests available at https://regex101.com/r/Ytzd72/1"
+              "pattern": "(/)(.*/){2}",
+              "$comment": "regex tests available at https://regex101.com/r/r7vB0r/2"
             }
           }
         ]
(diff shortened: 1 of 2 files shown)


mmccool:
Is this update for TD 2.0 (next version) or is the intention here to update the published resources for TD 1.1?  If the latter, this should be made clear and we probably need a resolution in the main call, etc.

Personally... without a clear call for resolution in the main call, I think this should NOT go into the resources for TD 1.1.

egekorkan:
Call of 02.11: We have removed the formatting change since it was added by mistake (done by my IDE). The change is only to the regex so that more regex libraries support it. Thus, there should be no breaking changes to implementations. We can have a main call resolution to merge it. Once merged, it will be copied over to wot-resources.

egekorkan:
We have agreed to merge it in the main call.

PR #1913, opened 2023-10-31
TD Figures Update

egekorkan:
There are some misalignments in the figures that are autogenerated and exported into png. I have fixed them in SVGs manually given the publication procedure but the toolchain should be fixed later on.

What I have done to generate the results:

- Run npm render and regenerate the SVGs. Then push them to GitHub
- Copyover those SVG files to publication folder
- Manually adapt the SVG files so that errors are fixed and the organization of boxes and arrows is nicer. Due to a toolchain issue, there were missing terms in SVG (and thus PNGs). Push these SVG files to GitHub
- Export PNGs from the previous SVG files. These overwrite the current PNGs

Status:
- Thing: Done
- JSON Schema: Done
- Security: DONE
- Forms: Done


sebastiankb (approved):

JKRhb (commented):
Thank you, @egekorkan! Just two tiny nits: It seems to me as if sometimes a space is missing between `(Array of)` and the respective type. Also, sometimes, the `Array of` in parentheses is italicized, and sometimes not.

sebastiankb:
This PR just updates the Figures so that they are in sync with Chapter 5.3. Approve this PR

egekorkan:
@JKRhb I have addressed your feedback. If it can be an array or string (e.g. `@type`), it is italic (also `(Array of)`). If it is always an array (e.g. `required`), then it not italic (also `Array of`)

I will also beautify the Thing diagram to remove weird placement of arrows.

egekorkan:
The Thing diagram is also updated. For preview, you can use the following URL: https://cdn.statically.io/gh/w3c/wot-thing-description/ege-image-update/publication/ver11/7-rec/Overview.html#namespaces

mmccool:
Met during main call, felt there were still some things to fix and we wanted a summary of the changes (we looked through the figures but it was hard to see exactly what changed due to the changes in layout).  So we decided to make a resolution in the main call to give the TD editors the authority to make the necessary changes to the figures prior to REC transition.  Ideally you would finalize this tomorrow in the TD coordination call.

I would be good to write a summary of the changes somewhere, however, e.g. "string changed to langString in title and description", "href added to Link and Form", etc.

As for reviews, I noticed in some places that "any type" is not formatted in black but I think it should be.   Also I don't think "subclass" arrows should go downward as a general rule, but I can live with the new layout.  However, as above, the group is giving you the authority to finalize this and you should do it as soon as possible.

mmccool (approved):
would have been good to see a summary of changes, but ok to publish, some minor points noted elsewhere

egekorkan:
The changes were:

## Thing
- id is mandatory --> should be optional
- PropertyAffordance is not inheriting DataSchema 
- @type missing from InteractionAffordance

## Data Schema
- const, enum, default missing

## Security
- proxy as anyURI is in SVG, but in PNG is missing
- scheme is missing

## Form
- op is missing

lu-zero (approved):

egekorkan:
Call of 02.11: @lu-zero and @mjkoster are fine with the changes. We have also updated the changes in the comment above. There is no preference regarding the placement of the parent class, general preference is to have less cluttered diagrams. We will merge it.

PR #1919, opened 2023-11-15
WoT toolchain diagram added

mahdanoura:
BPMN diagram showing the current process required for generating the WoT specifications. Links to [Specification Generation Tooling](https://github.com/w3c/wot/blob/main/planning/ThingDescription/work-items.md#refactoring-roadmap)


egekorkan (commented):
The diagram itself is fine and much appreciated. Some remarks:
- Can you push the source file as well?
- Can you add legend to what the small icons mean?

danielpeintner (commented):
I think it should be put in a dedicated folder .. maybe with a README along with it 

`images/` seems not to be the right place

egekorkan:
Call of 15.11:
- This diagram but actually our process should become simpler
- We can have some snippets of the big diagram as separate figures and explain them on their own. The big figure can be used as a reference.
- @lu-zero : It is better to spend time to simplify the process, rather than optimizing the figure.
- @egekorkan : There is also the step to generate the TM Schema
- Noting that black arrows are outputs and are mandatory for the publication process, either as the spec and its PNGs or as part of the wot-resources
- 1-2 sentences per step (in another document) would be nice
- Another figure would be nice to document what happens to the index.html with each publication, currently documented at https://github.com/w3c/wot-thing-description/blob/main/publication/README.md

egekorkan:
Before merging:
- the source file should be added
- A legend to what the small icons should be added
- A new folder called "toolchain" should be created where we can also have the associated text/guide in form of a readme.

Once those are done, we can merge the PR asynchronously.

egekorkan:
As agreed in the call yesterday, we will merge this since the points above are addressed. For more significant improvements, I will open an issue to track it. Thanks @mahdanoura !

PR #1935, opened 2023-12-12
Empty tm documents

mahdanoura:
addresses #1934


egekorkan:
Call of 13.12:
- We are fine to make changes to the TTL file since it is very unlikely that someone started using only the TTL file starting last week. Before that, the TTL file was not served.
- Editors: The TF is fine to have @sebastiankb and @mahdanoura as the editors of this ontology file.

mahdanoura:
Some limitations of our existing toolchain based on the concepts defined in the ontology:
- Datatype properties are only shown in the rendered table in the .html file if the rang is from schema:Text vocabulary and not e.g., xsd:string
- If prefixes are added to terms e.g. tm:version (which is quite normal to do), the tool breaks (templates.sparqls needs to be probably updated to support this case)

egekorkan:
Call of 11.01: We agree to merge. @egekorkan will update #1952 with these additions

PR #1937, opened 2023-12-13
Project Management Proposal Draft

egekorkan:
This only copies information from the wiki. Once we merge this, we can populate it with the process we agree on.


egekorkan:
We will work on this document here but a more generic version will be proposed for the whole WG to adopt (if everybody agrees).

PR #1942, opened 2023-12-13
Delete test-bed directory

egekorkan:
This is here for historical reasons but it is definitely not usable and relevant anymore.


egekorkan:
@egekorkan will merge asynchronously if no other wot repository is using it.

egekorkan:
I have checked all WoT repositories and none make use of this.

PR #1943, opened 2023-12-13
Delete images directory

egekorkan:
Misleading name and not relevant anymore. We will check other repositories if they are being used before merging.


egekorkan:
@egekorkan will merge asynchronously if no other wot repository is using it.

egekorkan:
The only place these were used are at https://github.com/w3c/wot-thing-description/tree/main/publication/ver10/cr/images . They are already there so it is fine to delete. The pptx was not there but I have just pushed it.

egekorkan:
Note: I have checked all WoT repositories indeed.

PR #1944, opened 2023-12-19
More concrete project management proposal

egekorkan:
This is my simple proposal.


egekorkan:
Call of 20.12:
- We should still analyze the other W3C groups do. https://github.com/w3c/strategy/projects/2 can be used for the Use Case extraction
- We need to see if it works, it is better than nothing.

PR #1945, opened 2023-12-20
Assertion id alignment

### LINKED ISSUE #1446: Alignment of assertion ids
Currently there are some assertion ids that start with `td`, `tm` and a bunch of others that have some other prefixes. We should align them in order to have:
- better ordering in the implementation report
- possiblity of grouping them for consuming

I'll do a passover in order to generate a possible list of prefixes and comment below

egekorkan:
fixes #1446 

This may seem like a big PR but it is not really. So far, most assertion id have the "td-" prefix but not all. This makes it annoying to sort them and confuses if an implementation has assertion files for two specs.

Merging this implies doing the same alignment in the testing procedure, i.e. changing the testing tool but also updating the inputs. As @mmccool suggested in the linked issue, we should automatically update the test inputs with a script.

One discussion to have is whether we want have better grouping. Maybe even indicating whether an assertion is manual in the id is a good idea.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1945.html" title="Last updated on Dec 20, 2023, 1:44 PM UTC (77a9a70)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1945/7c0b968...77a9a70.html" title="Last updated on Dec 20, 2023, 1:44 PM UTC (77a9a70)">Diff</a>


mmccool:
I have updated the IRs to point at the actual published REC rather than the github head so this should not break such links.  But in general - we should not change ids that often if we can help it.  But a big clean-up all at once seems appropriate at this point.

mmccool:
I guess we should check if any of our other specs point at specific assertion ids... I don't *think* so.

egekorkan:
TD Call 20.12:
- We should do the same for all specs to be aligned
- Implementation Report is fine since it links to publication folder
- @egekorkan will provide the script to update the inputs for testing

mjkoster:
We should make this a WoT-wide general policy

PR #1950, opened 2024-01-09
Adding publication folder to robots.txt

egekorkan:
Since that folder is just a preliminary area, crawlers should not show results from it.


egekorkan:
Call of 11.01: We can merge as it is adding the staging area.

PR #1951, opened 2024-01-09
Versioning of TD resources

egekorkan:
It is currently at https://www.w3.org/WoT/IG/wiki/WG_WoT_Thing_Description_WebConf but it should move to an md document.


egekorkan:
@egekorkan will fix the indentation and then merge it when fixed.

Issue #1940, opened 2023-12-13
Fixing the Ontology HTML files to static versions


egekorkan:
Looking at https://www.w3.org/2019/wot/td (and others), we should not be using Respec since it updates the date based on the repository. The file should be also fixed from such errors and warnings.

Sentences such as "Validation of the document by the Working Group is expected by the end of June 2019. " should be removed. There is no implementation impact since this is just the HTML document to be read.

PR #1952, opened 2024-01-10
Static Ontology HTMLs

egekorkan:
This PR does:

- There were many errors and warnings in the HTML files of the ontologies. I fixed them by changing the template.html files. They were all trivial errors
- I added abstract and intro to wot security ontology
- I have generated static versions of each HTML version in a new folder called staticFiles under resources/ontology.

Once the PR is merged, I will move the static files to wot-resources repo.

fixes https://github.com/w3c/wot-thing-description/issues/1940


egekorkan:
Call of 10.01: We will keep it open for people to review and merge it next week if no objections are raised.

egekorkan:
In the end, I have also put the tidied versions of the HTML files, exactly like we do for publications. These documents are really like a publication. @JKRhb @mmccool I recommend same for the discovery ontology.

egekorkan:
Merging as agreed in the call.

PR #1953, opened 2024-01-11
Delete obsolete tag review and document it in readme

egekorkan:
As agreed yesterday, we can delete obsolete documents. Since we have tagged branches for that, we can just provide the link in case somebody is looking for it.


egekorkan:
Call of 11.01: Agreed to merge

PR #1956, opened 2024-01-17
Auto formatting

### LINKED ISSUE #1926: Aligning formatting accross files
Similar to [binding templates repository](https://github.com/w3c/wot-binding-templates/blob/main/.editorconfig), we can add an editorconfig file so that :

- The documents we need to edit the source code of is easier to do and we avoid unwanted whitespace changes.
- Some documents like JSON Schema can be also read by developers so we should provide a "nice looking" document to them.

I would exclude files that are auto-generated and also all the existing HTML files in the publication folder.

egekorkan:
fixes #1926


.editorconfig
```diff
@@ -0,0 +1,20 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space # useTabs in prettier
```
JKRhb, line 7: Maybe the boolean value could be added here for clarification?

```suggestion
indent_style = space # useTabs=false in prettier
```

.prettierignore
```diff
@@ -0,0 +1,18 @@
...
+errata.html
+errata11.html
+
+# Automatically generated files
+ontology/*.html
+index.html
+validation/tm-json-schema-validation.json
+context/td-context-1.1.jsonld
+visualization/*.png
+visualization/*.svg
```
JKRhb, line 18: A trailing could be added here as well:

```suggestion
visualization/*.svg

```

.editorconfig
```diff
@@ -0,0 +1,20 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
JKRhb, line 1: Should we keep this line or should we maybe provide more context/documentation here?

JKRhb (commented):
Thank you, @egekorkan! :) See three minor comments below.

.editorconfig
egekorkan:
```suggestion
# This file allows you to configure your editor using [EditorConfig](https://EditorConfig.org).
# Make sure to install the EditorConfig plugin if your editor doesn't come with built-in support.
```

egekorkan: @JKRhb you can commit if that is what you had in mind :)

JKRhb: Looks good, thank you, @egekorkan! :)

egekorkan:
There are way too many conflicts with this PR. I am going to open a new one

PR #1960, opened 2024-01-23
Auto formatting

egekorkan:
Takes over #1956 

I had to also fix some HTML errors like missing closing tags etc.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/1960.html" title="Last updated on Jan 24, 2024, 4:36 PM UTC (c8c503c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/1960/73e21a8...c8c503c.html" title="Last updated on Jan 24, 2024, 4:36 PM UTC (c8c503c)">Diff</a>


egekorkan:
To make it easier to review, here is a summary of changes. I am writing it down but I will update this comment with the final state.
1. The "add render script results" commit (https://github.com/w3c/wot-thing-description/pull/1960/commits/46dd37e369a0e97ec7a6161e0101f73a70cc713e) only contains changes that happen after rendering.
2. Markdown lists are surrounded with empty lines. 
3. Single quotes are turned into double quotes
4. `**` is used instead of `__` for bold emphasis in markdown
5. `_` is used instead of `*` for italic emphasis in markdown
6. `-` for markdown listing instead of *
7. `[{` constructs are split into two lines
8. Space between `:` and `{` in JSON, e.g. `"fade": {` instead of `"fade":{`


Note: Even though this is a big diff, it should reduce diff size in the future. It is not possible to split this PR since the formatting has to apply everywhere.

egekorkan:
In the TD call, we noticed the following changes:
- We should add gitattribute file so that the first clone is using the correct line ending.
- Having trailing comma in code without inline JSON would be better. So we can add prettierconfig in folders with just "normal" code.

egekorkan:
@relu91 I have added the gitattributes file as well. I have taken [the one from node-wot](https://github.com/eclipse-thingweb/node-wot/blob/master/.gitattributes) but added some other file endings listed below:
- .puml
- .sparql
- .html
- .ttl
- .yml

I am merging the PR now. If there are other files I have missed, please provide a follow-up PR :)

PR #1961, opened 2024-01-23
Mastodon badge in readme

egekorkan:
Addressing point 1 of https://github.com/w3c/wot-marketing/issues/461


relu91 (approved):

egekorkan:
As it is a small editorial point, we can merge as is. We will replace it in other repos.

PR #1965, opened 2024-02-01
Reorganize Versioning document

egekorkan:
I am updating the versioning document by trying to categorize types of discussions that are documented in:
- This document
- Wiki page (which was copied into this document)
- https://github.com/w3c/wot/issues/1166
- Minutes of last week (https://www.w3.org/2024/01/25-wot-td-minutes.html)

This was a bit of a messy process. I think that this PR can be easily merged since I have kept everyone's opinion with their names. The next steps are:
- Writing user stories from the user point of view. 
- Agreeing on the content (ideally by removing names and representing TF opinion)

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -2,44 +2,84 @@
 
 In addition to the specifications, there are other resources that are managed in this repository.
 This file documents the discussion on how the versioning should be.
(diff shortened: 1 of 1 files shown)


egekorkan:
In the TD call we have merged this. More work will be done to refining

Issue #1345, opened 2021-12-28
Wrong validation errors with SHACL


egekorkan:
If one uses the shacl validation file at https://github.com/w3c/wot-thing-description/blob/main/validation/td-validation.ttl with a shacl validator like https://shacl.org/playground/ or https://www.itb.ec.europa.eu/shacl/any/upload , TDs that contain `op` in forms are not accepted. 

With the second tool, these additional errors also show up:

Details
- Value must be a valid literal of type string
Location:[Result path] - [https://www.w3.org/2019/wot/td#hasSecurityConfiguration]
Test:[Value] - [https://json-ld.org/playground/nosec_sc]
- Property needs to have at least 1 value
Location:[Result path] - [https://www.w3.org/2019/wot/td#securityDefinitions]

Since some part of the spec is generated from this file, it is probably normative and needs fixes.

Notes:
- Both validators show different errors and need a Turtle input for the TD instead of a JSON-LD. I am not sure why the error happens but it can be due to JSON-LD 1.1
- WoTHive (https://github.com/oeg-upm/wot-hive) developed by @AndreaCimminoArriaga uses the almost same shape file and also does not accept op values. I am not sure what the validation library is but I am guessing something from Apache Jena.

AndreaCimminoArriaga:
@egekorkan I changed the implementation since the ontology has changed and now the old shacl shapes throw new errors. For instance, before the key "title" was translated into dct:title but now it is translated into td:title. The shapes forced to have at least one dct:title in a valid TD, therefore now all TDs will not be valid since title is td:title. 
 
I think the file at [https://github.com/w3c/wot-thing-description/blob/main/validation/td-validation.ttl ](https://github.com/w3c/wot-thing-description/blob/main/validation/td-validation.ttl) has been updated to cope with this new terms. However, I tried with [https://shacl.org/playground/](https://shacl.org/playground/) specifying the rdf:type Thing (this is important because the shapes are endowed particularly for this type) and I get more errors:

`````rdf
[
	a sh:ValidationResult ;
	sh:resultSeverity sh:Violation ;
	sh:sourceConstraintComponent sh:InConstraintComponent ;
	sh:sourceShape _:n1855 ;
	sh:focusNode <urn:dev:ops:32473-WoTLamp-1234> ;
	sh:value "MyLampThing" ;
	sh:resultPath td:title ;
	sh:resultMessage "Value is not in Blank node _:n2014" ;
] .
[
	a sh:ValidationResult ;
[...]

PR #1966, opened 2024-02-02
SHACL validation errors

mahdanoura:
addresses #1345

### DIFF
--- a/index.html
+++ b/index.html
@@ -30,37 +30,37 @@
             name: "Sebastian Kaebisch",
             w3cid: "43064",
             company: "Siemens AG",
-            companyURL: "https://www.siemens.com/",
+            companyURL: "https://www.siemens.com/"
           },
           {
             name: "Michael McCool",
             w3cid: "93137",
             company: "Intel Corp.",
-            companyURL: "https://www.intel.com",
+            companyURL: "https://www.intel.com"
           },
           {
             name: "Ege Korkan",
             w3cid: "110131",
             company: "Siemens AG",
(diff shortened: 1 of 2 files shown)


index.html
egekorkan, line 1: Maybe minor but why are the commas removed?

danielpeintner: Seems like the `"trailingComma"` option has been set to `"none"` or so?

egekorkan: In the TD call, we had a look and it is very probably due to different prettier versions getting installed since we did not fix the version in the package.json file (npx usage).

egekorkan (commented):
This PR itself is fine but we will check the tooling before merging. This will happen asynchronously.

index.html
egekorkan: My comment can be ignored. This is happening due to index.template.html which is tracked by prettier. What is happening is normal behavior. I will fix prettier version in another PR.

validation/td-validation.ttl
```diff
@@ -32,7 +32,7 @@
                 a text for UI representation) based on a default
                 language."""^^rdf:HTML ;
         sh:nodeKind sh:Literal ;
-        sh:datatype rdf:langString ;
+        sh:or ( [ sh:datatype xsd:string ] [ sh:datatype rdf:langString ] ) ;
```
egekorkan, line 35: This was apparently not correct

relu91: I remember wrestling with something like this, maybe the comments https://github.com/w3c/wot-thing-description/pull/1684 can help @mahdanoura or @egekorkan .

PR #1969, opened 2024-02-08
WIP: Concrete Versioning Proposal

egekorkan:
@egekorkan , @lu-zero and @mahdanoura will work on it outside of the TF calls to come up with a concrete proposal based on the input provided by everyone.

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -21,6 +21,15 @@ Do we version anything until a REC release, i.e. for TD.next, do we want to publ
   - ( @egekorkan ) The versioning rules do not apply to different versions of a specification, e.g. TD 1.1 schema should be treated like a new release, not a next iteration of the 1.0 schema
   - @mmccool : We need specific snapshots for testing and plugfest purposes where we ask implementors to use a certain (pre-release) version of the resources. Better to name these versions with a word that reflects this (e.g. plugfest september 2025 version)
 
+---
+
+**Decision Making Discussion:**
+
+- McCool : I agree with versioning until development but with another namespace (e.g. unstable, development etc.) (+1 from Luca for the idea)
+- Luca: Semver can support "unstable" tags. Developers know that. McCool: the generic unstable iri should return the latest unstable though. Stable iri should return the stable version. Luca: the part after version number is up to us, we can use year-month-date for example.
+
+- Preliminary Decision: Do versioning until REC release
+
 ### Meaning of Changes
 
 Which changes are bugfixes, which are new features etc. For each type of artifact, we need to agree on the meaning. Then, these meanings can be reflected on the version information.
@@ -64,10 +73,21 @@ Notes:
 - @mmccool : we can version the folder and let git show what file has changed between versions
 - @danielpeintner : If all resources are synced, we will be talking about a certain spec version (pre-release)
 
+--- 
+
+Ege: From user's perspective, it would be better to not change the version if a file doesn't change.
+Luca: It is also a tooling problem. The tooling can replace a version tag. A warning can be displayed if we don't bump the version. Changing the tooling, readme etc. doesn't generate the warning.
+Luca: Artifacts from a release having different versions will confuse the users. Until a stable release, there is no meaning provided in semver anyways, you have to check the changes manually.
+Ege: after the talk, I tend to agree that we should sync all resources all the time (pre-release and release)
+
 ### Tooling
 
 - @egekorkan : Maybe there are options other than manually updating specific files and redirection.
 - The changelog point below is related to this
+
+---
+
+
 
 ### User Point of View
 
@@ -78,6 +98,10 @@ Who is targeted by each change and how are they affected?
 - @relu91 : There are Developers who try out the specs so there is no version "just for us".
 
 Note: We should write a user story per persona about this aspect of the discussion.
+
+---
+
+Luca: 
 
 #### Changelog
 


danielpeintner:
I think we should create a dedicated issue for it to keep better track of the problem, label it accordingly etc.

Note: there are other task forces waiting for a common versioning proposal (see https://github.com/w3c/wot-scripting-api/issues/488).

egekorkan:
I have organized the discussion and marked what are decided points between @lu-zero and @danielpeintner. There are still open points which are noted there as well.

Everything that has been discussed, it is moved to an Archive section

lu-zero:
In order to automate changelogs [conventional commits](https://www.conventionalcommits.org/en/about/) + something like [git-journal](https://github.com/saschagrunert/git-journal) might come handy, (@relu91 already mentioned it I think).

egekorkan:
TBC: CBOR-LD requires resources to be frozen. To be followed-up

VERSIONING.md
```diff
@@ -21,9 +151,18 @@ Do we version anything until a REC release, i.e. for TD.next, do we want to publ
   - ( @egekorkan ) The versioning rules do not apply to different versions of a specification, e.g. TD 1.1 schema should be treated like a new release, not a next iteration of the 1.0 schema
   - @mmccool : We need specific snapshots for testing and plugfest purposes where we ask implementors to use a certain (pre-release) version of the resources. Better to name these versions with a word that reflects this (e.g. plugfest september 2025 version)
```
TallTed, line 152:
```suggestion
  - ( @egekorkan ) The versioning rules do not apply to different versions of a specification, e.g., TD 1.1 schema should be treated like a new release, not a next iteration of the 1.0 schema
  - @mmccool: We need specific snapshots for testing and plugfest purposes where we ask implementers to use a certain (pre-release) version of the resources. Better to name these versions with a word that reflects this (e.g., plugfest September 2025 version)
```

VERSIONING.md
```diff
@@ -21,9 +151,18 @@ Do we version anything until a REC release, i.e. for TD.next, do we want to publ
...
 
+---
+
+**Decision Making Discussion:**
+
+- McCool : I agree with versioning until development but with another namespace (e.g. unstable, development etc.) (+1 from Luca for the idea)
+- Luca: Semver can support "unstable" tags. Developers know that. McCool: the generic unstable iri should return the latest unstable though. Stable iri should return the stable version. Luca: the part after version number is up to us, we can use year-month-date for example.
```
TallTed, line 159:
```suggestion
- McCool: I agree with versioning until development, but with another namespace (e.g., `unstable`, `development`, etc.) (+1 from Luca for the idea)
- Luca: Semver can support "unstable" tags. Developers know that. 
- McCool: The generic `unstable` IRI should return the latest unstable though. Stable IRI should return the stable version.
- Luca: The part after the version number is up to us; we can use year-month-date for example.
```

VERSIONING.md
```diff
@@ -21,9 +151,18 @@ Do we version anything until a REC release, i.e. for TD.next, do we want to publ
...
+
+- Preliminary Decision: Do versioning until REC release
+
 ### Meaning of Changes
 
-Which changes are bugfixes, which are new features etc. For each type of artifact, we need to agree on the meaning. Then, these meanings can be reflected on the version information.
+Which changes are bugfixes, which are new features etc. For each type of artifact, we need to agree on the meaning. Then, these meanings can be reflected in the version information.
```
TallTed, line 165:
```suggestion
Which changes are bugfixes, which are new features, etc. For each type of artifact, we need to agree on the meaning. Then, these meanings can be reflected in the version information.
```

VERSIONING.md
```diff
@@ -33,27 +172,42 @@ File types in question:
       - Patch: English typos etc.
       - Minor: Relaxing a constraint (longer strings, more oneof) so that more TDs can pass the schema.
       - Major: Adding or restricting constraints
-  - JSON-LD Context
+      - Also see: https://gitlab.openretailing.org/public-standards/api-design-guidelines/-/blob/main/Open%20Retailing%20API%20Design%20Rules%20for%20JSON.pdf
+      - 
+  - JSON-LD Context: How a JSON-LD context is versioned needs further input. No input from Ege, Daniel or Luca so far
```
TallTed, line 177:
```suggestion
  - JSON-LD Context: How a JSON-LD context is versioned needs further input. No input from Ege, Daniel, or Luca, so far.
```

VERSIONING.md
```diff
@@ -33,27 +172,42 @@ File types in question:
...
   - Ontology files
     - TTL
     - HTML
-  - TypeScript types
-  - UML Diagrams
+  - TypeScript types: Needs input from Scripting API TF what major, minor, patch means for type versions.
+  - UML Diagrams: How a UML diagram is versioned needs further input. No input from Ege, Daniel or Luca so far
```
TallTed, line 182:
```suggestion
  - UML Diagrams: How a UML diagram is versioned needs further input. No input from Ege, Daniel, or Luca, so far.
```

VERSIONING.md
```diff
@@ -33,27 +172,42 @@ File types in question:
...
 
 Notes: 
 - @mahdanoura : Also, the deprecation of terms needs to be considered, i.e. not removing but marking as deprecated.
```
TallTed, line 185:
```suggestion
- @mahdanoura: Also, the deprecation of terms needs to be considered, e.g., not removing but marking as deprecated.
```

VERSIONING.md
```diff
@@ -33,27 +172,42 @@ File types in question:
...
 - Best practices for RDF versioning should be studied. See https://www.w3.org/TR/prov-primer/
 
-#### Reflecting Meaning on Version Information
+### Reflecting Meaning of Version Information
 
 How do we reflect a patch change on the artifacts? Do we use semantic versioning, use date, and ignore semantic versioning, do we use alpha, beta etc.
```
TallTed, line 190:
```suggestion
How do we reflect a patch change on the artifacts? Do we use semantic versioning; use date and ignore semantic versioning; use alpha, beta, etc?
```

VERSIONING.md
```diff
@@ -33,27 +172,42 @@ File types in question:
...
 
 - ( @lu-zero ) A date afterward would be good "enough" for implementers. A release every month (in addition to semver) would also be good. We can also pipeline it and tag each resource exposed to github.io. 
 - ( @mjkoster ) Short commit hash would be fine. Monthly release may not make sense (unless there is a need)
 - ( @mmccool ) To make things easy we should map our directories to the url structure. We should make all URLs are versioned
 - @mahdanoura : According to OWL2 specification, for every ontology we can define an Ontology IRI (generic reference, should stay the same through versions) and a Ontology Version IRI. The difference is during dereferencing of the IRIs, the ontology IRI should redirect to the most recent version and the version IRI to the specific version.
```
TallTed, line 195:
```suggestion
- @mahdanoura : According to OWL2 specification, for every ontology we can define an Ontology IRI (generic reference, should stay the same through versions) and an Ontology Version IRI. The difference is during dereferencing of the IRIs, the ontology IRI should redirect to the most recent version, and the versioned IRI to the specific version.
```

VERSIONING.md
```diff
@@ -33,27 +172,42 @@ File types in question:
...
+  - others: Something like a changelog in the discovery can convey the meaning of change. Using one version would be simpler.
+- @egekorkan: We cannot do semantic versioning in the usual sense since TD 2.0 has set the expectation already. Or we decouple the spec version from resource version, e.g. TD 2.0 can have resources in version 25.2.1
+- Cris: CSS simply increases the major version -> TD 2 has TD 2 resources. Until we get there, we can have TD 2.pre1 TD 2.a1.
+- Luca: We can embed resource/format specific version inside the resource itself.
+- Luca: We assume the TD 2.0 will not change but the resources can have big changes since they are moving targets. E.g. JSON Schema Draft 7 becomes obsolete and we decide to update to JSON Schema 2026-09 with totally new terms etc. In this case, JSON Schema resource gets a major version bump.
+  - In this case, the resource needs its own semantic versioning lifecycle but refer to which spec version it refers to. Example for JSON Schema version: 4.0.0+2.0 becomes 5.0.0+2.0 following the example above. Next TD version, we will get 4.0.0+2.0 -> 5.0.0+3.0
```
TallTed, line 202:
```suggestion
- @egekorkan: we can do scoped versioning: `td-version:resource-version` -> `2.0.0.pre1+1.0.0.pre1`, but this can be complicated. -> Not in favor of it anymore after talking
  - others: Something like a changelog in the discovery can convey the meaning of change. Using one version would be simpler.
- @egekorkan: We cannot do semantic versioning in the usual sense, since TD 2.0 has already set the expectation. Or we decouple the spec version from the resource version, e.g., TD 2.0 can have resources in version `25.2.1`.
- Cris: CSS simply increases the major version -> TD 2 has TD 2 resources. Until we get there, we can have TD 2.pre1 TD 2.a1.
- Luca: We can embed resource/format specific version inside the resource itself.
- Luca: We assume the TD 2.0 will not change, but the resources can have big changes since they are moving targets, e.g., `JSON Schema Draft 7` becomes obsolete and we decide to update to `JSON Schema 2026-09` with totally new terms, etc. In this case, JSON Schema resource gets a major version bump.
  - In this case, the resource needs its own semantic versioning lifecycle but refer to which spec version it refers to. Example for JSON Schema version: `4.0.0+2.0` becomes `5.0.0+2.0` following the example above. Next TD version, we will get `4.0.0+2.0` -> `5.0.0+3.0`.
```

VERSIONING.md
```diff
@@ -33,27 +172,42 @@ File types in question:
...
+  - In this case, the resource needs its own semantic versioning lifecycle but refer to which spec version it refers to. Example for JSON Schema version: 4.0.0+2.0 becomes 5.0.0+2.0 following the example above. Next TD version, we will get 4.0.0+2.0 -> 5.0.0+3.0 
+
+Discussion between Luca, Daniel, Ege:
+- Scripting API's type definition versioning is currently based on JSON Schema version. Current thinking is to use `type-definition-1.2.0+json-schema-td-2.0.0+td-2.0.0` where everything is added after the + sign as informative text.
+- There are cases like Scripting API HTML spec and the type definitions, TD Ontology in HTML and TTL. We need to decide if the versions of those are always the same. In the case of Scripting API, the spec can change but the type definitions don't (1) and vice versa (2).
+  1. Informative text, explanation on the usage of a function in the TD spec (like not having `queryaction` in the spec but having it in the TD and JSON schema since 4 versions), WebIDL changes.
+  2. Automatically, the type definition gets updated but the spec hasn't caught up.
```
TallTed, line 208:
```suggestion
- Scripting API's type definition versioning is currently based on JSON Schema version. Current thinking is to use `type-definition-1.2.0+json-schema-td-2.0.0+td-2.0.0`, where everything is added after the `+` sign as informative text.
- There are cases like Scripting API HTML spec and the type definitions, TD Ontology in HTML and TTL. We need to decide if the versions of those are always the same. In the case of Scripting API, the spec can change but the type definitions don't (1) and vice versa (2).
  1. Informative text, explanation on the usage of a function in the TD spec (like not having `queryaction` in the spec but having it in the TD and JSON schema since 4 versions), WebIDL changes.
  2. Automatically, the type definition gets updated but the spec hasn't caught up.
```

VERSIONING.md
```diff
@@ -64,10 +218,13 @@ Notes:
...
-### Tooling
+--- 
 
-- @egekorkan : Maybe there are options other than manually updating specific files and redirection.
-- The changelog point below is related to this
+Ege: From user's perspective, it would be better to not change the version if a file doesn't change.
+Luca: It is also a tooling problem. The tooling can replace a version tag. A warning can be displayed if we don't bump the version. Changing the tooling, readme etc. doesn't generate the warning.
+Luca: Artifacts from a release having different versions will confuse the users. Until a stable release, there is no meaning provided in semver anyways, you have to check the changes manually.
+Ege: after the talk, I tend to agree that we should sync all resources all the time (pre-release and release)
+ScAPI: Which Scripting API version uses which TD version and which version of Scripting API implementations (e.g. node-wot) use which version of Scripting API and TD. So this kind of (external) synchronization is also relevant.
```
TallTed, line 227:
```suggestion
Luca: It is also a tooling problem. The tooling can replace a version tag. A warning can be displayed if we don't bump the version. Changing the tooling, readme, etc., doesn't generate the warning.
Luca: Artifacts from a release having different versions will confuse the users. Until a stable release, there is no meaning provided in semver anyway; you have to check the changes manually.
Ege: After the talk, I tend to agree that we should sync all resources all the time (pre-release and release)
ScAPI: Which Scripting API version uses which TD version and which version of Scripting API implementations (e.g., `node-wot`) use which version of Scripting API and TD. So this kind of (external) synchronization is also relevant.
```

TallTed (changes requested):
Various grammar, punctuation, markup fixes

PR #1970, opened 2024-02-09
Prettier Formatter Version alignment and GH Action

egekorkan:
Prettier does some basic formatting (like HTML Tidy but for other file types too). In this PR:
- I fix the prettier version in package.json and avoid using npx
- Move everything under dependencies. They were split between devDependencies and dependencies but given that we do not publish it to npm, there is no difference. However, it might be better to put everything under devDependencies?
- Run formatting once, which resulted in small changes.
- Add prettier workflow as gh action


context/td-context.ttl
egekorkan, line 1: I am not sure why there is a diff here but I do not see it

egekorkan:
> Questions:
> - As you acknoldge the are some unwanted changes, are we sure that they are just leftovers from the previous configuration? 

- Ttl file: I have no idea what the change really is. My IDE also doesn't display it. It might be encoding changes?
- Readme changes: The previous PR didn't run prettier before committing so they appear now

> - Instead of having an action that does the formatting for you and commits, I would have an action that just verifies if the code is formatted and block the PR if it isn't. If we have hasky configured properly it the verification stage here will be always green and we avoid prettify commits. 

- Husky: doesn't seem to work for some reason? 
- Overal reasoning: Sometimes we make PRs on GitHub, mostly readme files. If the PR fails due to formatting, one would need to pull the PR locally and format. So if it automatically formats, we avoid this. However, there's an  option in this action where the action overwrites the latest commit with formatting. One solution could be to write formatting on readme files and only check everything else.

context/td-context.ttl
danielpeintner: Most likely just the encoding!?

egekorkan:
Call of 27.03:
- We opted for "However, there's an option in this action where the action overwrites the latest commit with formatting."
- @egekorkan will add more information about the tool and why we want to have this.

egekorkan:
I have taken the reviews into account and this should now auto-format by overwriting the comment and not pollute the commit history.

Issue #1958, opened 2024-01-18
Higher level toolchain diagram


egekorkan:
As mentioned in yesterday's call, we should provide a simpler and higher level version of the toolchain diagram. Something like "here are the inputs you need to influence if you are working with the spec, here are the outputs it will produce"

Issue #1851, opened 2023-07-12
Examples Validation


egekorkan:
To avoid cases like https://github.com/w3c/wot-thing-description/pull/1845, we should have validation (including semantic validation) of all our examples.

JKRhb:
I suppose in order to do so, we would need to put the examples into external files and then include them [via ReSpec](https://respec.org/docs/#external-includes), right? Or could the validation also be performed on JSON embedded within the HTML document?

egekorkan:
I would imagine an approach like the first one. However, we need to find a solution to have full JSONs to run validation on but at the same time show only the relevant part of the example in the specification text.

PR #1975, opened 2024-02-19
Toolchain requirements

egekorkan:
Before starting the work on better tooling, I want to document what we currently generate and should keep on generating. Everything with a black output arrow in https://github.com/w3c/wot-thing-description/blob/main/toolchain/wot-toolchain-bpmn.png is also here but there is more.

This is **not** fixing https://github.com/w3c/wot-thing-description/issues/1958.

### DIFF
--- a/toolchain/requirements.md
+++ b/toolchain/requirements.md
@@ -6,6 +6,7 @@ by listing what kind of inputs we want to work on and what kind of outputs we wa
 ## Output Resources
 
 Always synced:
+
 - Editor's Draft index.html -> Should be provided with each PR. Used for PR Preview.
   - Users: Read by spec writes and early adopters of features.
 - UML Class Diagrams -> Integrated into index.html. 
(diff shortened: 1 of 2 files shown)


egekorkan:
@relu91 please have a final look. We want to agree on this set of requirements next week. The concrete change to another toolchain will come afterwards.

relu91:
BTW reading the requirements I remember the [validation](https://github.com/w3c/wot-thing-description/tree/main/validation) folder. Do we want to tackle that in this discussion or is rather something orthogonal?

egekorkan:
In the TD Call:

- We had another look at the requirements and added the lightweightness requirement by @relu91 . The second comment (https://github.com/w3c/wot-thing-description/pull/1975#issuecomment-1982954332) is also tracked via https://github.com/w3c/wot-thing-description/issues/1851 and is an orthogonal topic to the tooling to generate the spec and resources.
- Extended the introduction of resource files

PR #1977, opened 2024-02-20
Project Management Improvements

egekorkan:
Extending the previous description and adding a lifecycle figure.


egekorkan:
Feedback from the meeting:
- It would help to add numbers to the states so that they can be matched/linked from the text.
- Check American spelling


From offline discussion with M Koster

- PM tools

@lu-zero :
- GH checklist to issue workflow

@relu91 : simplicity

egekorkan:
I have addressed the feedback above and made the step information more obvious. It is ready for review.

egekorkan:
We have reviewed it today and there are two open points to be addressed later on:

- The numbering of steps should be more clear where step 2, 4 and 6 are "within" step 1
- Prioritization should be discussed.
Otherwise we are fine to merge.

PR #1983, opened 2024-03-12
Improvements Existing Toolchain Figure

### LINKED ISSUE #1955: Toolchain diagram changes
I have noticed a couple of issues with the toolchain diagram:
- TM Schema is not written manually, it is auto generated with this script: https://github.com/w3c/wot-thing-description/blob/main/validation/tmSchemaGenerator.js
- We integrate TypeScript type generation via the Scripting API TF: https://github.com/w3c/wot-scripting-api/blob/main/.github/workflows/sync.yml
- We integrate the test case generation (also mentioned at #1920). Basically any change to the spec that changes an assertion needs to be reflected there.

egekorkan:
fixes #1955 

I am not 100% sure about the correct BPMN usage but I will check with @mahdanoura when she is back. In the meantime, I want to merge it quickly and show it in the breakout today.


egekorkan:
In the TD coordination call we have reviewed it and agreed to merge. As mentioned above, we will check with @mahdanoura for consistency. Also to remind that we still need to generate a higher level diagram as mentioned in #1958

PR #1989, opened 2024-03-27
High-level Diagram for Toolchain and Requirements Improvement

### LINKED ISSUE #1958: Higher level toolchain diagram

egekorkan:
fixes #1958


egekorkan:
In the call of 27.03 we have looked at this and there are the following improvements needed:

- Explaining the difference between templates and source of truth
- Merge requirements and this explanation document

egekorkan:
I have incorporated PR #1991 into this and also taken the feedback into account

egekorkan:
Fixed the merge conflicts and improved the text added during the call. I also noticed some English issues that I have fixed as well. Merging as agreed upon in the call.

PR #1990, opened 2024-03-27
Simplified lifecycle diagram

egekorkan:
In order to address https://github.com/w3c/wot-thing-description/pull/1977#issuecomment-1994606748 I have simplified the diagram a bit.


egekorkan:
Call of 11.04:
- Steps 8 and 9 should be flipped in the diagram since we generally number from left to right (even in this diagram). Ege to fix this and merge asynchronously.
- This document and diagram will be presented in the next main call

PR #1998, opened 2024-04-03
Adding ignored paths to GH Action to match prettierignore for prettier changes

### LINKED ISSUE #1994: Not running prettier GH Action in unrelated changes
There are paths ignored by prettier as by https://github.com/w3c/wot-thing-description/blob/main/.prettierignore . However, a GH Action will still run and waste time. We should simply align the action to ignore the same way.

egekorkan:
fixes #1994 

Mirroring: https://github.com/w3c/wot-thing-description/blob/main/.prettierignore

### DIFF
--- a/.github/workflows/prettify.yml
+++ b/.github/workflows/prettify.yml
@@ -3,6 +3,16 @@ name: Prettifying
 on:
   pull_request:
     branches: [main]
+    paths-ignore:
+      - "publication/**"
+      - "errata.html"
+      - "errata11.html"
+      - "ontology/*.html"
+      - "index.html"
+      - "validation/tm-json-schema-validation.json"
+      - "context/td-context-1.1.jsonld"
+      - "visualization/*.png"
+      - "visualization/*.svg"
 
 jobs:
   prettier:


lu-zero (approved):

.github/workflows/prettify.yml
```diff
@@ -3,6 +3,16 @@ name: Prettifying
 on:
   pull_request:
     branches: [main]
+    paths-ignore:
+      - "publication/**"
+      - "errata.html"
+      - "errata11.html"
+      - "ontology/*.html"
+      - "index.html"
```
danielpeintner, line 11: Why do we exclude index.html ? It is auto-generated anyway...

.github/workflows/prettify.yml
```diff
@@ -3,6 +3,16 @@ name: Prettifying
...
+      - "validation/tm-json-schema-validation.json"
```
danielpeintner, line 12: Not sure about this either and if we exclude this file we should exclude https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json also

.github/workflows/prettify.yml
```diff
@@ -3,6 +3,16 @@ name: Prettifying
...
+      - "context/td-context-1.1.jsonld"
```
danielpeintner, line 13: Not sure again why we exclude just this file in the folder?

danielpeintner (commented):
Some questions below...

.github/workflows/prettify.yml
egekorkan: So the initial decision was to format only what we write and not format auto generated stuff. That is why this and index.html are ignored

egekorkan: See also https://github.com/w3c/wot-thing-description/issues/1926#issue-2006498787

These two comments should address all the questions

danielpeintner: I see the rational. Anyhow, the consequence is that once we update the tool that generates those other files might create huge diffs even though it didn't really change that much. Anyhow, I don't feel very strongly about it for now.

egekorkan: > we would directly not commit any generated file to the tree

We have to submit at least the index.html but the rest can live somewhere else.

> Anyhow, the consequence is that once we update the tool that generates those other files might create huge diffs even though it didn't really change that much

That is a good point. I am fine to format those as well, it wouldn't hurt to do so.

PR #1999, opened 2024-04-05
Term inconsistency among the ontology and context file

### LINKED ISSUE #1988: td:hasInstanceConfiguration used in JSON-LD context is not in the ontology
The JSON-LD context states: 
```
"hasInstanceConfiguration": {
      "@id": "td:hasInstanceConfiguration",
      "@type": "@id"
    },
```
Source:
https://github.com/w3c/wot-thing-description/blob/d947efd120acfd237fe2e5ba6bec8c5e5e6207d9/context/td-context.jsonld#L47

When I tried parsing the WebhookThing example (https://www.w3.org/TR/wot-thing-description11/#example-70), the generated RDF used this property (td:hasInstanceConfiguration) from the JSON-LD context, which could not be resolved in the actual ontology. 

However, I found that td:hasConfigurationInstance is available (https://github.com/w3c/wot-thing-description/blob/d947efd120acfd237fe2e5ba6bec8c5e5e6207d9/ontology/td.ttl#L133)

BTW: Thanks for your work on this spec!

mahdanoura:
fixes #1988


egekorkan:
The fixes are fine. Should we also now say that we are working on the TD.next? We should let the issue owner know that we will not be updating the 1.1 ontologies I guess

egekorkan:
Call of 11.04:
- The spec HTML has no issues. Ontology ttl and HTML files are getting a fix. Even then, we should mention this in the [errata11.html](https://github.com/w3c/wot-thing-description/blob/main/errata11.html).
- We should do a follow-up PR to the resources directory which will overwrite the current resource without versioning.
- Versioning discussion should take the errata mechanism into account.

PR #2000, opened 2024-04-18
Binding mechanism refactoring

### LINKED ISSUE #1987: Moving Binding Mechanism Text to TD
As we have agreed upon, we will move the binding mechanism of the Binding Templates document to the TD. We can start with a simple copy paste but there are already some overlapping parts. These are mentioned below but we should all have a look to make sure. Note that we do not move the individual bindings, only the core document. This can affect other WG documents but we will start with TD and Bindings first.

By looking at the Table of Contents:

- Appendix A of TD overlaps with Bindings section 5
- TD 7.2 and 7.3 have overlaps with the binding mechanism.
- TD 8.3 overlaps with HTTP binding -> We can remove 8.3 carefully by making sure that there is no information loss.
- Anything that mentions the forms container needs to be checked. E.g. TD 6.3.9 and 5.3.4.

Otherwise, we can move the whole binding document to a new section. Between sections 7 and 8 of the TD document.

One important question is: What happens to the binding templates core document? First, remove the redundant content. If the registry is a section in TD, we would deprecate the core binding. If not, the binding document becomes the registry (this is fine from process point of view).

egekorkan:
first direction to address #1987

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/2000.html" title="Last updated on Apr 25, 2024, 1:25 PM UTC (b292166)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/2000/1e16b24...b292166.html" title="Last updated on Apr 25, 2024, 1:25 PM UTC (b292166)">Diff</a>


egekorkan (commented):
Also, congrats everyone for reaching issue/pr 2000!

Issue #2007, opened 2024-04-26
Issues in Errata Document


egekorkan:
I noticed one minor issue in the published version.

> Latest Publication Date:
>    2 April 2020

https://w3c.github.io/wot-thing-description/errata11.html

Shall we quickly update the date as well? 

BTW, I am not sure whether the date is meant to relate to the TD publication  or the errata publication.
Anyhow, I think the date is wrong for both

_Originally posted by @danielpeintner in https://github.com/w3c/wot-thing-description/issues/2001#issuecomment-2078828512_

egekorkan:
I also see wrong links for technical report (point to 1.0 currently)

PR #2008, opened 2024-04-29
Errata fixes

egekorkan:
I wanted to fix https://github.com/w3c/wot-thing-description/issues/2007 but I have noticed quite a bit of problems. We were not using the correct issue labels and now that I have changed to the correct ones, we have very long erratas. The automatically loaded w3c script assumes that a repo has one specification version. So now our errata is wrong: https://w3c.github.io/wot-thing-description/errata11.html

Some ways forward:
- Option 1. Do not use the official W3C labels and create the rendered version ourselves (do not include the script)
- Option 2. Create a new repo for each REC version
- Option 3. Fork that script and filter based on TD version etc. This seems the best approach at this point but we will need to agree together.
- Option 4. Use the official labels but do not include the script and create the rendered version ourselves
- Option 5. Contribute back to the script: https://github.com/w3c/display_errata/

Notes:
- Until we solve the issue, we will not use the labels that produce unwanted text in an official document
- https://www.w3.org/2023/Process-20230612/#erratum gives a high-level description of the process.
- The entire WG should be aware that they should not use Errata label without a TF resolution.
- This procedure should be replicated for Discovery and Architecture. We can agree and test it as the TD TF and show the results in the main call.

Screenshot of the errata document with correct labels:
image

A PDF-print of the page to show the issues:
[Open Errata for the Web of Things (WoT) Thing Description 1.1 Specification.pdf](https://github.com/w3c/wot-thing-description/files/15162653/Open.Errata.for.the.Web.of.Things.WoT.Thing.Description.1.1.Specification.pdf)

### DIFF
--- a/errata11.html
+++ b/errata11.html
@@ -29,7 +29,7 @@
         <dt>Latest errata update:</dt>
         <dd><span id="date"></span></dd>
         <dt>Number of recorded errata:</dt>
-        <dd><span id="number"></span></dd>
+        <dd><span id="number"></span>2</dd>
         <dt>Link to all errata:</dt>
         <dd><span id="errata_link"></span></dd>
       </dl>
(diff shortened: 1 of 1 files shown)


egekorkan:
TD call on 2 May:

- We will merge the PR since it fixes the static version already
- For the way forward, we will ask on https://github.com/w3c/display_errata/ if we can change the script to include spec version. If yes, we will take option 5. If not, we will try option 3. Option 1 is always a fallback.

Issue #2013, opened 2024-05-08
Holistic id definition rules


egekorkan:
To avoid https://github.com/w3c/wot-thing-description/issues/1957 and possibly similar issues, we should be careful on the ids we choose for sections, examples, tables, figures. On one side, we can think of consistent patterns and on the other side, blacklist some.

Below are some proposals

Pattern:
- Examples `ex-EXAMPLENAME` (these are TD snippets)
- Sections `sec-SECTIONNAME`
- Figure `fig-FIGURENAME`

Blacklist:
- `title` since it is auto-generated for the document title by ReSpec

danielpeintner:
As sketched having our `own` prefix might be a good solution. Having said that, we could also add something based on the actual document like "td-", "arch-" etc . Not sure though if this is an overkill 🙈

PR #2014, opened 2024-05-08
remove title id from static files

### LINKED ISSUE #1957: `title` reference is not correctly linked
# WHAT

The `title` term in TD points to https://www.w3.org/2019/wot/td#title

However, clicking on this link does not take you to the "title" section (3.3.12) of the TD ontology document. 

It seems that the id of this section is `x3-3-12-title` (https://www.w3.org/2019/wot/td#x3-3-12-title).

# How

Fix the section id of 3.3.12 to `id="title"`.

egekorkan:
fixes #1957

Future fixes will be at #2013


danielpeintner:
Note: this fixes the static HTML files. Anyhow, every time we re-generate the HTML with ReSpec we need to manually remove the automatically added `id="title"` for the header.

danielpeintner (approved):

lu-zero:
An alternative is to prefix all the id for the Thing elements with `thing_` so we'd have `thing_index`.

egekorkan:
> An alternative is to prefix all the id for the Thing elements with `thing_` so we'd have `thing_index`.

I would move this discussion to #2013 . We cannot do that for 1.1 ontologies since that will break implementations

lu-zero:
I mean have respec generate the anchor this way, not changing anything on our side.

PR #2017, opened 2024-05-15
Add README for json-ld files

relu91:
Note there are ghost changes, any clue why they are happening?

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-thing-description/pull/2017.html" title="Last updated on May 15, 2024, 9:42 AM UTC (73085b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/2017/6b1b79d...relu91:73085b8.html" title="Last updated on May 15, 2024, 9:42 AM UTC (73085b8)">Diff</a>

PR #2018, opened 2024-05-15
Add text regarding choosing between events and observable properties

### LINKED ISSUE #1818: Better explain difference between observable properties and events
Further informative explanations of the differences between the two mechanisms would be nice. When to pick an event or a property is not clear.

JKRhb:
Resolves #1818.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-thing-description/pull/2018.html" title="Last updated on May 28, 2024, 11:30 AM UTC (bf11854)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/2018/2cf0b1c...JKRhb:bf11854.html" title="Last updated on May 28, 2024, 11:30 AM UTC (bf11854)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1896,6 +1896,19 @@
               <code>unsubscribeevent</code> unless the protocol supports implicit unsubscription mechanisms (e.g.,
               heartbeat to detect connection loss).
             </p>
+            <p>
+              EventAffordances are similar to observable PropertyAffordances in the sense that the Thing itself informs
+              interested Consumers about state changes. However, a main difference of EventAffordances is that not every
+              change of the associated resource needs to trigger an event message to be emitted, e.g. a critical
+              threshold for a numeric value. Furthermore, EventAffordances allow for more complex subscribing and
+              unsubscribing mechanisms via the definition of <code>subscription</code>, <code>dataResponse</code>, and
+              <code>cancellation</code> DataSchema definitions. Not all protocols might support these more advanced
+              mechanisms, however, which is why in some scenarios, events may be very similar to observable
+              PropertyAffordances. In these cases, the choice between a Property and a Event for modelling the
+              Affordance should be made based on the semantics of the underlying resource; for example, if the state of
+              the affordance is also supposed to be read or written by a Consumer, then a Property is most likely the
+              appropriate choice.
+            </p>
           </section>
           <section>
             <h3><code>VersionInfo</code></h3>
--- a/validation/td-validation.ttl
+++ b/validation/td-validation.ttl
@@ -542,7 +542,15 @@
           <code>unsubscribeevent</code>, or both terms within an
           <a href="#dfn-array" class="internalDFN" data-link-type="dfn">Array</a>.</span></p><p class="note">
 		  It is considered to be good practice that each <code>subscribeevent</code> has a corresponding <code>unsubscribeevent</code> unless the protocol
-		  supports implicit unsubscription mechanisms (e.g., heartbeat to detect connection loss).</p>"""^^rdf:HTML .
+		  supports implicit unsubscription mechanisms (e.g., heartbeat to detect connection loss).</p>
+          <p>
+            EventAffordances are similar to observable PropertyAffordances in the sense that the Thing itself informs interested Consumers about state changes.
+            However, a main difference of EventAffordances is that not every change of the associated resource needs to trigger an event message to be emitted, e.g. a critical threshold for a numeric value.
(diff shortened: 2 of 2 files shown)


JKRhb:
I was a bit unsure where to put the explanation – so far, I added it to the bottom of the section on `EventAffordance`s, but this could maybe also become part of an appendix or a more general section on best practices for TD design.

index.html
```diff
@@ -1451,7 +1451,15 @@ <h2>Core Vocabulary Definitions</h2>
           <code>unsubscribeevent</code>, or both terms within an
           <a href="#dfn-array" class="internalDFN" data-link-type="dfn">Array</a>.</span></p><p class="note">
 		  It is considered to be good practice that each <code>subscribeevent</code> has a corresponding <code>unsubscribeevent</code> unless the protocol
-		  supports implicit unsubscription mechanisms (e.g., heartbeat to detect connection loss).</p></section>
+		  supports implicit unsubscription mechanisms (e.g., heartbeat to detect connection loss).</p>
+          <p>
+            EventAffordances are similar to observable PropertyAffordances in the sense that the Thing itself informs interested Consumers about state changes.
+            However, a main difference of EventAffordances is that not every change of the associated resource needs to trigger an event message to be emitted, e.g. a critical threshold for a numeric value.
+        Furthermore, EventAffordances allow for more complex subscribing and unsubscribing mechanisms via the definition of <code>subscription</code>, <code>dataResponse</code>, and <code>cancellation</code> DataSchema definitions.
```
egekorkan, line 1458: Not exactly where but I would add the fact that if an affordance can be read, it will be a property, probably.

egekorkan (commented):
I think that the text addition is good. I have proposed a further distinction.

Regarding your question " or a more general section on best practices for TD design.". I feel that we have similar stuff in multiple places. Going forward, I think we can think of adding some tags for HTML elements that can be best practices. Like the implementation report, we can extract such best practices as a separate document for implementers. It can be a separate section but I fear that many texts can be too detached from their context. Definitely something to discuss in a call but the PR itself is fine.

index.html
```diff
@@ -1451,7 +1451,15 @@ <h2>Core Vocabulary Definitions</h2>
...
+            Not all protocols might support these more advanced mechanisms, however, which is why in some scenarios, events may be very similar to observable PropertyAffordances.
+            In these cases, the choice between a Property and a Event for modelling the Affordance should be made based on the semantics of the underlying resource.
```
JKRhb, line 1460: Thank you for your feedback, @egekorkan! I tried to address [your comment](https://github.com/w3c/wot-thing-description/pull/2018/files#r1602172783) with the following:

```suggestion
            In these cases, the choice between a Property and a Event for modelling the Affordance should be made based on the semantics of the underlying resource; for example, if the state of the affordance is also supposed to be read or written by a Consumer, then a Property is most likely the appropriate choice.
```

index.html
JKRhb, line 1: Note that the "ghost changes" @relu91 mentioned in the context of #2017 also appear in this PR – maybe we should regenerate the `index.html` file on the `main` branch?

index.html
egekorkan: Sorry for suggesting in the wrong document as well :) I like the wording and committed the suggestion

JKRhb: Thank you! :) I've now also added the lines to the source file, which actually did not change the rendered version :)

index.html
egekorkan: See #2021

JKRhb:
Hmm, for some reason, the version I generate locally is now even more different from the main branch branch than before :/ To simplify things, I would simply remove my regeneration commit, and we could then have a clean version generated on the main branch, for example.

egekorkan:
I have merged my latest pr so maybe that messed it up. If you want, please create a new PR with the same changes and we can merge async

JKRhb:
> I have merged my latest pr so maybe that messed it up. If you want, please create a new PR with the same changes and we can merge async

Hmm, do you mean a PR that regenerates the current state on the main branch? Otherwise, I could also “wrap” the commit that adds the text to the `.ttl` file in between two regeneration commits, the latter of which would illustrate the changes that would be added to the main document.

egekorkan:
Ok I have understood the problem. Basically, if you render, the index.html generated does not follow our prettify rules so it overwrites the previous formatting. Once I ran formatting after rendering, the unwanted changes were gone. I hope that we will not have these small but annoying issues with new tooling.

PR #2040, opened 2024-08-08
Initial/Common Connection Container Proposal

egekorkan:
First just moving the existing points from HackMD. The comments that were given in yesterday's call were:

- @danielpeintner: The TD got complex for implementations and humans at the same time.
- @egekorkan : This should be used in complex TDs. A simple TD version should be provided. I will also provide the "expanded" version. (This is why this PR is now a draft)
- @lu-zero : The connection is a form in the end. If we keep it that way, the parser should be fine.

### DIFF
--- a/planning/work-items/usability-and-design.md
+++ b/planning/work-items/usability-and-design.md
@@ -278,6 +278,101 @@ In this case, the Thing has enough resources and contains its own HTTP server.
 }
 ```
 
+- The following two are the same TD
+
+```js
+{
+    "connections": {
+        "basichttp" : { 
+            "href": "https://example.com",
+            "contentType": "application/cbor",
+            "security":"basic_sc",
+            "htv:methodName":"POST",
+            "reusable": false
+        }
+    },
+    "title": "test",
+    "securityDefinitions": {
+        "basic_sc":{
+            "scheme": "basic"
+        }
+    },
+    "connection": "basichttp", // can be ignored if there is only one connection?
+    "properties": {
+      "prop1": {
+           "type":"string",
+            "forms": [
+                {
+                   "op": "readproperty",
+                   "href": "myDevice/properties/prop1"
+                   // reading with POST in this case
+                }
+            ]
+        },
+        "prop2": {
+            "type":"string",
+            "forms": [
+                {
+                    "op":["readproperty"],
+                    "href": "myDevice/properties/prop2",
+                    "htv:methodName":"GET"
+                },
+                {
+                    "op":"writeproperty",
+                    "href": "myDevice/properties/prop2"
+                    // default is POST
+                }
+            ]
+        }
+    }
+}
+```
+
+```js
+{
+    "base": "https://example.com",
+    "security":"basic_sc"
+    "title": "test",
+    "securityDefinitions": {
+        "basic_sc":{
+            "scheme": "basic"
+        }
+    },
+    "properties": {
+      "prop1": {
+           "type":"string",
+            "forms": [
+                {
+                   "op": "readproperty",
+                   "contentType":"application/cbor",
+                   "href": "myDevice/properties/prop1",
+                   "htv:methodName":"POST"
+                }
+            ]
+        },
+        "prop2": {
+            "type":"string",
+            "forms": [
+                {
+                    "op":["readproperty"],
+                    "contentType":"application/cbor",
+                    "href": "myDevice/properties/prop2",
+                    "htv:methodName":"GET"
+                },
+                {
+                    "op":"writeproperty",
+                    "contentType":"application/cbor",
+                    "href": "myDevice/properties/prop2"
+                    "htv:methodName":"POST"
+                }
+            ]
+        }
+    }
+}
+```
+
+With the second TD (v1.1), we can see that it got longer due to not using the defaults of the binding. This effect would be amplified the affordances and forms the TD has.
+
 ### Data Schema Mapping
 
 ![GitHub labels](https://img.shields.io/github/labels/w3c/wot-thing-description/data%20mapping)
--- a/planning/work-items/usability-and-design.md
+++ b/planning/work-items/usability-and-design.md
@@ -205,6 +205,79 @@ In this case, the Thing has enough resources and contains its own HTTP server.
 
 ![Message Sequence](./images/initial-connection-Proxy-sequence.svg)
 
+#### New Proposals
+
+##### Common Connection Information is placed in a top-level element
+
+- An object like `securityDefinitions`, where keys can be chosen by the TD designer.
+- Inside each key, we have all elements of a form except the op keyword and an additional `reusable` keyword with a boolean. It would be false for HTTP, true for broker-based connections or WebSockets.
+- Any form that redefines the values of the connection, replaces them.
+- If a value is not defined in the connection, the value in forms are taken
+- If a value is not defined in the connection nor form, the defaults of the binding or TD spec are taken
+
+```js
+{
+    "connections": { // Should this be named to something more generic? e.g. `defaults`, `components` (like OpenAPI), `commonDefinitions`, `definitions`, `commonConnectionInformation`
+        "basichttp" : { //trying to put EVERYTHING possible
+            "href": "https://example.com", // usual base URI
+            "contentType": "application/cbor", // This is the default for this Thing's forms
+            "security":["basic_sc"], // must be defined in securityDefinitions first
+            "htv:methodName":"POST", // This is the default for this Thing. Even a property read would be with POST unless otherwise specified
+            "reusable": false, // to be discussed. Can be deduced from binding
+        },
+       "broker" : {
+            "href": "mqtt://www.w3.org/2019/wot/broker",
+            "contentType": "text/plain",
+            "security":"no_sc",
+            "reusable": true,
+            // for how long can the connection be reused?
+        }
+    },
+    "title": "test",
+    "securityDefinitions": { // should these be also embedded into connections?
+        "no_sec": {
+            "scheme": "nosec"
+        },
+        "basic_sc":{
+            "scheme": "basic"
+        }
+    },
+    "security": "no_sec", // it is probably not needed anymore -> Should we say that if there is connections element, security and base are not allowed?
+    "connection": "basichttp", // like security, this is the default connection to be used throughout
+    "properties": {
+      "prop1": {
+           "type":"string",
+            "forms": [
+                {
+                   "connection": "broker",
+                   "op": "readproperty",
+                   "href": "",
+                   "base":"application/devices/" // kind of weird, let's discuss :)
+                   "mqv:topic": "application/devices/thing1/program/commands/reset""mqv:"
+                }
+            ]
+        },
+        "prop2": {
+            "type":"string",
+            "forms": [
+                {
+                    "connection": "basichttp",
+                    "op":["readproperty"],
+                    "href": "myDevice/properties/prop2",
+                    "htv:methodName":"GET"
+                },
+                {
+                    "connection": "basichttp",
+                    "op":"writeproperty",
+                    "href": "myDevice/properties/prop2"
+                    // default is POST
+                }
+            ]
+        }
+    }
+}
+```
+
 ### Data Schema Mapping
 
 ![GitHub labels](https://img.shields.io/github/labels/w3c/wot-thing-description/data%20mapping)
--- a/planning/work-items/usability-and-design.md
+++ b/planning/work-items/usability-and-design.md
@@ -393,7 +393,7 @@ With the second TD (v1.1), we can see that it got longer due to not using the de
         },
         "basichttp2" : {
             "connection":"basichttp1",
-            "htv:methodName":"PUT", 
+            "htv:methodName":"PUT",
             "op":"readproperty"
         },
        "broker" : {
--- a/planning/work-items/usability-and-design.md
+++ b/planning/work-items/usability-and-design.md
@@ -379,6 +379,7 @@ With the second TD (v1.1), we can see that it got longer due to not using the de
 ##### Luca - Form-reuse + Connection descriptors
 
 Design ideas for Form-reuse:
+
 - `Form` gains two additional fields `reusable` and `base`.
 - `Thing` gains an additional field `bases` (`connections` in Ege's proposal).
 - `Thing::bases` is a container of `Form defaults`, a dictionary of Forms.
@@ -414,10 +415,11 @@ Design ideas for Connection descriptors
 - `Form::connection` contains a string matching a key in `Thing::connections`
 
 Open questions:
+
 - Can we move the security machinery to connection?
 - Shall we add connection to Link as well?
 - Shall we add another dictionary `key - {protocol, connection}` ?
-  - Should the key be a scheme and shall we use it to set thing-wide defaults?  e.g.:
+  - Should the key be a scheme and shall we use it to set thing-wide defaults? e.g.:
     - `https` implies `{"protocol": "htv", "connection": "tls"}`
 
 ```js
--- a/planning/work-items/usability-and-design.md
+++ b/planning/work-items/usability-and-design.md
@@ -283,7 +283,7 @@ In this case, the Thing has enough resources and contains its own HTTP server.
 ```js
 {
     "connections": {
-        "basichttp" : { 
+        "basichttp" : {
             "href": "https://example.com",
             "contentType": "application/cbor",
             "security":"basic_sc",
--- a/planning/work-items/usability-and-design.md
+++ b/planning/work-items/usability-and-design.md
@@ -380,21 +380,21 @@ With the second TD (v1.1), we can see that it got longer due to not using the de
 
 Design ideas for Form-reuse:
 
-- `Form` gains two additional fields `reusable` and `base`.
-- `Thing` gains an additional field `bases` (`connections` in Ege's proposal).
+- `Form` gains two additional fields: `reusable` and `base`.
+- `Thing` gains an additional field: `bases` (`connections` in Ege's proposal).
 - `Thing::bases` is a container of `Form defaults`, a dictionary of Forms.
 - The `base` field is a string matching one of the keys of `Thing::bases`.
 - There is a single-inheritance pattern given by `Form::base`
   - you can use the fields in the `Form` referred in `Form::base` as default for all the fields not present in the current `Form`
-  - `href` is additive, if present must be added to the `href` provided by the `base` (see normal approach to adding URIs)
-  - if `base` is set in the Form referred by the current Form `base`, the base form has to be resolved first.
-    - Implementation-wise, all the Forms in `Thing::bases`, can be resolved first once and the result cached.
+  - `href` is additive; if present, it must be added to the `href` provided by the `base` (see normal approach to adding URIs)
+  - if `base` is set in the Form referred to by the current Form `base`, the base form has to be resolved first.
+    - Implementation-wise, all the Forms in `Thing::bases` can be resolved once and the result cached.
   - All the other fields simply replace the default if they exist in the current Form
 - The `reusable` field is a boolean
-- A Consumer MAY decide to keep a connection alive and reuse it if `Form::reusable` is true, the exact behaviour is protocol-dependent
+- A Consumer MAY decide to keep a connection alive and reuse it if `Form::reusable` is true; the exact behaviour is protocol-dependent
   - if `Form::reusable` is set to false, the Consumer MUST tear down the connection and start anew.
 
-This make simpler to reuse Forms and provides the minimum amount of information to tell if a Consumer can reuse a connection.
+This make it simpler to reuse Forms, and provides the minimum amount of information to tell if a Consumer can reuse a connection.
 
 Design ideas for Connection descriptors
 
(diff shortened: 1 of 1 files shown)


planning/work-items/usability-and-design.md
```diff
@@ -376,13 +376,53 @@ Title: Common Connection Information is placed in a top-level element
...
+- The `base` field is a string matching one of the keys of `Thing::bases`.
+- There is a single-inheritance pattern given by `Form::base`
+  - you can use the fields in the `Form` referred in `Form::base` as default for all the fields not present in the current `Form`
+  - `href` is additive, if present must be added to the `href` provided by the `base` (see normal approach to adding URIs)
+  - if `base` is set in the Form referred by the current Form `base`, the base form has to be resolved first.
+    - Implementation-wise, all the Forms in `Thing::bases`, can be resolved first once and the result cached.
+  - All the other fields simply replace the default if they exist in the current Form
+- The `reusable` field is a boolean
+- A Consumer MAY decide to keep a connection alive and reuse it if `Form::reusable` is true, the exact behaviour is protocol-dependent
+  - if `Form::reusable` is set to false, the Consumer MUST tear down the connection and start anew.
```
TallTed, line 395:
```suggestion
- `Form` gains two additional fields: `reusable` and `base`.
- `Thing` gains an additional field: `bases` (`connections` in Ege's proposal).
- `Thing::bases` is a container of `Form defaults`, a dictionary of Forms.
- The `base` field is a string matching one of the keys of `Thing::bases`.
- There is a single-inheritance pattern given by `Form::base`
  - you can use the fields in the `Form` referred in `Form::base` as default for all the fields not present in the current `Form`
  - `href` is additive; if present, it must be added to the `href` provided by the `base` (see normal approach to adding URIs)
  - if `base` is set in the Form referred to by the current Form `base`, the base form has to be resolved first.
    - Implementation-wise, all the Forms in `Thing::bases` can be resolved once and the result cached.
  - All the other fields simply replace the default if they exist in the current Form
- The `reusable` field is a boolean
- A Consumer MAY decide to keep a connection alive and reuse it if `Form::reusable` is true; the exact behaviour is protocol-dependent
  - if `Form::reusable` is set to false, the Consumer MUST tear down the connection and start anew.
```

planning/work-items/usability-and-design.md
```diff
@@ -376,13 +376,53 @@ Title: Common Connection Information is placed in a top-level element
...
+
+This make simpler to reuse Forms and provides the minimum amount of information to tell if a Consumer can reuse a connection.
```
TallTed, line 397:
```suggestion
This make it simpler to reuse Forms, and provides the minimum amount of information to tell if a Consumer can reuse a connection.
```

planning/work-items/usability-and-design.md
```diff
@@ -376,13 +376,53 @@ Title: Common Connection Information is placed in a top-level element
...
+Design ideas for Connection descriptors
+
+- `Form` loses `subprotocol`
+- `Form` gains `protocol`
+- `Thing` gains `protocols` and `connections`
+- `Thing::connections` is a dictionary
+  - The key is a string
+  - The value is an Object with the following fields
+    - `href` providing the connection path if applies
+    - Any protocol-specific fields to be use to describe the connection
```
TallTed, line 408:
```suggestion
    - Any protocol-specific fields to be used to describe the connection
```

planning/work-items/usability-and-design.md
```diff
@@ -376,13 +376,53 @@ Title: Common Connection Information is placed in a top-level element
...
+- Can we move the security machinery to connection?
+- Shall we add connection to Link as well?
+- Shall we add another dictionary `key - {protocol, connection}` ?
+  - Should the key be a scheme and shall we use it to set thing-wide defaults?  e.g.:
+    - `https` implies `{"protocol": "htv", "connection": "tls"}`
 
 ```js
 {
-    "connections": { // Should this be named to something more generic? e.g. `defaults`, `components` (like OpenAPI), `commonDefinitions`, `definitions`, `commonConnectionInformation`
+    "bases": { // Should this be named to something more generic? e.g. `defaults`, `components` (like OpenAPI), `commonDefinitions`, `definitions`, `commonConnectionInformation`
```
TallTed, line 427:
```suggestion
    "bases": { // Should this be named to something more generic? e.g., `defaults`, `components` (like OpenAPI), `commonDefinitions`, `definitions`, `commonConnectionInformation`
```

TallTed (changes requested):
small, editorial

benfrancis:
I like the general direction this is going in.

I was curious how this would work for WebSockets, so as an excercise I tried to re-write the example Thing Description from my [Web Thing Protocol WebSocket Sub-protocol strawman proposal](https://docs.google.com/document/d/1KWv-aQfMgsqBFg0v4rVqzcVvzzisC7y4X4CMUYGc8rE/edit). This is what I came up with:

```json
{
  "@context": [
    "https://www.w3.org/2022/wot/td/v2.0"
  ],
  "id": "https://myweblamp.io",
  "title": "My Lamp",
  "connections": {
    "websocket": {
      "href": "wss://myweblamp.io",
      "subprotocol": "webthingprotocol",
      "security": "oauth2",
      "reusable": true
    }
  },
  "connection": "websocket",
  "securityDefinitions": {
    "oauth2": {
      "scheme": "oauth2",
      "flow": "code",
      "authorization": "https://myweblamp.io/oauth/authorize",
      "token": "https://myweblamp.io/oauth/token",
      "name": "jwt",
      "in": "query"
    }
  },
  "security": "oauth2",
  "properties": {
    "on": {
      "type": "boolean",
      "title": "On/Off",
      "description": "Whether the lamp is turned on",
      "forms": [
        {
          "op": [
            "readproperty",
            "writeproperty",
            "observeproperty",
            "unobserveproperty"
          ]
        }
      ]
    },
    "level": {
      "type": "integer",
      "title": "Brightness",
      "description": "The level of light from 0-100",
      "unit": "percent",
      "minimum": 0,
      "maximum": 100,
      "forms": [
        {
          "op": [
            "readproperty",
            "writeproperty",
            "observeproperty",
            "unobserveproperty"
          ]
        }
      ]
    }
  },
  "actions": {
    "fade": {
      "title": "Fade",
      "description": "Fade the lamp to a given level",
      "input": {
        "type": "object",
        "properties": {
          "level": {
            "type": "integer",
            "minimum": 0,
            "maximum": 100,
            "unit": "percent"
          },
          "duration": {
            "type": "integer",
            "minimum": 0,
            "unit": "milliseconds"
          }
        }
      },
      "forms": [
        {
          "op": [
            "invokeaction",
            "queryaction",
            "cancelaction"
          ]
        }
      ]
    }
  },
  "events": {
    "overheated": {
      "title": "Overheated",
      "data": {
        "type": "number",
        "unit": "degree celsius"
      },
      "description": "The lamp has exceeded its safe operating temperature",
      "forms": [
        {
          "op": [
            "subscribeevent",
            "unsubscribeevent"
          ]
        }
      ]
    }
  },
  "forms": [
    {
      "op": [
        "readallproperties",
        "readmultipleproperties",
        "writeallproperties",
        "writemultipleproperties",
        "observeallproperties",
        "unobserveallproperties",
        "queryallactions",
        "subscribeallevents",
        "unsubscribeallevents"
      ]
    }
  ]
}
```

Have I understood Ege's proposal correctly? If so feel free to use this example (or tweak it).

A couple of observations:
- I don't mind if the new member is called "connections" or "bases" (they are essentially form defaults), but I do question the idea of replacing `subprotocol` member of forms with `protocol` because these are still distinct concepts. E.g. In the example above, would `protocol` be `wss` or `webthingprotocol`? URIs are already a very well established way of serialising protocol connection information* (e.g. URI scheme) and I don't think that needs to be reinvented with a separate `protocol` member. I think a connection/base definition needs a `subprotocol` member.
- Just to highlight that a default `href` is semantically different to a `base` since in one case the default is replaced and in the other a base is extended. It needs to be possible to do both. A URI can be provided in a connection/base definition which is treated as a base URI but can be overridden with an absolute URI in a form. If the term "bases" is used then `href` is probably OK, but if the term "connections" is used perhaps `href` needs to be called `base` instead, to make this more explicit.

\* In fact you might argue that URIs are the fundamental concept that makes the web the web.

lu-zero:
> * I don't mind if the new member is called "connections" or "bases" (they are essentially form defaults), but I do question the idea of replacing `subprotocol` member of forms with `protocol` because these are still distinct concepts. E.g. In the example above, would `protocol` be `wss` or `webthingprotocol`? URIs are already a very well established way of serialising protocol connection information* (e.g. URI scheme) and I don't think that needs to be reinvented with a separate `protocol` member. I  do think a connection/base definition needs a `subprotocol` member though.

It isn't as clear as it were back in time :)

- `http://` is a subprotocol of tcp or udp?
- `https://` is a subprotocol of which version of tls?
- how to describe `socket-io` via `websocket` via `http3` using `mTLS` with a specific CA ? 

My proposal is to split the two concerns:
- one is having stackable form defaults (`bases`) 
- the other is to have a place to keep both protocol configurations and connection configurations 

This way you may say `"protocol": "webthingprotocol"` with webthingprotocol implying `over websocket over anything both consumer and thing support, at least http-1.1` or you may restrict it more by being explicit.

danielpeintner:
Just a side-comment:
Since consistency is important I would like to see the term `connections` renamed to `connectionDefinitions`. Doing so would nicely align it with the tuple `security` / `securityDefinitions`.

benfrancis:
@lu-zero wrote:
> It isn't as clear as it were back in time :)
>   * `http://` is a subprotocol of tcp or udp?
>   * `https://` is a subprotocol of which version of tls?
>   * how to describe `socket-io` via `websocket` via `http3` using `mTLS` with a specific CA ?

Of course all protocols exist somewhere in a protocol stack but in the context of the World Wide Web, and the Web of Things, protocols are identified by URI schemes. URI schemes are (supposed to be) [centrally registered](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) with IANA to avoid name conflicts.

Anything defined on top of _that_ layer can be considered a "sub-protocol", which is a formally defined concept in WebSockets (which also has a [central IANA registry](https://www.iana.org/assignments/websocket/websocket.xml)) but is also less formally used by WoT binding templates for other protocols and "[should be understood together with the protocol indicated in href of the forms (or the base)](https://w3c.github.io/wot-thing-description/#subprotocols)".

A fundamental part of the Web of Things architecture is that protocol bindings are created using hypermedia controls, which rely on URIs. If we start to split protocol identifiers out of URIs into a separate mechanism then it's a step towards making an already large problem space even larger, by opening the door to protocols which don't have a URI scheme and are not included in the IANA registry. This is a step towards facilitating bindings for directly describing non-web and non-Internet protocols like Zigbee and Z-Wave which have so far been considered out of scope of the Web of Things. I feel strongly that the working group should not go in that direction, because the scope of the Web of Things already feels unreasonably large (especially with people pushing the boundaries by inventing informal URI schemes for protocols which don't natively define one). At least anything which goes in this direction should not be called the **Web** of Things.

> My proposal is to split the two concerns:
>     * one is having stackable form defaults (`bases`)
>     * the other is to have a place to keep both protocol configurations and connection configurations

That's fine, but protocol bindings should use URIs and the scheme part of a URI (together with an optional subprotocol) unambiguously identifies the protocol being used.

> This way you may say `"protocol": "webthingprotocol"` with webthingprotocol implying `over websocket over anything both consumer and thing support, at least http-1.1` or you may restrict it more by being explicit.

In order to avoid ambiguity this would surely require the WoT Working Group to maintain its own protocol registry, in parallel to the one defined by IANA, which seems like a bad idea to me.

lu-zero:
> @lu-zero wrote:
> 
> > It isn't as clear as it were back in time :)
> > 
> > * `http://` is a subprotocol of tcp or udp?
> > * `https://` is a subprotocol of which version of tls?
> > * how to describe `socket-io` via `websocket` via `http3` using `mTLS` with a specific CA ?

> Anything defined on top of _that_ layer can be considered a "sub-protocol", which is a formally defined concept in WebSockets (which also has a [central IANA registry](https://www.iana.org/assignments/websocket/websocket.xml)) but is also less formally used by WoT binding templates for other protocols and "[should be understood together with the protocol indicated in href of the forms (or the base)](https://w3c.github.io/wot-thing-description/#subprotocols)".

I'd rather spare 3 letters and potentially have a place different from the Form to hold the protocol and connection configuration details. 

> A fundamental part of the Web of Things architecture is that protocol bindings are created using hypermedia controls, which rely on URIs. If we start to split protocol identifiers out of URIs into a separate mechanism then it's a step towards making an already large problem space even larger, by opening the door to protocols which don't have a URI scheme and are not included in the IANA registry. This is a step towards facilitating bindings for directly describing non-web and non-Internet protocols like Zigbee and Z-Wave which have so far been considered out of scope of the Web of Things. I feel strongly that the working group should not go in that direction, because the scope of the Web of Things already feels unreasonably large (especially with people pushing the boundaries by inventing informal URI schemes for protocols which don't natively define one). At least anything which goes in this direction should not be called the **Web** of Things.

If I want to describe a Thing that uses mTLS and implements only http/3 I need a place to store that information. We already have the `security*` terms that do that so it is just expanding their scope and/or rename the terms.

Having per-scheme defaults is a possible the situation in which you have multiple protocols and each of them have different security schemes that may apply.

Alternatively we can leverage the stackable defaults approach and also remove the security-related defaults since they would be redundant.

> > My proposal is to split the two concerns:
> > * one is having stackable form defaults (`bases`)
> > * the other is to have a place to keep both protocol configurations and connection configurations
> 
> That's fine, but protocol bindings should use URIs and the scheme part of a URI (together with an optional subprotocol) unambiguously identifies the protocol being used.

No disagreement on this, but you have situations in which you need and have to put additional details.
Right now we already have the situation of binding-specific terms for the "protocol-specific verb", and if we want to go that way, stackable form defaults are sufficient. 

But if we want to have more structure, and I'd argue the first step is to unify `htv:methodName`, `cov:method`, `mqv:controlPacket`, to have a single term holder.

> > This way you may say `"protocol": "webthingprotocol"` with webthingprotocol implying `over websocket over anything both consumer and thing support, at least http-1.1` or you may restrict it more by being explicit.
> 
> In order to avoid ambiguity this would surely require the WoT Working Group to maintain its own protocol registry, in parallel to the one defined by IANA, which seems like a bad idea to me.

The current consensus is that in order to have a binding in our registry, the uri schemes it uses have to be registered with IANA.

egekorkan:
This will be discussed in today's meeting. I observe that there is a discussion on the relationship between the definitions of protocol, href, URI scheme, subprotocol. For the first phase of the discussion, I would avoid going in that direction. Those can be defined under that reusable container with keywords we are free to define later on.

egekorkan:
Call of today:
- The initial idea is to start with defining the container and come back to the discussion of redefining how protocols are defined in a TD and how the physical "connection" is reused and signalled in the TD. 
- There is a question to the group on whether we should put `connectionDefinitions`, `securityDefinitions` and `schemaDefinitions` into one object, let's say `commonDefinitions` that can be referred to other places in the TD instance. This would be a lot like the `components` of the OpenAPI specs. See https://spec.openapis.org/oas/v3.1.0.html#components-object . While this is a "clean" design, it would require quite a bit of more processing effort from the Consumers. Of course, usage would not be mandatory and would make sense to use in complicated TDs.

egekorkan:
Call of 10th of October:
- The mechanism overall is fine.
- The name should be picked. Some proposals: connections and connection, bases and base, formDefinitions and form (or defaultForm). This will be rediscussed. 
- Luca: The form compositionality comes for free if we define the container values as form objects. 
  - Ege: I feel weird about op and circularity. I see a bit of the point that `op` does not have to be defined but if it is defined with `writeproperty`, actions need to override it. So it makes sense for "property-heavy` TDs.
  - Cris: Circular dependencies is a bit of a concern

@relu91 and @lu-zero feel free to comment your points.

lu-zero:
> Call of 10th of October:
> 
>     * The mechanism overall is fine.
> 
>     * The name should be picked. Some proposals: connections and connection, bases and base, formDefinitions and form (or defaultForm). This will be rediscussed.
> 
>     * Luca: The form compositionality comes for free if we define the container values as form objects.
>       
>       * Ege: I feel weird about op and circularity. I see a bit of the point that `op` does not have to be defined but if it is defined with `writeproperty`, actions need to override it. So it makes sense for "property-heavy` TDs.

`op` can be left unset as any other field, but a TD can have it set if most of its affordances have the same `op`.
The problem with circular references has to be tackled already because of `ComboSecurityScheme` so there is no additional logic needed, just has to be applied to one more item.

>       * Cris: Circular dependencies is a bit of a concern
> 
> 
> @relu91 and @lu-zero feel free to comment your points.

egekorkan:
- Form Circularity: 
  - @egekorkan I am more ok on this direction. If we add instructions on avoiding invalid TDs, it should be fine. This is nice for defining defaults for write and defaults for read.
- Reusability of Connection:
  - @lu-zero :Server may not have resources and close. Client may not need to keep it open. Cases where the Thing wants the Consumer to play nice. None of the protocols we currently have, have such a case where there is no graceful closing in the protocol level. Also it is protocol-specific.
- More Default Containers: TBC

planning/work-items/usability-and-design.md
```diff
@@ -486,6 +486,79 @@ Open questions:
...
+    "connection":"basichttp1",
+    "properties": {
+      "prop1": {
+           "type":"string",
+            "forms": [
+                {
+                   "connection": "broker",
+                   "op": "readproperty",
+                   "mqv:topic": "application/devices/thing1/program/commands/reset",
+                   "reusable": true,
```
benfrancis, line 536: Can `reusable` be defined in `connectionDefinitions` as well?

planning/work-items/usability-and-design.md
```diff
@@ -486,6 +486,79 @@ Open questions:
...
+                     "schema":"schema1",
+                   }
+                }
+            ]
+        },
+        "prop2": {
+            "type":"string",
+            "forms": [
+                {
+                    "base": "basichttp2",
```
benfrancis, line 547: Should this be `connection` rather than `base`, or are there two separate terms which behave differently?

planning/work-items/usability-and-design.md
```diff
@@ -486,6 +486,79 @@ Open questions:
...
+      "connectionDefinitions": {  // Name to be decided
+          "basichttp1" : {
+              "href": "https://example.com", // usual base URI
+              "contentType": "application/cbor", // This is the default for this Thing's forms
+              "security":["basic_sc"], // must be defined in securityDefinitions first
+              "htv:methodName":"POST", // This is the default for this Thing. Even a property read would be with POST unless otherwise specified
+              "op":"writeproperty"
+          },
+          "basichttp2" : {
+              "connection":"basichttp1", //
```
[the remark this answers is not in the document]
lu-zero: I want to have stackable defaults for forms and a container for the protocol specific terms. if there is consensus to conflate the two so be it, but...

planning/work-items/usability-and-design.md
```diff
@@ -205,6 +205,360 @@ In this case, the Thing has enough resources and contains its own HTTP server.
...
+  - The key is a string
+  - The value is a protocol identifier or an Object with the following fields
+    - A mandatory `protocol` containing a protocol identifier
+    - Additional protocol-specific fields
+- `Form::protocol` contains a string matching a key in `Thing::protocols`
+- `Form::connection` contains a string matching a key in `Thing::connections`
+
+Open questions:
+
+- Can we move the security machinery to connection?
```
lu-zero, line 419: This item hadn't been discussed yet, it boils down to either
- remove `security` and related terms and use `Form` as container for what we have in the securityDefinitions. (less tidy IMHO)
- rename `security` to `connection` and `securityDefinitions` to `connectionDefinitions` and use it to contain the connection setup terms.

``` json
"connectionDefinitions": { "no_sec+http3": {
    "scheme": "nosec",
    "htv:version": "3",
    "transport": "quick",
    "tls:cyphers": "...
  }
}
```

```json
"bases": {
  "my_read_only_http3_properties": {
    "op": "readproperty",
    "connection": "no_sec+http3",
    "htv:methodName": "...
  } 
}
```

The reason why is that `security` is just a specialized kind of connection setup covering from trasport (tls) to application (oauth).

planning/work-items/usability-and-design.md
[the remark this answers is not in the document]
egekorkan: I think this can be added later on based on the needs of the TD designers. This only makes sense if there are two defaults defined, e.g. DELETE for reading and POST for writing to a property and you have a many properties in the TD.

planning/work-items/usability-and-design.md
```diff
@@ -486,6 +486,79 @@ Open questions:
 }
 ```
 
+##### Ege: Putting Containers Together
+
+```js
+{
+    "commonDefinitions":{ // Name to be decided
```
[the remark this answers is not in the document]
egekorkan: Also makes sense. I will present it :)

planning/work-items/usability-and-design.md
egekorkan: We have decided to remove this in last week's meeting. This is up to the implementation to decide or can be deduced from the binding. If we find a good use case, it can be introduced

planning/work-items/usability-and-design.md
egekorkan: Yes indeed, it was a mistake. Please only look the latest proposal

egekorkan:
I will prepare a final example with the proposal from Luca of splitting the connection (security and other stuff) and defaults for forms. There will be also a simple markdown table with types etc. This can be used in the plugfest for testing and getting real-world experience.

egekorkan:
As agreed in the TD call yesterday, we agreed to close this PR without merging. It will be referred to in a new PR to save the history.

Issue #2045, opened 2024-09-13
JSON Schema VersionInfo


mahdanoura:
According to the specification, the [version](https://www.w3.org/TR/wot-thing-description11/#versioninfo) must have an `instance` and can have an optional `model` attribute. This is not fully reflected in the current JSON schema.

egekorkan:
Even though the table says that, the `model` keyword is restricted to TMs only. That is why, the TM JSON Schema contains it at https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json#L1914 . It would also not make sense to have the TM version in a TD since linking to a TM should be enough.

As a separate point, assuming that this is a real mistake, we would mark this as errata since this fix would need to be applied to TD 1.1 schemas. Furthermore, this gives a nice example on how we would need to version our JSON Schemas. There was a discussion before whether this would be a breaking change or not since this adds a new restriction. We agreed that it is not a breaking change.

JKRhb:
> Even though the table says that, the `model` keyword is restricted to TMs only. That is why, the TM JSON Schema contains it at https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json#L1914 . It would also not make sense to have the TM version in a TD since linking to a TM should be enough.

If I am not mistaken, this is partly related to the fact that you need to take over all the fields of a TM (except for affordances including in `tm:optional`) into the resulting TD – so if there is a `model` version present in the TM, it will also end up in the TD (also see #2042). So for that case, it might make sense to include it in the JSON Schema until TD 2.0.

relu91:
I found the description a little bit confusing too, but then I realized @JKRhb's point so that's why I think that in the end adding to TD 1.1 JSON schema would be correct even if not perfect.

mahdanoura:
The TD generated in [Example 65](https://www.w3.org/TR/wot-thing-description11/#example-65) from TM composition process provides both `model` and `instance`.

egekorkan:
Ok, you are all correct. Given that the example shows that and the spec does not say it must be removed, it should be fixed in the schema.

I am adding erratum raised label (its temporary version, a PR for errata is coming soon).

egekorkan:
TD Call: The `model` keyword is defined twice in the ontologies. @mahdanoura will check that as well.

PR #2046, opened 2024-09-13
version property updated to include the model property

mahdanoura:
addresses #2045


relu91 (approved):

danielpeintner:
Question: Do we need the same fix for other JSON schema files like https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json ?

PR #2049, opened 2024-09-18
Errata Management Update

### LINKED ISSUE #2012: Errata Management Implementation
Following up #2008. 

Some ways forward:

-     Option 1. Do not use the official W3C labels and create the rendered version ourselves (do not include the script)
-     Option 2. Create a new repo for each REC version
-     Option 3. Fork that script and filter based on TD version etc. This seems the best approach at this point but we will need to agree together.
-     Option 4. Use the official labels but do not include the script and create the rendered version ourselves
-     Option 5. Contribute back to the script: https://github.com/w3c/display_errata/

For the way forward, we will ask on https://github.com/w3c/display_errata/ if we can change the script to include a spec version. If yes, we will take option 5. If not, we will try option 3. Option 1 is always a fallback.

egekorkan:
fixes #2012 by using the policy at https://github.com/w3c/wot/blob/main/policies/errata-management.md

### DIFF
--- a/errata.html
+++ b/errata.html
@@ -12,91 +12,130 @@
 </head>
 
 <body>
-  <header>
-    <p class="banner"><a accesskey="W" href="/"><img width="72" height="48" alt="W3C"
-          src="https://www.w3.org/Icons/w3c_home" /></a> </p>
-    <br />
-    <h1 class="title">Open Errata for the Web of Things (WoT) Thing Description Specification</h1>
+
+  <p class="banner"><a accesskey="W" href="/"><img width="72" height="48" alt="W3C"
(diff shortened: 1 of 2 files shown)


egekorkan:
TD Call of 18.09:
- The footer should include the TD TF as the "creators" of the document
- The usage of TF and WG should be consistent throughout the document. Better to always say "Thing Description Task Force of the Web of Things Working Group".

egekorkan:
I have done the fixes and created the needed labels. This search's first 5 result show the labels: https://github.com/w3c/wot-thing-description/labels?q=Errata

egekorkan:
Call of 19.09: There are no objections, merging

PR #2058, opened 2024-11-24
Initial Connection Feature Description

egekorkan:
Before going to the spec document, we will use this folder for working with the feature, e.g. in the plugfest for testing. This PR takes over the discussion in #2040 and incorporates the information available at https://hackmd.io/mV9KjUC9QpOizp9kaDjqZw# that was worked on during the TD calls as well. 

A separate script will be provided to simplify testing where a new TD is transformed to the old style.

Some limitations I am seeing:

- Difficult to assert dependencies in JSON Schema, i.e. if there is a connectionDefinition, there should be connection at some point (root or form itself) or even worse: if there is no flattened or inline connection in the root form, there should be a connection object.


egekorkan:
Comments during the call today:
- Expansion should be required (or at least encouraged) for machine consumption, and simplified versions are for human readability, simple checks of binding support, etc.
- Typescript Types generation should be checked since a more complex schema can generate less usable types. We may have to manually write them.

egekorkan:
TD Call:
- Add information to the readme about the content of the folder: sandbox-like area to test the proposal
- TBC

PR #2064, opened 2024-12-19
fix: folder name for hypermedia-control-3

danielpeintner:
"hypermerdia-control-3" vs "hypermedia-control-3"

resolves issue with linking to it in https://github.com/w3c/wot-thing-description/tree/main/proposals


egekorkan (approved):
Merging as it is a simple bugfix

Issue #2022, opened 2024-05-18
Type and schema definitions for @context are too loose


JKRhb:
Dealing with the @context definition once more, I noticed that our JSON Schema currently allows for @context definitions that contain objects with non-string values (e.g., `{"foo": 42}`, see [here](https://www.jsonschemavalidator.net/s/DvSFjmEK) and the example TD below). However, passing the same TD into a JSON-LD processor (see [here](https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%5B%22https%3A%2F%2Fwww.w3.org%2F2022%2Fwot%2Ftd%2Fv1.1%22%2C%7B%22foo%22%3A2%7D%5D%2C%22title%22%3A%22Test-Thing%22%2C%22%40type%22%3A%22foo%3Ablah%22%2C%22securityDefinitions%22%3A%7B%22nosec_sc%22%3A%7B%22scheme%22%3A%22nosec%22%7D%7D%2C%22security%22%3A%22nosec_sc%22%7D&context=%7B%7D)) leads to an error since according to JSON-LD, all values have to be strings.

I therefore wonder if we could on the one hand fix the JSON Schema to enforce the usage of valid JSON-LD within TDs and, on the other hand, if we should make the type definition in the TD Specification a bit stricter for TD 2.0. While there are [some restrictions](https://www.w3.org/TR/wot-thing-description11/#td-context-ns-thing-mandatory) on the shape of the key-value pairs of the Maps within the `@context`, I think there is no MUST requirement for a map entry to have values of type `string`. So maybe this is something we can revisit for the next version.

<details>
  <summary>
    Example Thing Description
  </summary>

```json
{
    "@context": [
      "https://www.w3.org/2022/wot/td/v1.1",
      {
        "foo": 2
      }
    ],
    "title": "Test-Thing",
    "securityDefinitions": {
        "nosec_sc": {"scheme": "nosec"}
    },
    "security": "nosec_sc"
}
```
</details>

egekorkan:
Pasting this in [Playground](https://playground.thingweb.io/) results in JSON-LD validation error ` jsonld.SyntaxError: Invalid JSON-LD syntax; @context term values must be strings or objects.`. However, as you have noted, there is no assertion that says that TDs MUST be valid JSON-LDs in the first place. There are many implicit wordings but nothing very obvious.

egekorkan:
[...]

PR #2080, opened 2025-02-13
fix: align JSON Schema with JSON-LD @context format

JKRhb:
This PR fixes https://github.com/w3c/wot-thing-description/issues/2022 by specifying an explicit type of `string` for the `additionalProperties` in a TD's `@context` definition.

While resolving the issue, I noticed a small bug in the schema related to `@context`s that contain both the old and the new TD context URL. To resolve it, I adjusted the `contains` definition that is present in the corresponding "subschema".

With the changes from this PR, the following invalid TDs are now rejected by a JSON Schema validator, as the specification intends it.

#### TD with new @context URL only

```json
{
  "@context": [
    "https://www.w3.org/2022/wot/td/v1.1",
    {
      "foo": 42,
      "bar": "baz"
    }
  ],
  "title": "Test-Thing",
  "securityDefinitions": {
    "nosec_sc": {"scheme": "nosec"}
  },
  "security": "nosec_sc"
}
```

#### TD with old @context URL

```json
{
  "@context": [
    "https://www.w3.org/2022/wot/td/v1.1",
      "https://www.w3.org/2022/wot/td/v1",
    {
      "foo": 42,
      "bar": "baz"
    }
  ],
  "title": "Test-Thing",
  "securityDefinitions": {
    "nosec_sc": {"scheme": "nosec"}
  },
  "security": "nosec_sc"
}
```

### DIFF
--- a/errata11.html
+++ b/errata11.html
@@ -15,9 +15,9 @@
   <h1 class="title">Open Errata for the Web of Things (WoT) Thing Description 1.1 Specification</h1>
   <dl>
     <dt>Latest errata update:</dt>
-    <dd><span id="date"></span>06 May 2024</dd>
+    <dd><span id="date"></span>12 March 2025</dd>
     <dt>Number of recorded errata:</dt>
-    <dd><span id="number"></span>3</dd>
+    <dd><span id="number"></span>4</dd>
   </dl>
 
   <section>
@@ -103,6 +103,10 @@
     <section>
       <h3>Editorial Issues</h3>
       <ul>
+        <li> <b>2025-03-12</b>: In the <a href="https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json">Thing
+            Description JSON Schema</a> and it's associated <a href="https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json">TM Validation Schema</a>, the term <code>@context</code> was allowing adding maps (JSON objects) where the name-value pairs could contain a non-string value. Such JSON documents would be invalid JSON-LD documents, which the TD users should not create. As this only effects the resources, <a
+            href="https://www.w3.org/TR/wot-thing-description11/">WoT TD 1.1.</a> HTML specification is not effected. This is fixed in <a href="https://github.com/w3c/wot-thing-description/pull/2080">the Pull Request 2080</a>.
+        </li>
         <li> <b>2024-04-24</b>: In the <a href="https://w3c.github.io/wot-resources/td/v1.1/ontology/td.ttl">Thing
             Description ontology</a> and it's associated <a
             href="https://www.w3.org/2019/wot/td#hasConfigurationInstance">HTML page</a>, the term
--- a/errata11.html
+++ b/errata11.html
@@ -104,8 +104,8 @@
       <h3>Editorial Issues</h3>
       <ul>
         <li> <b>2025-03-12</b>: In the <a href="https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json">Thing
-            Description JSON Schema</a> and it's associated <a href="https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json">TM Validation Schema</a>, the term <code>@context</code> was allowing adding maps (JSON objects) where the name-value pairs could contain a non-string value. Such JSON documents would be invalid JSON-LD documents, which the TD users should not create. As this only effects the resources, <a
-            href="https://www.w3.org/TR/wot-thing-description11/">WoT TD 1.1.</a> HTML specification is not effected. This is fixed in <a href="https://github.com/w3c/wot-thing-description/pull/2080">the Pull Request 2080</a>.
+            Description JSON Schema</a> and its associated <a href="https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json">TM Validation Schema</a>, the term <code>@context</code> was allowing adding maps (JSON objects) where the name-value pairs could contain a non-string value. Such JSON documents would be invalid JSON-LD documents, which TD users should not create. As this only effects the resources, <a
+            href="https://www.w3.org/TR/wot-thing-description11/">WoT TD 1.1.</a> HTML specification is not affected. This is fixed in <a href="https://github.com/w3c/wot-thing-description/pull/2080">the Pull Request 2080</a>.
         </li>
         <li> <b>2024-04-24</b>: In the <a href="https://w3c.github.io/wot-resources/td/v1.1/ontology/td.ttl">Thing
             Description ontology</a> and it's associated <a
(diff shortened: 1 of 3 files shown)


egekorkan:
TD Call 20 Feb:

- We should test if this breaks something. This will be done with the new tooling but for the current tooling, we need to use manual tests or doing a PR in TD Playground at https://github.com/eclipse-thingweb/playground/blob/master/packages/core/td-schema.json , which goes through the known TDs there.
- We may need to remove the changes to the line 115. This should be already checked with the prefixedItems just above but @JKRhb will double-check.
- We still need to decide whether to push this to wot-resources or not

validation/td-json-schema-validation.json
```diff
@@ -110,15 +113,25 @@
...
           "contains": {
-            "$ref": "#/definitions/thing-context-td-uri-v1.1"
+            "allOf": [
+              {
+                "$ref": "#/definitions/thing-context-td-uri-v1.1"
+              },
+              {
+                "$ref": "#/definitions/thing-context-td-uri-v1"
+              }
+            ]
```
JKRhb, line 123: @egekorkan So, without this change, an invalid context definition is actually not being properly detected for some reason, see here: https://www.jsonschemavalidator.net/s/qp7EUNgs

danielpeintner: I think there is still a problem. The first entry can be any url if the second one is `https://www.w3.org/2022/wot/td/v1.1`
I don't think we want that to happen...

https://www.jsonschemavalidator.net/s/wDzd30PL

JKRhb: With the updated schema from this PR, that is actually being picked up by the validator: https://www.jsonschemavalidator.net/s/C6X5uTuI

danielpeintner: You are right. I used the wrong schema. I think I found another problem, but it gets dirt and might be difficult to fix. One can flip the context URLs and it is still valid.

https://www.jsonschemavalidator.net/s/h7KdeCHp

I am fine if the schema does not catch this issue 

grafik

JKRhb: Hmm, that is a good point,I wasn't sure if that is actually a bug or if a version 1.1 TD could also just add the 1.0 URL to its context :thinking: Maybe the assertion could even be clarified by an erratum, but then again, I am also not sure if the problem is that severe.

egekorkan: The links sadly do not work as the tool seems to be offline? Could you paste them to a gist?

egekorkan:
The changes relating to the restriction of non-string values in `@context` are fine. The URI enforcement does not seem to do anything so we can remove them first for this PR and make an issue to follow up on that. It is indeed the case that a wrong first URI is not causing an issue. https://www.jsonschemavalidator.net/s/Bhkf8zjH has the old schema and wrong TD.

egekorkan:
Ok this took longer than I expected/wanted :D Here is my summary:

- The tool I have shared before SHOULD NOT be used. I do not know why but it requires `$id` to be present. And once it is present, it picks up the schema defined there and not the one in the editor. I am not sure why this is a "feature". Can someone explain why? Until then, I will go back to the old stuff
- As we are using draft 7, the validation keyword `prefixItems` is simply ignored. It is introduced only in draft 2020-12, meaning any version before that is seeing some unknown keywords that have no influence. To achieve the same validation behavior, we should simply use `"items":[item1, item2]` syntax, which enforces order. Sadly, we have `prefixItems` in the official REC publications at https://github.com/w3c/wot-resources/blob/main/td/v1.1/validation/td-json-schema-validation.json#L107
- When changing the `prefixItems` to `items` and removing `contains` (not needed now), the validation behavior is as expected. Meaning that @JKRhb was in the right direction already. It was probably me adding this keyword. I am sorry for the frustrating experience fixing this @JKRhb :/
- I am not sure but we allow only the v1 context URL to be present in the context. Does anyone know why we allow that? Under the table at https://www.w3.org/TR/2023/REC-wot-thing-description11-20231205/#table-vocabulary-terms-in-thing-level we don't have any mentions of this. @danielpeintner any idea?

egekorkan:
TD Call:

- `I am not sure but we allow only the v1 context URL to be present in the context. Does anyone know why we allow that?` is intended to not break compatibility
- We are fine with the changes but we will check if anything breaks and merge. https://github.com/eclipse-thingweb/playground/pull/626 is looking at this and first results are good. @egekorkan will do some more tests.
- TM Schema should be updated as well via the script
- Should this change be pushed to wot-resources? 
  - Decision: Given that this is fixing something that is not allowed, we are fine to make this change. The errata11.html will be updated as well. A WG resolution will happen before merging this. @egekorkan will ask for an email-based resolution

egekorkan:
https://github.com/eclipse-thingweb/playground/pull/626 is working as expected with all new schemas and no previous tests failing. Once we merge this PR, I will also add more tests there to be able to see failures.

errata11.html
```diff
@@ -103,6 +103,10 @@ <h3>Substantive Issues</h3>
     <section>
       <h3>Editorial Issues</h3>
       <ul>
+        <li> <b>2025-03-12</b>: In the <a href="https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json">Thing
+            Description JSON Schema</a> and it's associated <a href="https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json">TM Validation Schema</a>, the term <code>@context</code> was allowing adding maps (JSON objects) where the name-value pairs could contain a non-string value. Such JSON documents would be invalid JSON-LD documents, which the TD users should not create. As this only effects the resources, <a
```
TallTed, line 107:
```suggestion
            Description JSON Schema</a> and its associated <a href="https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json">TM Validation Schema</a>, the term <code>@context</code> was allowing adding maps (JSON objects) where the name-value pairs could contain a non-string value. Such JSON documents would be invalid JSON-LD documents, which TD users should not create. As this only effects the resources, <a
```

errata11.html
```diff
@@ -103,6 +103,10 @@ <h3>Substantive Issues</h3>
...
+            href="https://www.w3.org/TR/wot-thing-description11/">WoT TD 1.1.</a> HTML specification is not effected. This is fixed in <a href="https://github.com/w3c/wot-thing-description/pull/2080">the Pull Request 2080</a>.
```
TallTed, line 108:
```suggestion
            href="https://www.w3.org/TR/wot-thing-description11/">WoT TD 1.1.</a> HTML specification is not affected. This is fixed in <a href="https://github.com/w3c/wot-thing-description/pull/2080">the Pull Request 2080</a>.
```

egekorkan:
WG Resolution to merge this: https://www.w3.org/2025/03/26-wot-minutes.html#d30b

egekorkan:
The changes will be reflected at https://github.com/w3c/wot-resources/tree/main/td/v1.1

PR #2081, opened 2025-02-14
Make `contentType` optional in `ExpectedResponse` and `AdditionalExpectedResponse` classes

### LINKED ISSUE #1780: Make `contentType` optional in `ExpectedResponse` class
In the `ExpectedResponse` class, it is currently mandatory to provide a `contentType`, which makes a form like the following invalid when used in a Thing Description:

```json
{
    "href": "https://example.org/foo",
    "response": {
        "description": "Success response",
        "htv:statusCodeValue": 201
    }
}
```

As identified in https://github.com/w3c/wot-thing-description/pull/1776 and https://github.com/w3c/wot-discovery/issues/465, this poses a problem for TDDs (whose API description does not always contain a `contentType` member in its `response` objects, leading to invalid TDs if no adjustments are made) and for other use-cases where a Thing does not include a Content-Type in its response.

Therefore, it should be considered to make the `contentType` member optional. As it is a normative change, this will probably have to be deferred to TD version 2.0, though.

JKRhb:
## Description of Changes

This PR deals with an issue that was discovered in the context of https://github.com/w3c/wot-discovery/issues/465 and https://github.com/w3c/wot-thing-description/pull/1776: At the moment, there is always a `contentType` assumed for an `ExpectedResponse` or `AdditionalExpectedResponse`, although there are cases where a server does not return any kind of payload (e.g., in the case of an HTTP `No Content` response). This PR provides a potential fix for the issue, although there might be some more discussion needed here.

## Related Issue

Closes #1780 

## Type of Change

Check the correct box and add the corresponding label if you are an editor.

- [ ] [Editorial (non-normative)](https://github.com/w3c/wot/blob/main/policies/async-decision.md#editorial-non-normative-changes) with label Editorial Label.

- [x] [Normative](https://github.com/w3c/wot/blob/main/policies/async-decision.md#editorial-non-normative-changes) with label Normative Label

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-thing-description/pull/2081.html" title="Last updated on Aug 27, 2025, 11:04 AM UTC (bd96988)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/2081/fdc2239...JKRhb:bd96988.html" title="Last updated on Aug 27, 2025, 11:04 AM UTC (bd96988)">Diff</a>


benfrancis (approved):

JKRhb:
I noticed in the meantime that there are also assertions related to the `ExpectedResponse` and `AdditionalExpectedResponse` classes (e.g., in [section 5.3.4.2.2](https://www.w3.org/TR/wot-thing-description11/#sec-response-usage)) that need to be adjusted. I will update this PR accordingly.

JKRhb:
I've now made a first attempt at adjusting the assertions for the two classes. I have the impression that this requires a bit more care than initially expected 😅 But I hope that the changes I made so far point in the right direction. One area that also needs to be revisited carefully is the table on the content type usage, which I haven't taken a closer look at so far.

validation/td-validation.ttl
```diff
@@ -1028,32 +1024,29 @@
...
-          content types differ and the <code>response</code>
-          name-value pair has to be used to provide response
-          content type (<code>image/jpeg</code>) information to the
-          <a href="#dfn-consumer" class="internalDFN"
-          data-link-type="dfn">Consumer</a>.</p>
+          <span class="rfc2119-assertion" id="td-expectedResponse-contentType">
+            As there is no default value defined, the content type of additional
+            responses MUST be explicitly specified.
+          </span>
+          That means that the lack of a content type value implies that the
```
egekorkan, line 1031: I would be more specific/assertive here: When there is payload in the response, the `contentType` field MUST be omitted from the expected response.

egekorkan: This would allow us to test both cases easily. Having `contentType` or not

JKRhb: > When there is payload in the response, the `contentType` field MUST be omitted from the expected response.

Hmm, do you mean when there is _no_ payload, the field must be omitted? That would sound good to me :)

egekorkan: Yes... A very well placed "typo" from my side :D

benfrancis:
@egekorkan wrote:
> contentType is optional in all levels without any default mechanism

That sounds like a big change. Would that mean that all forms which previously defaulted to `application/json` would now need an explicit `contentType`?

JKRhb:
> That sounds like a big change. Would that mean that all forms which previously defaulted to `application/json` would now need an explicit `contentType`?

Hmm, I think by default, we could stick to `application/json` as the default content type for input and output data, the latter of which could be overridden via the response classes. One thing, though, I am wondering about just now is whether we also need a way to express that a request should be empty/not have a content type. Maybe, though, we could also use `null` to specify that?

egekorkan:
> @egekorkan wrote:
> 
> > contentType is optional in all levels without any default mechanism
> 
> That sounds like a big change. Would that mean that all forms which previously defaulted to `application/json` would now need an explicit `contentType`?

Not a big change in my opinion. We will have a container to define defaults for all forms anyways. Also see https://github.com/w3c/wot-binding-templates/issues/357#issuecomment-2516671996 and the comments below it

egekorkan:
TD call 20 Feb:
- Some more changes are needed to the assertions around this feature
- The contentType in the form level (requests) can be optional as well, which will be useful for cases where the request requires no payload (e.g. an action invoke without input such as toggling a light)
- This will be a breaking change as TD2.0 TDs not having the `contentType` will mean JSON for TD 1.1 Consumers
- This can be a non-issue for some protocols that have automatic mechanisms
- It should be possible to require it at the binding level. Some protocols do not have a mechanism to provide this information in-band. So it has to be out-of-band via TD.
- The discussion (LINK TO BE INSERTED) in CoAP can be taken into account, where only the bindings define the keyword related to contentType
- We should check the impact to Scripting API algorithms.

egekorkan:
TD call 05 March: 
- We are fine to make `contentType` optional with no default in all levels.
- The relationship between the lack of data schema (e.g. no input or no output in an action or no data in an event) and the `contentType` should be explained in the text, probably with assertions. The same for the presence of data schema, i.e. when there is data schema, there must be `contentType` (no default value)

benfrancis:
> We are fine to make contentType optional with no default in all levels.

I don't love the global default being removed. Can this at least wait until there is somewhere to specify a default for a specific Thing Description? Otherwise I'm going to have to start adding `"contentType": "application/json"` in every single Form.

egekorkan:
@benfrancis there will be the way to add a global default for a data schema as a result of https://github.com/w3c/wot-thing-description/tree/main/proposals/initial-connection

JKRhb:
Revisiting this issue now after a while, I am not entirely sure which direction to take regarding the default value for `contentType`. Should we maybe use `application/json` for all `contentType` instances for now and then change the default behavior afterward, as part of a different discussion? Or should the "default discussion" be resolved first?

validation/td-validation.ttl
```diff
@@ -991,22 +990,18 @@
...
-          be applied. <span class="rfc2119-assertion" id=
-          "td-expectedResponse-default-contentType">If no
-          <code>response</code> name-value pair is provided, it
-          MUST be assumed
-          that the content type of the response is equal to the
-          content type assigned to the Form instance.</span> Note
-          that <code>contentType</code> within an
+          be applied.
+          <span class="rfc2119-assertion" id=
+          "td-expectedResponse-default-contentType">
```
egekorkan: the id should be changed since the assertion text is different now

validation/td-validation.ttl
```diff
@@ -991,22 +990,18 @@
...
+          If no <code>response</code> name-value pair is provided, it MUST be
```
egekorkan: Not expected change but it makes sense and is more correct. This will make it annoying for actions with an input and output as we need to have contentType in form and in response objects. Some more thinking is needed as this will have implementation impact.

validation/td-validation.ttl
```diff
@@ -1028,32 +1023,27 @@
...
-          name-value pair has to be used to provide response
-          content type (<code>image/jpeg</code>) information to the
-          <a href="#dfn-consumer" class="internalDFN"
-          data-link-type="dfn">Consumer</a>.</p>
+          <span class="rfc2119-assertion" id="td-expectedResponse-missing-contentType">
+            As there is no default value defined, the lack of a <code>contentType</code> value
+            implies that the response <em class="rfc2119" title="MUST NOT">MUST NOT</em> contain
+            any payload.
+          </span>
+          If a content type is defined, a
```
egekorkan: TD Call: This can be simplified by saying that if there is an form level content type, that should imply that the response also has that content type, unless it is overwritten

egekorkan: But no, this would mean that we cannot describe the lack of response, going back to the beginning. To be discussed.

validation/td-validation.ttl
```diff
@@ -1028,32 +1023,27 @@
...
+          <a href="#dfn-consumer" class="internalDFN"data-link-type="dfn">
```
egekorkan: Based on the discussion above, this should be an assertion too

egekorkan (commented):
We had a look in the TD call. There are some unintended side effects that the TF will look at outside of the call. One weird change: 5.3.4.4 table has a change in the preview but there should be none?

validation/td-validation.ttl
JKRhb: > But no, this would mean that we cannot describe the lack of response, going back to the beginning. To be discussed.

Reiterating an earlier idea: Would it be possible to specify the lack of a content type with `null`? Otherwise, I guess we would need to assume that responses by default do not have a `contentType`.

egekorkan: > we would need to assume that responses by default do not have a contentType

The whole reason why there is a `response` is to signal that there is a different contentType for the response. I feel we are a bit stuck here but only a bit. Some ways forward:

- response having `contentType` has `null` value. I am not sure if this complies with https://datatracker.ietf.org/doc/html/rfc6838#section-1 . If an implementation needs special treatment to handle `null` value, it would be bad for implementers. 
- response can have the `null` value. so something like below:
```js
{
//...
  "contentType":"application/json",
  "response":null
}
```
- We add an assertion that if the action has no `output` field, there can be no response payload, thus no `response` field in the form. If the response is an image (or similarly non structured format), there should be an empty output object and of course a `response` in the form.

benfrancis: @egekorkan wrote:

> The whole reason why there is a response is to signal that there is a different contentType for the response.

If no `contentType` member means no content and the `contentType` member of `ExpectedResponse` is optional, then how about to signal an operation with a JSON request but a response with no body you could use an empty `ExpectedResponse` object:

```json
{
  ...
  "contentType":"application/json",
  "response": {}
}
```

If the response member is not included at all then it is assumed the response uses the same `contentType` as the request.

@egekorkan wrote:
> We add an assertion that if the action has no output field, there can be no response payload, thus no response field in the form.

If I have understood correctly this would not allow for the case of an asynchronous action in HTTP where the `invokeaction` request has an action input in JSON and expects a response, but rather than containing the action output the response is a `201 Created` response with no body but a `Location` header containing the URL of a resource to query for the action's status.

JKRhb: > how about to signal an operation with a JSON request but a response with no body you could use an empty `ExpectedResponse` object

Thank you, @benfrancis, I think that would be a good solution :)

egekorkan: `response:{}` is also ok for me.

> If I have understood correctly this would not allow for the case of an asynchronous action in HTTP where the invokeaction request has an action input in JSON and expects a response, but rather than containing the action output the response is a 201 Created response with no body but a Location header containing the URL of a resource to query for the action's status.

OK so that is not possible to describe atm but should be possible with the new data mapping work where the data schema content can be mapped to different parts of a message (header vs body etc.). Even then, we would want to say that there is an output (let's say string) and it is in header. So the output empty idea of mine doesn't work and I like this one more. Given that @JKRhb also liked this, I would say go ahead :) However, some examples are going to be needed to explain this case.

egekorkan: Approved in the TD call today :)

index.html
```diff
@@ -4094,28 +4092,26 @@ <h4>Response-related Terms Usage</h4>
...
+                  If a content type is defined in the <code>response</code> object within a form, a
+                  <a href="#dfn-consumer" class="internalDFN" data-link-type="dfn"> Consumer </a>
+                  MUST expect a response containing a payload in the respective format, e.g., an image in the case of
+                  <code>image/jpeg</code>.
+                </span>
               </p>
               <p>
-                Similar considerations apply to additional responses, although in this case the
-                <code>contentType</code> is optional if it is the same as the input content Type (e.g. JSON).
+                Similar considerations apply to additional responses.
```
JKRhb: Reading this sentence once again, I am wondering whether this should be turned into an explicit assertion, as technically, this does not have a normative effect right now, I assume.

egekorkan:
TD call today:
- Example 37 should have an input to show that the contentType belongs to the input
- @egekorkan will check the HTML diff to see if the ids of the assertions are ok.

index.html
```diff
@@ -4068,11 +4068,11 @@ <h4>Response-related Terms Usage</h4>
 
               <p>
                 The optional <code>response</code> name-value pair can be used to provide metadata for the expected
-                response message. With the core vocabulary, it only includes content type information, but TD Context
+                response message. With the core vocabulary, it may only include content type information, but TD Context
```
```suggestion
                response message. With the core vocabulary, it might only include content type information, but TD Context
```

index.html
```diff
@@ -4092,15 +4092,18 @@ <h4>Response-related Terms Usage</h4>
...
+                is no default value defined,
                 <span class="rfc2119-assertion" id="td-expectedResponse-missing-contentType">
-                  As there is no default value defined, the lack of a <code>contentType</code> value implies that the
-                  response <em class="rfc2119" title="MUST NOT">MUST NOT</em> contain any payload.
+                  the lack of a <code>contentType</code> value in a <code>response</code> object within a form implies
+                  that the response <em class="rfc2119" title="MUST NOT">MUST NOT</em> contain any payload.
+                </span>
+                <span class="rfc2119-assertion" id="td-expectedResponse-defined-contentType">
+                  If a content type is defined in the <code>response</code> object within a form, a
+                  <a href="#dfn-consumer" class="internalDFN" data-link-type="dfn"> Consumer </a>
```
```suggestion
                  <a href="#dfn-consumer" class="internalDFN" data-link-type="dfn">Consumer</a>
```

index.html
```diff
@@ -6219,6 +6207,30 @@ <h3><code>response</code></h3>
     },
     // ...
 }
+</pre
+            >
+
+            <p>
+              In cases where responses are not expected to contain a payload and therefore no `contentType` as well, an
```
```suggestion
              In cases where responses are not expected to contain a payload and therefore no `contentType` either, an
```

index.template.html
```diff
@@ -3231,6 +3216,30 @@ <h3><code>response</code></h3>
     },
     // ...
 }
+</pre
+            >
+
+            <p>
+              In cases where responses are not expected to contain a payload and therefore no `contentType` as well, an
```
TallTed:
```suggestion
              In cases where responses are not expected to contain a payload and therefore no `contentType` either, an
```

validation/td-validation.ttl
```diff
@@ -988,13 +988,13 @@
           <p>The optional <code>response</code> name-value pair can
           be used to provide metadata for the expected response
           message. 
-          With the core vocabulary, it only includes
+          With the core vocabulary, it may only include
```
TallTed:
```suggestion
          With the core vocabulary, it might only include
```

index.html
JKRhb: Addressed via 89936e32da2d7fc5ed4e05c0c732ca2119e44e03

index.html
JKRhb: Addressed via 89936e32da2d7fc5ed4e05c0c732ca2119e44e03 (the HTML was unnecessarily complicated here).

index.html
JKRhb: With "may", I actually wanted to say that the TD core vocabulary only _allows_ for the inclusion of content type information, so I am not sure whether "might" is a better word to use here (but I am of course not a native speaker).

JKRhb:
> TD call today:
> 
>     * Example 37 should have an input to show that the contentType belongs to the input
> 
>     * @egekorkan will check the HTML diff to see if the ids of the assertions are ok.
> 
>     * The paragraph starting with `Similar considerations apply to additional responses. ` should reflect - with assertions - about the lack of `contentType` in response.
>       
>       * Here, also splitting it into a list would help as some "structured thinking" is needed with conditions.

I think these aspects _should_ now be addressed :) Looking forward to your feedback here.

index.html
TallTed: `may` often gets misread as RFC2119 MAY; `might` has almost the same connotation, without the misunderstanding. A slightly larger change might suit you better --
```suggestion
                response message. With the core vocabulary, it can include only content type information, but TD Context
```

validation/td-validation.ttl
TallTed: As above...
```suggestion
          With the core vocabulary, it can include only
```

index.html
JKRhb: @TallTed Thank you! I incorporated your suggestions via d369fef7f8078ce73ad47e3c55a69265a9582479 and c6ced0b3af8e027ff7aa9a6901195ac8d486c9bc :)

PR #2086, opened 2025-03-05
PR Template: Remove Checkboxes

egekorkan:
(not using the template since it is about changing it)

I am doing this since the PRs now show with "tasks" due to checkboxes (see below). The PR submitter can simply delete a line instead.

Screenshot 2025-03-05 at 15 51 17


egekorkan:
Agreed in the call, merging

PR #2090, opened 2025-03-26
Initial Connection Restructuring

egekorkan:
## Description of Changes

I am making the initial connection part easier to read. First turning bullet list into tables and then moving the diagrams to another readme.

## Type of Change

### DIFF
--- a/proposals/initial-connection/README.md
+++ b/proposals/initial-connection/README.md
@@ -54,7 +54,7 @@ It is possible to use the base term for simplifying the endpoint but it has limi
 1. If the media type is common across forms but is not `application/json`, it is repeated in each form.
 2. If there are common protocol stack configurations such as different default verbs, baud rates, and endianness, they are repeated in each form
 3. Multiple bases are not possible. Thus, each form repeats multiple bases. This is relevant when a TD has local and public IP addresses
-4. For protocols that are based on an initial connection and then subsequent messages, the semantics are not clear. Thus, a Consumer can establish multiple connections instead of reusing the initial connection. See the Example of the Message Flow section below
+4. For protocols that are based on an initial connection and then subsequent messages, the semantics are not clear. Thus, a Consumer can establish multiple connections instead of reusing the initial connection. See the [Example of the Message Flow section](#examples-of-message-sequences) at the end of the document.
 
 Related Issues:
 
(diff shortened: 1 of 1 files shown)


proposals/initial-connection/README.md
```diff
@@ -255,6 +255,13 @@ Fields that override the starting default for all the elements of that kind.
 
 These set a default for the whole Thing
 
+| Vocabulary Term | Description | Assignment | Type | Remarks |
+|---|---|---|---|---|
+| connection | A reference to or an in-place definition of a connection definition, if missing a protocol binding default is in place. | optional | String pointing to the `connectionDefinitions` map or Object with type `Connection`. | If string, it MUST refer to a first-level key in `connectionDefinitions`. |
```
lu-zero, line 260:
```suggestion
| connection | A reference to or an in-place definition of a connection definition, protocol bindings MAY provide default definitions. | optional | String pointing to the `connectionDefinitions` map or Object with type `Connection`. | If string, it MUST refer to a first-level key in `connectionDefinitions`. |
```

Maybe?

lu-zero:
Probably we could write a generic paragraph explaining that the mechanism is always the same before this table?

proposals/initial-connection/README.md
egekorkan: That is a discussion to be had :D I want to make this PR rather straightforward with some basic text wrangling

lu-zero: it is in any case an improvement I'd say.

egekorkan: I think that defaults will be possible in all cases but I want to have it as a separate discussion. All cases meaning HTTP binding can say the default http version, tls version. This affects the connection term. security can be also affected in a way but these are then some longer discussions. Until we know how the defaults are going to be specified, no need to add an assertion

egekorkan: Sorry I have misunderstood something here... I am not sure what we meant here in the beginning. I will copy over but create an issue to follow up.

proposals/initial-connection/README.md
```diff
@@ -287,46 +287,29 @@ Note: Even if a single form of an affordance is not complete, a defaultable elem
...
-  - _Mandatory_: with default
-  - _Description_:
-  - _Remarks_:
+| Vocabulary Term | Description                                                                                                                      | Assignment   | Type                                  | Remarks                                                                   |
+| --------------- | -------------------------------------------------------------------------------------------------------------------------------- | ------------ | ------------------------------------- | ------------------------------------------------------------------------- |
+| connection      | A reference to or an in-place definition of a connection definition                                                              | optional     |                                       | If string, it MUST refer to a first-level key in `connectionDefinitions`. |
+| op              | Indicates the semantic intention of performing the operation(s) described by the form.                                           | with default | String or Array of String (no change) |                                                                           |
```
TallTed, line 306:
```suggestion
| op              | Indicates the semantic intention of performing the operation(s) described by the form.                                           | with default | String or Array of Strings (no change) |                                                                           |
```

proposals/initial-connection/README.md
```diff
@@ -287,46 +287,29 @@ Note: Even if a single form of an affordance is not complete, a defaultable elem
...
+| contentType     | Assign a content type based on a media type (e.g., text/plain) and potential parameters (e.g., charset=utf-8) for the media type | with default | String (no change)                    |                                                                           |
```
TallTed, line 307:
```suggestion
| contentType     | Assign a content type based on a media type (e.g., `text/plain`) and potential parameters (e.g., `charset=utf-8`) for the media type | with default | String (no change)                    |                                                                           |
```

PR #2091, opened 2025-04-01
Publication Folder Sync for Schemas

egekorkan:
## Description of Changes

copying over the fixed schemas to the publication folder. Now going to replicate in resources repo

### DIFF
--- a/publication/ver11/7-resources/validation/td-json-schema-validation.json
+++ b/publication/ver11/7-resources/validation/td-json-schema-validation.json
@@ -1,6 +1,6 @@
 {
   "title": "Thing Description",
-  "version": "1.1-09-November-2023",
+  "version": "1.1-12-March-2025",
   "description": "JSON Schema for validating TD instances against the TD information model. TD instances can be with or without terms that have default values",
   "$schema": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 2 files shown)


publication/ver11/7-resources/validation/td-json-schema-validation.json
```diff
@@ -1,9 +1,9 @@
 {
   "title": "Thing Description",
-  "version": "1.1-09-November-2023",
+  "version": "1.1-12-March-2025",
```
TallTed, line 3: I strongly advise using [ISO8601](https://xkcd.com/1179/) date format (YYYY-MM-DD) for most if not all purposes.

Among other things, this ensures that any sortable (e.g., tabular) listing will always put "March" ("03") earlier than "December" ("12"), while the format used above will always put "July" ("07") before "June" ("06").

danielpeintner: I agree with the statement. I just wonder whether it has any *negative* effect since we were using the other format in the past 🤷‍♂️

egekorkan:
TD call 02 April: we decide to merge it and adopt a better versioning scheme in TD 2.0 resources. Also see #1969

Issue #2094, opened 2025-04-17
Collecting Analysis on Data Mapping


egekorkan:
The TF welcomes new additions to the data mapping analysis at https://github.com/w3c/wot-thing-description/blob/main/planning/work-items/analysis/analysis-data-mapping.md .

Please provide your input in terms of related issues, existing approaches, use cases etc. Feel free to write new user stories as well. However, the summary or the requirements will be done later.

https://github.com/w3c/wot-thing-description/blob/main/proposals/initial-connection/README.md can be used as a template

egekorkan:
https://github.com/w3c/wot-thing-description/blob/main/planning/work-items/analysis/analysis-data-mapping.md is the place to update the analysis. First we should review the document and then start adding the user stories, which should be written in the "Who, What, Why" format. See https://github.com/w3c/wot-thing-description/tree/main/proposals/initial-connection#user-stories

egekorkan:
In general, we should put the concrete examples. For uriVariables in HTTP, one example is https://developer.weather.com/forecastsandinsights/docs/currents-on-demand-v-3-0#/Currents%20On%20Demand/get_v3_wx_observations_current?geocode but we should find more. Afaik, most weather APIs were very similar.

PR #2105, opened 2025-05-28
Add HTTP to the data mapping analysis

lu-zero:
## Description of Changes

Data mapping for the HTTP protocol, we might spin out and have a section for URL itself alone.

## Related Issue

Part of #2094

## Type of Change


relu91 (approved):
Looking good, it would be beneficial to have examples of "services expecting such patterns".

PR #2117, opened 2025-07-11
fix uri variable example

### LINKED ISSUE #2038: Inconsistency or unexplained value transformation: metric and metric_value
The enum `unitSystem` has the value `metric_value` but the example url after the code block uses `&unitSystem=metric`.

It is unclear if `_value` is a special suffix that needs to be stripped by clients or if this is an inconsistency in the WoT spec.

https://www.w3.org/TR/wot-thing-description/#td-forms-urivariables-thing-example

egekorkan:
## Description of Changes

Just a simple oversight in an example mentioned in #2038

## Related Issue

Closes #2038

## Type of Change

### DIFF
--- a/index.html
+++ b/index.html
@@ -1123,862 +1123,363 @@
           <h2>Core Vocabulary Definitions</h2>
 
           <!-- rendered from RDF definitions -->
-          <section>
-            <h3><code>Thing</code></h3>
-            <p>
-              An abstraction of a physical or a virtual entity whose metadata and interfaces are described by a WoT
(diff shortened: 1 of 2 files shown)


TallTed:
It seems rather odd for a "simple oversight in an example" to change ~5000 lines. This might want another look.

egekorkan:
> It seems rather odd for a "simple oversight in an example" to change ~5000 lines. This might want another look.

That is true but index.html is always auto-generated. All the changes seem like formatting changes though. I will have a deeper look.

PR #2122, opened 2025-08-20
Profile explanation

### LINKED ISSUE #2121: Adding explanation about the difference to profiles
In the binding section, we should add a text that explains the difference between bindings and profiles. This is a discussion that probably exists somewhere but we should track it for now. @benfrancis could you point somewhere?

egekorkan:
## Description of Changes

I am adding explanation about how bindings relate to profiles.

## Related Issue

Closes #2121 

## Type of Change

### DIFF
--- a/index.html
+++ b/index.html
@@ -7526,6 +7526,27 @@
           </p>
         </section>
       </section>
+
+      <section id="profile-relationship">
+        <h4>Relationship to Profiles</h4>
+        <p>
+          The Web of Things also define a profiling mechanism in [[WOT-PROFILES]] that allows the reduction of
+          implementation effort. Profiles are built on top of existing Bindings by constraining the flexibility of TDs.
+          Such constraints can be bindings, semantic contexts, link relations, security schemes and discovery
+          mechanisms. As there is less flexibility, the implementation effort to achieve out-of-the-box interoperability
+          is reduced.
+        </p>
+        <p>
+          For example, a profile can constrain the usage of HTTP GET Method for <code>readproperty</code> operations and
+          require payloads to be always serialized in JSON. The usage of a profile is indicated by the
+          <code>profile</code> term in the TD, which is a URI that points to the profile definition.
+          <span class="rfc2119-assertion" id="td-consumer-profile-ignore">
+            A Consumer MAY ignore the <code>profile</code> term and use the TD as it is, but the knowledge of a profile
+            allows the Consumer to implement the profile-specific constraints as assumptions and thus reduce the
+            implementation effort on understanding the TD.
+          </span>
+        </p>
+      </section>
     </section>
 
     <section id="behavior" class="normative">
--- a/index.template.html
+++ b/index.template.html
@@ -4531,6 +4531,27 @@
           </p>
         </section>
       </section>
+
+      <section id="profile-relationship">
+        <h4>Relationship to Profiles</h4>
+        <p>
+          The Web of Things also define a profiling mechanism in [[WOT-PROFILES]] that allows the reduction of
+          implementation effort. Profiles are built on top of existing Bindings by constraining the flexibility of TDs.
+          Such constraints can be bindings, semantic contexts, link relations, security schemes and discovery
+          mechanisms. As there is less flexibility, the implementation effort to achieve out-of-the-box interoperability
+          is reduced.
+        </p>
+        <p>
+          For example, a profile can constrain the usage of HTTP GET Method for <code>readproperty</code> operations and
+          require payloads to be always serialized in JSON. The usage of a profile is indicated by the
+          <code>profile</code> term in the TD, which is a URI that points to the profile definition.
+          <span class="rfc2119-assertion" id="td-consumer-profile-ignore">
+            A Consumer MAY ignore the <code>profile</code> term and use the TD as it is, but the knowledge of a profile
+            allows the Consumer to implement the profile-specific constraints as assumptions and thus reduce the
+            implementation effort on understanding the TD.
+          </span>
+        </p>
+      </section>
     </section>
 
     <section id="behavior" class="normative">


index.html
```diff
@@ -7526,6 +7526,27 @@ <h5>Protocol and Payload Bindings</h5>
           </p>
         </section>
       </section>
+
+      <section id="profile-relationship">
+        <h4>Relationship to Profiles</h4>
+        <p>
+          The Web of Things also define a profiling mechanism in [[WOT-PROFILES]] that allows the reduction of
+          implementation effort. Profiles are built on top of existing Bindings by constraining the flexibility of TDs.
```
benfrancis: I agree this is how Profiles should work in WoT 2.0. However, this arguably isn't true for the set of profiles defined in the Profiles 1.0 specification which you are citing here, which define their own protocol bindings which can be used independent of 1.0 binding templates.

In other words, I think this wording will be fine once there is a WOT-PROFILES-2 specification to cite, with a  set of 2.0 Profiles built on top of 2.0 Bindings, but currently it's not completely true.

Not sure how to fix this in the meantime other than removing the reference.

index.html
```diff
@@ -7526,6 +7526,27 @@ <h5>Protocol and Payload Bindings</h5>
...
+          Such constraints can be bindings, semantic contexts, link relations, security schemes and discovery
+          mechanisms. As there is less flexibility, the implementation effort to achieve out-of-the-box interoperability
+          is reduced.
```
benfrancis: "Less flexibility" is a bit of negative way of framing the constraints that Profiles impose.

This sentence implies that it is possible to achieve out-of-the-box interoperability between Consumers and Things without Profiles, if you just put enough effort into your implementation. I don't think this is actually true, because with an unbounded set of protocols and payload formats etc. it is _never_ possible to guarantee that a given Consumer can communicate with all possible Things.

I understand that you don't like to say that Profiles "guarantee interoperability" since there are arguably levels of interoperability (e.g. semantic interoperability) which are currently still left open, but they certainly provide _additional_ interoperability guarantees over relying on bindings.

index.html
```diff
@@ -7526,6 +7526,27 @@ <h5>Protocol and Payload Bindings</h5>
...
+        </p>
+        <p>
+          For example, a profile can constrain the usage of HTTP GET Method for <code>readproperty</code> operations and
+          require payloads to be always serialized in JSON. The usage of a profile is indicated by the
+          <code>profile</code> term in the TD, which is a URI that points to the profile definition.
+          <span class="rfc2119-assertion" id="td-consumer-profile-ignore">
+            A Consumer MAY ignore the <code>profile</code> term and use the TD as it is, but the knowledge of a profile
+            allows the Consumer to implement the profile-specific constraints as assumptions and thus reduce the
+            implementation effort on understanding the TD.
+          </span>
```
benfrancis: I agree with this assertion, but I'm not sure it belongs in the Bindings section of the specification, since the constraints of Profiles extend beyond bindings.

benfrancis (changes requested):
I agree that it's a good idea to try to explain the relationship between Bindings and Profiles in 2.0.

However, I wonder whether it might be better to add this kind of text to a future Profiles section rather than the Bindings section, since bindings are one of multiple extension points that we are aiming for Profiles to constrain, and we can explain all of them in one place.

index.html
egekorkan: Yes I went with the assumption of the new profile mechanism. I propose removing the reference here and adding an editor's note that explains the situation?

index.html
egekorkan: > but they certainly provide additional interoperability guarantees over relying on bindings.

I like this formulation. I will add this.

index.html
egekorkan: I am fine to remove this assertion but we don't have a profiles section in this spec. The only mention of profile is at https://w3c.github.io/wot-thing-description/#thing table. 

How about referencing https://w3c.github.io/wot-profile/#profiling-mechanism ?

benfrancis: > How about referencing https://w3c.github.io/wot-profile/#profiling-mechanism ?

That is probably the best we can do for now.

I do think the assertion seems a bit out of place here though.

egekorkan: Done :)

egekorkan:
@benfrancis I have applied your feedback. If it is fine for you, you can resolve or approve the PR.

index.template.html
```diff
@@ -4531,6 +4531,28 @@ <h5>Protocol and Payload Bindings</h5>
           </p>
         </section>
       </section>
+
+      <section id="profile-relationship">
+        <h4>Relationship to Profiles</h4>
+        <p>
+          The Web of Things also define a profiling mechanism that allows the reduction of implementation effort.
```
danielpeintner:
```suggestion
          The Web of Things also defines a profiling mechanism that allows the reduction of implementation effort.
```

index.template.html
```diff
@@ -4531,6 +4531,28 @@ <h5>Protocol and Payload Bindings</h5>
...
+          Profiles are built on top of existing Bindings by constraining the flexibility of TDs. Such constraints can be
+          bindings, semantic contexts, link relations, security schemes and discovery mechanisms. With these
+          constraints, they provide additional interoperability guarantees over bindings.
+        </p>
+        <p class="ednote">
+          The profiling mechanism currently specified in [[WOT-PROFILES]] does not exactly follow the above-mentioned
```
danielpeintner, line 4544: [[WOT-PROFILES]] cannot be found by ReSpec

index.template.html
```diff
@@ -4531,6 +4531,28 @@ <h5>Protocol and Payload Bindings</h5>
...
+          mechanism. However, there is agreement on the future of the profiling mechanism. Once the profiling mechanism
+          is updated, this section will be updated accordingly.
+        </p>
+        <p>
+          For example, a profile can constrain the usage of HTTP GET Method for <code>readproperty</code> operations and
+          require payloads to be always serialized in JSON. The usage of a profile is indicated by the
+          <code>profile</code> term in the TD, which is a URI that points to the profile definition. How a Consumer can
+          use a profile is explained in the [[WOT-PROFILES]] specification under the section
```
danielpeintner: Again, [[WOT-PROFILES]] cannot be found by ReSpec

index.template.html
egekorkan: I have fixed the wrong usage of PROFILE. It should have been PROFILES all along.

egekorkan:
@danielpeintner I have applied your feedback and fixed the S usage. I accidently committed the images, that is why the revert commits.

index.html
```diff
@@ -7530,21 +7530,22 @@ <h5>Protocol and Payload Bindings</h5>
...
         <h4>Relationship to Profiles</h4>
         <p>
-          The Web of Things also define a profiling mechanism in [[WOT-PROFILES]] that allows the reduction of
-          implementation effort. Profiles are built on top of existing Bindings by constraining the flexibility of TDs.
-          Such constraints can be bindings, semantic contexts, link relations, security schemes and discovery
-          mechanisms. As there is less flexibility, the implementation effort to achieve out-of-the-box interoperability
-          is reduced.
+          The Web of Things also defines a profiling mechanism that allows the reduction of implementation effort.
+          Profiles are built on top of existing Bindings by constraining the flexibility of TDs. Such constraints can be
+          bindings, semantic contexts, link relations, security schemes and discovery mechanisms. With these
```
```suggestion
          bindings, semantic contexts, link relations, security schemes, and discovery mechanisms. With these
```

index.html
```diff
@@ -7530,21 +7530,22 @@ <h5>Protocol and Payload Bindings</h5>
...
           require payloads to be always serialized in JSON. The usage of a profile is indicated by the
-          <code>profile</code> term in the TD, which is a URI that points to the profile definition.
-          <span class="rfc2119-assertion" id="td-consumer-profile-ignore">
-            A Consumer MAY ignore the <code>profile</code> term and use the TD as it is, but the knowledge of a profile
-            allows the Consumer to implement the profile-specific constraints as assumptions and thus reduce the
-            implementation effort on understanding the TD.
-          </span>
+          <code>profile</code> term in the TD, which is a URI that points to the profile definition. How a Consumer can
+          use a profile is explained in the [[WOT-PROFILES]] specification under the section
+          <emph>Profiling Mechanism</emph>.
```
```suggestion
          use a profile is explained in the <emph>Profiling Mechanism</emph> section of
          the [[WOT-PROFILES]] specification.
```

index.template.html
```diff
@@ -4535,21 +4535,22 @@ <h5>Protocol and Payload Bindings</h5>
...
           require payloads to be always serialized in JSON. The usage of a profile is indicated by the
-          <code>profile</code> term in the TD, which is a URI that points to the profile definition.
-          <span class="rfc2119-assertion" id="td-consumer-profile-ignore">
-            A Consumer MAY ignore the <code>profile</code> term and use the TD as it is, but the knowledge of a profile
-            allows the Consumer to implement the profile-specific constraints as assumptions and thus reduce the
-            implementation effort on understanding the TD.
-          </span>
+          <code>profile</code> term in the TD, which is a URI that points to the profile definition. How a Consumer can
+          use a profile is explained in the [[WOT-PROFILES]] specification under the section
+          <emph>Profiling Mechanism</emph>.
```
TallTed:
```suggestion
          use a profile is explained in the <emph>Profiling Mechanism</emph> section of the
          [[WOT-PROFILES]] specification.
```

TallTed (changes requested):
Small language tweaks to better fit common English expression

index.html
egekorkan: done in template

index.html
egekorkan: done in template

relu91 (approved):

PR #2127, opened 2025-08-28
Moving Sebastian and McCool to formerEditors

egekorkan:
This was confirmed individually with both of them


egekorkan:
Merging in the TD call

Issue #2126, opened 2025-08-27
Categorizing Architecture Assertions


egekorkan:
We should go over https://github.com/w3c/wot-thing-description/blob/main/planning/work-items/other-documents-sync/arch.md and add the category labels. 

- [x] 1-9  : @k-toumura 
- [x] 10-18: @egekorkan 
- [x] 19-27: @JKRhb 
- [x] 28-36: @mjkoster 
- [x] 37-45: @danielpeintner 
- [x] 46-54: @egekorkan

PR #2123, opened 2025-08-21
Removing HTTP Binding

egekorkan:
## Description of Changes

This is removing the HTTP Binding from the TD spec. A corresponding PR is being worked on for the HTTP binding in the Binding Templates repo. It is normative as I am moving some HTTP-specific assertions to the whole spec, which makes sense for me at least. With this, I am seeing yet another overlap of the binding-related chapters. In behavior assertions there are binding assertions and then we have the binding section too. I will handle those in a separate PR.

## Related Issue

Closes #1259

## Type of Change

- [Normative](https://github.com/w3c/wot/blob/main/policies/async-decision.md#editorial-non-normative-changes) with label Normative Label

PR #2129, opened 2025-08-28
Add categorization for TD-related Architecture assertions 19 to 27

JKRhb:
## Description of Changes

This PR contains my first attempt at categorizing the TD-related assertions from the architecture document that have been assigned to me via #2126.

## Related Issue

#2126

### DIFF
--- a/planning/work-items/other-documents-sync/arch.md
+++ b/planning/work-items/other-documents-sync/arch.md
@@ -7,7 +7,7 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
 ## Categorization of assertions (tentative)
 
 - A: Those related to the Thing Description document
-  - A1: Assertions that are already included in the Thing Description document. No changes to the Thing Description document are necessary. 
+  - A1: Assertions that are already included in the Thing Description document. No changes to the Thing Description document are necessary.
   - A2: Assertions that differ from what is written in the Thing Description document. If necessary, update the Thing Description document to ensure consistency.
   - A3: Assertions not included in the Thing Description document. Add them if necessary.
 - B: Those related to other documents
@@ -23,6 +23,7 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
 ##### 6.1.1.1 Thing Descriptions
 
 - [ ] 1. In W3C WoT, the description metadata for a Thing instance MUST be available as a WoT Thing Description (TD) [WOT-THING-DESCRIPTION]. ([arch-td-metadata](https://www.w3.org/TR/wot-architecture11/#arch-td-metadata))
+
   - Category: A1 / A2 / A3 / B / C (example)
 
 - [ ] 2. There MAY be other representations of a Thing such as an HTML-based user interface, simply an image of the physical entity, or even non-Web representations in closed systems. ([arch-other-thing-representations](https://www.w3.org/TR/wot-architecture11/#arch-other-thing-representations))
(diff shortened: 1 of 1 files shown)


planning/work-items/other-documents-sync/arch.md
```diff
@@ -75,39 +75,48 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
 
 - [ ] 18. Form fields are optional and MAY further specify the expected request message for the given operation. ([arch-op-expected-request](https://www.w3.org/TR/wot-architecture11/#arch-op-expected-request))
 
-- [ ] 19. Form fields MAY depend on the protocol used for the submission target as specified in the URI scheme. ([arch-op-form-fields-protocol](https://www.w3.org/TR/wot-architecture11/#arch-op-form-fields-protocol))
+- [x] 19. Form fields MAY depend on the protocol used for the submission target as specified in the URI scheme. ([arch-op-form-fields-protocol](https://www.w3.org/TR/wot-architecture11/#arch-op-form-fields-protocol))
```
egekorkan: We have a race condition ;) See first point at the end of https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/2123.html#using-a-thing-description-with-a-protocol-binding

planning/work-items/other-documents-sync/arch.md
```diff
@@ -75,39 +75,48 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
...
 ### 6.7 Protocol Bindings
 
 #### 6.7.1 Hypermedia-driven
 
-- [ ] 20. Interaction Affordances MUST include one or more Protocol Bindings. ([arch-hypermedia](https://www.w3.org/TR/wot-architecture11/#arch-hypermedia))
+- [x] 20. Interaction Affordances MUST include one or more Protocol Bindings. ([arch-hypermedia](https://www.w3.org/TR/wot-architecture11/#arch-hypermedia))
+    - A3
 
-- [ ] 21. Protocol Bindings MUST be serialized as hypermedia controls to be self-descriptive on how to activate the Interaction Affordance. ([arch-hypermedia-protocol-binding](https://www.w3.org/TR/wot-architecture11/#arch-hypermedia-protocol-binding))
+- [x] 21. Protocol Bindings MUST be serialized as hypermedia controls to be self-descriptive on how to activate the Interaction Affordance. ([arch-hypermedia-protocol-binding](https://www.w3.org/TR/wot-architecture11/#arch-hypermedia-protocol-binding))
```
egekorkan: Same as above but point 2. both can be reworked to give the exact same meaning

planning/work-items/other-documents-sync/arch.md
```diff
@@ -75,39 +75,48 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
...
+    - A3
 
-- [ ] 22. The hypermedia controls MAY be cached outside the Thing and used for offline processing if caching metadata is available to determine the freshness. ([arch-hypermedia-caching](https://www.w3.org/TR/wot-architecture11/#arch-hypermedia-caching))
+- [x] 22. The hypermedia controls MAY be cached outside the Thing and used for offline processing if caching metadata is available to determine the freshness. ([arch-hypermedia-caching](https://www.w3.org/TR/wot-architecture11/#arch-hypermedia-caching))
```
egekorkan: Not sure if I understand this one

egekorkan (changes requested):
Could you add a note about my two comments that we should sync when that linked PR is merged?

planning/work-items/other-documents-sync/arch.md
JKRhb: Yeah, this one is a bit difficult to parse :/

I guess the intention here is simply that if you have a protocol that allows for caching (and determining whether the cache entry is still fresh) you can ... use the data from the cache?

What confuses me a bit is that this assertion talks about caching the _hypermedia controls_ (i.e., the form itself if I see it correctly) which does not make that much sense to me. So maybe we can simply discard this assertion.

JKRhb:
> Could you add a note about my two comments that we should sync when that linked PR is merged?

Sure, will do :) (You mean #2123, right?) Thank you for the review!

planning/work-items/other-documents-sync/arch.md
egekorkan: 1. Approach 1: Trying harder to understand :D So I understand this in two directions which depends on what "offline processing" refers to. 

    1. It can refer to offline processing of hypermedia controls, i.e. forms, when the Thing is not available. I can for example create (process the hypermedia controls) a script that uses them for an application. Here, caching should mean that the hypermedia controls do not change for a while, e.g., the resource endpoints do not change. In this case, we are talking about very dynamic Things.
    2. It can refer to the data. E.g., If I know that the temperature does not change for an hour, the Consumer or an intermediary can cache this value and use it an application logic.

I think that the first one makes more sense but this is a case we very rarely see. It can make sense in manageable affordances though.

2. Approach 2: Understanding why it is added:

Found it being added at https://github.com/w3c/wot-architecture/commit/7954c318f9b0baef68efd991da11bfc66a97bfc6 but it says just a split but it seems it is adding assertions. 

Approach 2 seems to fail here

**Proposal**: I agree with `So maybe we can simply discard this assertion.` if we don't understand it ourselves as implementers after this long.

planning/work-items/other-documents-sync/arch.md
```diff
@@ -75,39 +75,48 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
...
+    - A3
 
 ### 6.8 Media Types
 
-- [ ] 23. All data (a.k.a. content) exchanged when activating Interaction Affordances MUST be identified by a media type [RFC2046] in the Protocol Binding. ([arch-media-type](https://www.w3.org/TR/wot-architecture11/#arch-media-type))
+- [x] 23. All data (a.k.a. content) exchanged when activating Interaction Affordances MUST be identified by a media type [RFC2046] in the Protocol Binding. ([arch-media-type](https://www.w3.org/TR/wot-architecture11/#arch-media-type))
```
TallTed:
```suggestion
- [x] 23. All data (also known as <em>content</em>) exchanged when activating Interaction Affordances MUST be identified by a media type [RFC2046] in the Protocol Binding. ([arch-media-type](https://www.w3.org/TR/wot-architecture11/#arch-media-type))
```

JKRhb: Thank you, @TallTed, but I think the assertions are copied over from the architecture document and should remain as they are (for now).

planning/work-items/other-documents-sync/arch.md
JKRhb: > Found it being added at [w3c/wot-architecture@7954c31](https://github.com/w3c/wot-architecture/commit/7954c318f9b0baef68efd991da11bfc66a97bfc6) but it says just a split but it seems it is adding assertions.

Thank you for your researching this, @egekorkan! I am wondering then whether @mkovatsc could help us understand this assertion? Otherwise, I think discarding it should do no harm.

Maybe we could also list/add the use cases you've mentioned somewhere and revisit this topic later, or rather come up with more fitting assertions if appropriate.

k-toumura: It seems to refer to cases where the Thing Description (specifically the forms) changes dynamically according to the latest state of the Thing (such as its IP address). And I believe the (optional) modified attribute in the TD is provided so that, when cached, one can tell whether the TD is up to date (of course, it’s still necessary to implicitly decide on a TTL somewhere...). 
Since the mechanism is covered by the TD document, how about classifying it as category `A1`?

k-toumura (approved):

planning/work-items/other-documents-sync/arch.md
JKRhb: Thank you, @k-toumura! A1 would sound good to me then :)

planning/work-items/other-documents-sync/arch.md
egekorkan: That PR is merged, can be A1.

planning/work-items/other-documents-sync/arch.md
egekorkan: 20 and 21 are in the TD spec now, A1

planning/work-items/other-documents-sync/arch.md
egekorkan: We talked about it in the TD call and the TF agrees to A1.

PR #2147, opened 2025-09-25
Add categorization for TD-related Architecture assertions 37 to 45

danielpeintner:
## Description of Changes

Categorizes assertions 37 to 45 in the Architecture document.
TD Spec has a [Security Considerations section](https://w3c.github.io/wot-thing-description/#sec-security-consideration), but I think those assertions all belong to the [WoT Security and Privacy Guidelines](https://w3c.github.io/wot-security/) document.

## Related Issue

https://github.com/w3c/wot-thing-description/issues/2126

## Type of Change

[Editorial (non-normative)](https://github.com/w3c/wot/blob/main/policies/async-decision.md#editorial-non-normative-changes) with label Editorial Label.


k-toumura (approved):
I agree with categorizing assertions 37 to 40 as category B.
There’s room for discussion on where and to what extent the WoT standard should define how Things are implemented and how Things are exposed on the network.

egekorkan:
> There’s room for discussion on where and to what extent the WoT standard should define how Things are implemented and how Things are exposed on the network.

This is discussed today. The TF thinks that this goes beyond W3C. We are not sure if this can be normative in any way. Even making informative guidelines requires people with experience in the domain.

PR #2149, opened 2025-09-30
Arch Assertion Categories

egekorkan:
## Description of Changes

This contains the categories for my and @mjkoster assertions. @relu91 and @danielpeintner there are some assertions related to "WoT Runtime", which is not necessarily the Scripting API. Feel free to comment on those.

## Related Issue

part of #2126 

## Type of Change

### DIFF
--- a/planning/work-items/other-documents-sync/arch.md
+++ b/planning/work-items/other-documents-sync/arch.md
@@ -111,6 +111,7 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
 - [ ] 27. There SHOULD be a strict separation of Public Security Metadata and Private Security Data. ([arch-security-consideration-separate-security-data](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-separate-security-data))
 
 - [ ] 28. Authentication and authorization SHOULD be established based on separately managed Private Security Data. ([arch-security-consideration-auth-private-data](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-auth-private-data))
+
   - A3: Fits to https://w3c.github.io/wot-thing-description/#behavior-security
 
 - [ ] 29. Producers of TDs MUST ensure that no Private Security Data is included in TDs. ([arch-security-consideration-no-private-security-data](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-no-private-security-data))
@@ -120,6 +121,7 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
 #### 10.1.2 Thing Description Communication Metadata Risk
 
 - [ ] 30. Whenever possible, TD creators SHOULD use the vetted communication metadata provided in the WoT Binding Templates. ([arch-security-consideration-communication-binding](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-communication-binding))
+
   - D (also the meaning is not clear, which is a sign for not making it normative)
 
 - [ ] 31. When generating TDs for an IoT Platform not covered by the WoT Binding Templates, TD creators SHOULD ensure that all the security requirements of the IoT Platform are satisfied. ([arch-security-consideration-communication-platform](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-communication-platform))
@@ -130,6 +132,7 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
 #### 10.2.1 Cross-Script Security Risk
 
 - [ ] 32. The WoT Runtime SHOULD perform isolation of script instances and their data from each other in cases when scripts handle sensitive data. ([arch-security-consideration-isolation-sensitive](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-isolation-sensitive))
+
   - D (can also be B and in Scripting API but this is not limited to scripting api implementations)
 
 - [ ] 33. Similarly, the WoT Runtime implementation SHOULD perform isolation of WoT Runtime instances and their data from each other if a WoT device has more than one tenant. ([arch-security-consideration-isolation-tenants](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-isolation-tenants))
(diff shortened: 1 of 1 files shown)


k-toumura (approved):
Regarding assertion 48: Since QUIC can be used as a secure transport over UDP, we might need to rephrase this assertion.

planning/work-items/other-documents-sync/arch.md
```diff
@@ -112,33 +112,47 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
...
+  - D (can also be B and in Scripting API but this is not limited to scripting api implementations)
 
 #### 10.2.2 Physical Device Direct Access Risk
 
 - [ ] 34. The WoT Runtime SHOULD NOT directly expose low-level device hardware interfaces to the script developers. ([arch-security-consideration-avoid-direct](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-avoid-direct))
 
+  - D (can also be B and in Scripting API but this is not limited to scripting api implementations)
+
 - [ ] 35. A WoT Runtime implementation SHOULD provide a hardware abstraction layer for accessing the low-level device hardware interfaces. ([arch-security-consideration-use-hal](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-use-hal))
+  - D (can also be B and in Scripting API but this is not limited to scripting api implementations)
```
danielpeintner, line 148: Categorization is fine.

BTW, I don't really see any difference to 34 🤷‍♂️

planning/work-items/other-documents-sync/arch.md
```diff
@@ -166,26 +180,41 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
 
 - [ ] 46. When secure transport over TCP is appropriate, then at least TLS 1.3 [RFC8446] SHOULD be used. ([arch-security-consideration-tls-1-3](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-tls-1-3))
 
+  - A3: This should be a generic assertion to all bindings that can support TCP. The implementation enforcement can only happen in the binding, which is informative. However, we should recommend all implementers to do that. In the binding documents, there should be informative notes about this that can point back to TD.
+
 - [ ] 47. If TLS 1.3 cannot be used for compatibility reasons but secure transport over TCP is appropriate, TLS 1.2 [RFC5246] MAY be used. ([arch-security-consideration-tls-1-2](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-tls-1-2))
 
+  - A3: This should be a generic assertion to all bindings that can support TCP. The implementation enforcement can only happen in the binding, which is informative. However, we should recommend all implementers to do that. In the binding documents, there should be informative notes about this that can point back to TD.
```
danielpeintner, line 188: Maybe it is just me, but I think we should state that the most secure transport should be used unless not possible. Specifying 1.3 while it might be possible that in the near future 1.4 is released does not make sense to me.

danielpeintner (approved):
LGTM.

I have 2 questions:
* shall we really state various "secure" versions while they might get out of date?
*  what does it mean if we tick or not tick the checkboxes?

planning/work-items/other-documents-sync/arch.md
egekorkan: I agree. We can put an informative note after the assertion saying something like "At the time of writing, this is TLS and DTLS 1.3, but please check the most recent version"

planning/work-items/other-documents-sync/arch.md
egekorkan: I would say that 35 is what you should do to avoid 34. 35 is more of a guideline so we should keep that.

Note: Technically by not exposing anything device-related, one would achieve 34. Like adding two numbers that happen purely in software.

planning/work-items/other-documents-sync/arch.md
```diff
@@ -166,26 +180,41 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
...
+
 - [ ] 48. If DTLS 1.3 cannot be used for compatibility reasons but secure transport over UDP is appropriate, then DTLS 1.2 [RFC6347] MAY be used. ([arch-security-consideration-dtls-1-2](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-dtls-1-2))
 
+  - A3: This should be a generic assertion to all bindings that can support TCP. The implementation enforcement can only happen in the binding, which is informative. However, we should recommend all implementers to do that. In the binding documents, there should be informative notes about this that can point back to TD.
```
egekorkan, line 192:
```suggestion
  - A3: This should be a generic assertion to all bindings that can support TCP. The implementation enforcement can only happen in the binding, which is informative. However, we should recommend all implementers to do that. In the binding documents, there should be informative notes about this that can point back to TD.
  - Note: Since QUIC can be used as a secure transport over UDP, we need to rephrase this assertion, i.e., UDP doesn't imply DTLS.
```

planning/work-items/other-documents-sync/arch.md
```diff
@@ -166,26 +180,41 @@ Continuation of https://github.com/w3c/wot-thing-description/issues/2120
 
 - [ ] 46. When secure transport over TCP is appropriate, then at least TLS 1.3 [RFC8446] SHOULD be used. ([arch-security-consideration-tls-1-3](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-tls-1-3))
 
+  - A3: This should be a generic assertion to all bindings that can support TCP. The implementation enforcement can only happen in the binding, which is informative. However, we should recommend all implementers to do that. In the binding documents, there should be informative notes about this that can point back to TD.
```
egekorkan, line 183:
```suggestion
  - A3: This should be a generic assertion to all bindings that can support TCP. The implementation enforcement can only happen in the binding, which is informative. However, we should recommend all implementers to do that. In the binding documents, there should be informative notes about this that can point back to TD.
  - Note: the same concern seem to apply to OPC UA, which can use other mechanisms that TLS for security. We need to change the 4 following assertions to not constrain to a specific transport layer security.
```

PR #2154, opened 2025-10-09
Common Definitions Examples

egekorkan:
I wanted to add more use cases driven examples. More coming as seen in comments

### DIFF
--- a/proposals/common-definitions/tooling/tds.js
+++ b/proposals/common-definitions/tooling/tds.js
@@ -931,7 +931,7 @@ const recommendedTDs = [
         ]
       }
     },
-    "actions":{
+    "actions": {
       "act1": {
         "input": { "type": "string" },
         "forms": [
@@ -953,7 +953,7 @@ const recommendedTDs = [
         ]
       }
     }
-  },
+  }
 ];
 
 module.exports = {


proposals/common-definitions/tooling/tds.js
```diff
@@ -447,9 +449,122 @@ const invalidExpandedTDs = [
...
+            "href": "/props/prop2"
+          }
+        ]
+      }
+    }
+  }
+  // multi protocols (http and coap)
+  // readproperty and writeproperty defaults (GET and POST)
+  // read and write requring different security for all operations
+  // modbus with all parameters in connection
```
egekorkan:
```suggestion
  // modbus with all parameters in connection (one property not same endianness)
  // mqtt with qos 0 for all topics except some
  // default content type not being json
  // multiple content types
```

proposals/common-definitions/tooling/tds.js
```diff
@@ -447,9 +449,438 @@ const invalidExpandedTDs = [
...
+  // there is no default connection/protocol.
+  // both use the same security but different content type
+  // if they were both using the same content type, we could have used two connection definitions, like in the example above
+  {
+    "@context": "https://www.w3.org/ns/wot-next/td",
+    "title": "recommended-test-multi-protocol",
+    "formDefinitions": {
+      "http": {
+        "base": "https://192.168.1.10:8080",
+        "contentType": "application/json"
```
danielpeintner, line 685: Note: For clarity, it makes sense to put it. Otherwise, JSON is still implied.
Not sure if we want to keep it or remove it. Both is fine 🤷‍♂️

proposals/common-definitions/tooling/tds.js
```diff
@@ -447,9 +449,438 @@ const invalidExpandedTDs = [
...
+    "@context": "https://www.w3.org/ns/wot-next/td",
+    "title": "recommended-test-multi-protocol",
+    "formDefinitions": {
+      "json": {
+        "contentType": "application/json"
+      },
+      "cbor": {
+        "contentType": "application/cbor"
+      }
+    },
```
danielpeintner, line 870: I am not sure if I would label this TD as "recommended" since it make the TD even more verbose.--

egekorkan: Technically less bytes but more keys

egekorkan: And also we need to give guidance on how to semantically express this. Putting this into forms might be more efficient in some cases but still makes it a lot of work to parse what codecs are needed to communicate with the Thing.

Note: I also don't find our proposal very elegant at the moment

proposals/common-definitions/tooling/tds.js
egekorkan: Now we don't have a default value for it anymore so we have to put it here

lu-zero:
I feel we could have a container called "default" for the default form/connection/etc even if it makes probably the whole thing a little more verbose.

egekorkan:
> Since a single network device can have multiple IP addresses, it is necessary to consider its behavior.
> 
> I think it is necessary to use UseCase for this consideration.

@chachamimm we have the user stories at https://github.com/w3c/wot-thing-description/blob/main/proposals/common-definitions/README.md#user-stories . Do you have something else in mind? In any case, we can discuss it today

proposals/common-definitions/tooling/tds.js
```diff
@@ -447,9 +449,517 @@ const invalidExpandedTDs = [
...
+  },
+  // 4. Multi IP addresses (IPv4 and IPv6)
+  // Use case: Device with two IP assignments
+  // notes:
+  // there is no default connection/ip format.
+  // both use the same security and content type
+  // there is no advantage of the new mechanism much other than reducing TD length. The forms have the same amount of terms in the end.
+  // However, this is a recommended way to use multiple IP addresses
+  // without duplicating security and content type in each form
+  // and also without duplicating base in each form (as done in the past)
```
TallTed, line 575:
```suggestion
  // Notes:
  // - There is no default connection/ip format.
  // - Both use the same security and content type.
  // - There is not much advantage of the new mechanism other than reducing
  //   TD length. The forms have the same amount of terms in the end.
  // - However, this is a recommended way to use multiple IP addresses
  //   without duplicating security and content type in each form, and
  //   also without duplicating base in each form (as done in the past).
```

proposals/common-definitions/tooling/tds.js
```diff
@@ -447,9 +449,517 @@ const invalidExpandedTDs = [
...
+        ]
+      }
+    }
+  },
+  // 6. multi protocols: http with json and coap with cbor
+  // Use case: Device supporting both coap and http
+  // notes:
+  // there is no default connection/protocol.
+  // both use the same security but different content type
+  // if they were both using the same content type, we could have used two connection definitions, like in the example above
```
TallTed, line 678:
```suggestion
  // Notes:
  // - There is no default connection/protocol.
  // - Both use the same security but different content type.
  // - If they were both using the same content type, we could have
  //   used two connection definitions, like in the example above.
```

PR #2158, opened 2025-10-23
fix: syntax error in wotsec: \#

fennibay:
## Description of Changes

Apparently \\# is not a valid escape sequence and some RDF parsers complain about it. It's not needed anyway.

## Related Issue

no issues

## Type of Change

/cc @ektrah


sebastiankb:
maybe it needs to be percent-encoded?

PR #2163, opened 2025-11-10
Array instead of string for common definition references

### LINKED ISSUE #2159: Adding array for common definitions referencing
As a result of the discussion at #2154 , we have decided that the new keywords like `connection`, `form`, etc. cannot be a string type anymore. The TDs will have either an object for an in-place definition of the definition or will have an array which lists all the possible definitions that are applied. Thus we can have examples like the following:

```js
  {
    "@context": "https://www.w3.org/ns/wot-next/td",
    "connectionDefinitions": {
      "ipv4": {
        "base": "https://192.168.1.10:8080"
      },
      "ipv6": {
        "base": "https://[2001:db8:85a3::8a2e:370:7334]:8080"
      }
    },
    "connection": ["ipv4", "ipv6"],
    "form": {
      "contentType": "application/json"
    },
    "security": {
      "scheme": "nosec"
    },
    "properties": {
      "prop1": {
        "type": "string",
        "forms": [
          {
            "href": "/props/prop1"
          }
        ]
      },
      "prop2": {
        "type": "string",
        "forms": [
          {
            "href": "/props/prop2"
          }
        ]
      }
    }
  }
```

Internally, that `prop1` would expand to something like the following:

```js
      "prop1": {
        "type": "string",
        "forms": [
          {
            "href": "https://192.168.1.10:8080/props/prop1",
            "security": {
              "scheme": "nosec"
            },
            "contentType": "application/json"
          },
          {
            "href": "https://[2001:db8:85a3::8a2e:370:7334]:8080/props/prop1",
            "security": {
              "scheme": "nosec"
            },
            "contentType": "application/json"
          }
        ]
      }

As a result, the proposal and the examples in there should be updated in a PR.

### PR DESCRIPTION
**Pull request by egekorkan**
## Description of Changes

As discussed in the previous PR of #2154, we have seen that an array with OR relation would solve some of the problems. I am adding github comments on sections that do not look good in the end.

In my opinion, this is a definite improvement that covers the most basic and common use case without any changes and the other use cases are also easier to cover too.

## Related Issue

Closes #2159

### CODE REVIEW

File: proposals/common-definitions/tooling/tds.js [hunk 1]
```diff
@@ -163,9 +163,9 @@ const validCompactTDs = [
         "type": "string",
         "forms": [
           {
-            "connection": "conn1",
-            "form": "form1",
-            "security": "sec1",
+            "connection": ["conn1"],
+            "form": ["form1"],
+            "security": ["sec1"],
```
egekorkan, line 168: referencing in the forms with array is a bit weird since you cannot have OR relationship here anymore, it happens in the array just above (forms). We can say that these terms MUST be string in the form level? Not sure since we start changing the type based on the level etc.

proposals/common-definitions/tooling/tds.js
```diff
@@ -595,24 +596,18 @@ const recommendedTDs = [
...
-          {
-            "connection": "ipv6",
             "href": "/props/prop1"
           }
         ]
       },
       "prop2": {
         "type": "string",
         "forms": [
+          // TODO: leaving it here for now to give an idea. It would look weird but keep more consistency
```
egekorkan, line 677: This is a possible (but not nice!) solution to @relu91 comment at #2159

egekorkan:
One use case we have in node-wot is that the user can add http and coap, also json and cbor. Meaning there are 4 ways to interact per affordance. That would be a `connection` with 2 items and `form` with 2 items.

proposals/common-definitions/tooling/tds.js
egekorkan: Or the top level ones are called `defaultForms` and `defaultConnections`, whereas the form level ones are without the default prefix and only accept strings

proposals/common-definitions/tooling/tds.js
```diff
@@ -235,7 +235,7 @@ const validCompactTDs = [
         "type": "string",
         "forms": [
           {
-            "href": "https://example.com/props/prop1",
+            "href": "https://example.comprops/prop1",
```
TallTed: `"href": "https://example.comprops/prop1"` doesn't seem likely to be correct anywhere, even if `"href": "props/prop2"` is correct wherever it now occurs.

egekorkan: I have fixed all occurances. Thank you for spotting this!

TallTed: Happy to help. Please include a `coauthored by` statement (as automatically happens if you "apply suggestion" or "add to batch") on future commits to which I contribute, so my activity shows up more accurately. :-)

PR #2172, opened 2026-01-05
IRI to URI change

### LINKED ISSUE #2156: Reference to the URI Spec (RFC3986) should be replaced with IRI Spec (RFC3987)
At base term under [Thing](https://w3c.github.io/wot-thing-description/#thing) information model we reference to RFC3986. However, the IRI spec is behind the anyUri term. We should change that reference.

egekorkan:
## Description of Changes

I am changing based on the result at https://github.com/w3c/wot-thing-description/issues/2156#issuecomment-3607420631. However, the type is still anyUri, which is not correct (draft until that is solved?). In the meantime, I am also doing a fix to the link's href, which also mentions a form.

## Related Issue

Closes #2156 

## Type of Change

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/2172.html" title="Last updated on Jan 28, 2026, 3:04 PM UTC (fab2b6b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/2172/782c750...fab2b6b.html" title="Last updated on Jan 28, 2026, 3:04 PM UTC (fab2b6b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -760,13 +760,17 @@
           <dfn id="dfn-absolute-uri">Absolute URI</dfn>
         </dt>
         <dd>
-          A URI, as defined by [[RFC3986]], section 5.1. It can be found in the <code>base</code> member or <code>href</code> member of a <a href="#form">Form</a> in a Thing Description or obtained by resolving a <a>Relative URI</a> against a <a>Base URI</a>.
+          A URI, as defined by [[RFC3986]], section 5.1. It can be found in the <code>base</code> member or
+          <code>href</code> member of a <a href="#form">Form</a> in a Thing Description or obtained by resolving a
+          <a>Relative URI</a> against a <a>Base URI</a>.
         </dd>
         <dt><dfn id="dfn-target-uri">Target URI</dfn></dt>
         <dd>
-          A URI conforming to the absolute URI definition by [[RFC3986]], section 4.3, that locates a resource on the network
-          and can be directly used to send the correct protocol message. It can be found in the <code>href</code> member of a
-          <a href="#form">Form</a> in a Thing Description or obtained by resolving a <a>Relative URI</a> against a <a>Base URI</a>. A <a>Base URI</a> on its own is not a valid <a>Target URI</a> as it may not point to a specific resource.
+          A URI conforming to the absolute URI definition by [[RFC3986]], section 4.3, that locates a resource on the
+          network and can be directly used to send the correct protocol message. It can be found in the
+          <code>href</code> member of a <a href="#form">Form</a> in a Thing Description or obtained by resolving a
+          <a>Relative URI</a> against a <a>Base URI</a>. A <a>Base URI</a> on its own is not a valid
+          <a>Target URI</a> as it may not point to a specific resource.
         </dd>
         <dt>
           <dfn id="dfn-inf-model">TD Information Model</dfn>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -3677,7 +3677,10 @@ <h3><code>Form</code></h3>
               <tbody>
                 <tr class="rfc2119-table-assertion" id="td-vocab-href--Form">
                   <td><code>href</code></td>
-                  <td>Target IRI of a link or submission target of a form.</td>
+                  <td>
+                    Target URI[[RFC3986]] of a submission target of a <a>Form</a>. If relative, it is combined with the
```
```suggestion
                    Target URI [[RFC3986]] of a submission target of a <a>Form</a>. If relative, it is combined with the
```

index.html
```diff
@@ -3812,6 +3815,16 @@ <h3><code>Form</code></h3>
               </tbody>
             </table>
 
+            <p class="note">
+              Please note that the value of <code>href</code> is of of type URI[[RFC3986]] and what is sent as a
+              protocol message is protocol-specific and can be decoded from <code>href</code>.
+              <span class="rfc2119-assertion" id="td-form-href-unicode-usage">
+                In cases of protocols where Universal Character Set (Unicode) [[ISO10646]] characters are to be sent in
+                the protocol message, the resource MAY be represented internally as an IRI[[!RFC3987]] and can be
```
```suggestion
              Please note that the value of <code>href</code> is of of type URI [[RFC3986]] and what is sent as a
              protocol message is protocol-specific and can be decoded from <code>href</code>.
              <span class="rfc2119-assertion" id="td-form-href-unicode-usage">
                When Universal Character Set (Unicode) [[ISO10646]] characters are to be sent in
                the protocol message, the resource MAY be represented internally as an IRI [[!RFC3987]] and can be
```

ontology/hctl.html
```diff
@@ -148,27 +148,27 @@ <h2>Terminology</h2>
...
-      are possible, e.g. for error reporting.  Each additional response needs to be 
+      are possible, e.g. for error reporting.  Each additional response needs to be
```
TallTed:
```suggestion
      are possible, e.g., for error reporting.  Each additional response needs to be
```

ontology/hctl.html
```diff
@@ -148,27 +148,27 @@ <h2>Terminology</h2>
...
-<section id="hasTarget"><h4>hasTarget</h4><p>IRI: <code>https://www.w3.org/2019/wot/hypermedia#hasTarget</code></p><span>target IRI of a link or submission target of a form.</span><table class="def numbered"><tbody><tr><td>Domain includes</td><td><code><a href="#Form">hctl:Form</a></code><br><code><a href="#Link">hctl:Link</a></code></td></tr></tbody></table></section>
+<section id="hasTarget"><h4>hasTarget</h4><p>IRI: <code>https://www.w3.org/2019/wot/hypermedia#hasTarget</code></p><span>Target URI[[RFC3986]] of a submission target of a <a>Form</a>. If relative, it is combined with the value in the <code>base</code> in Thing-level to create an absolute URI.</span><table class="def numbered"><tbody><tr><td>Domain includes</td><td><code><a href="#Form">hctl:Form</a></code><br><code><a href="#Link">hctl:Link</a></code></td></tr></tbody></table></section>
```
TallTed:
```suggestion
<section id="hasTarget"><h4>hasTarget</h4><p>IRI: <code>https://www.w3.org/2019/wot/hypermedia#hasTarget</code></p><span>Target URI [[RFC3986]] of a submission target of a <a>Form</a>. If relative, it is combined with the value in the <code>base</code> in Thing-level to create an absolute URI.</span><table class="def numbered"><tbody><tr><td>Domain includes</td><td><code><a href="#Form">hctl:Form</a></code><br><code><a href="#Link">hctl:Link</a></code></td></tr></tbody></table></section>
```

ontology/hctl.ttl
```diff
@@ -81,7 +81,7 @@ For example, for HTTP and Events, it indicates which of several available mechan
     schema:domainIncludes :Link .
 
 :hasTarget rdf:type owl:DatatypeProperty ;
-    rdfs:comment "target IRI of a link or submission target of a form." ;
+    rdfs:comment "Target URI[[RFC3986]] of a submission target of a <a>Form</a>. If relative, it is combined with the value in the <code>base</code> in Thing-level to create an absolute URI." ;
```
TallTed:
```suggestion
    rdfs:comment "Target URI [[RFC3986]] of a submission target of a <a>Form</a>. If relative, it is combined with the value in the <code>base</code> in Thing-level to create an absolute URI." ;
```

testing/assertions.csv
```diff
@@ -130,6 +130,7 @@
 "td-expectedResponse-missing-contentType","null","If there is no `contentType` value defined in a `response` object within a form, a Consumer MUST expect that the respective response does not contain any payload."
 "td-form-additional-mediatype-metadata","null",", Bindings MAY have additional information that specifies representation formats in more detail than the media type alone."
 "td-form-context-submission-target","null","As a Form but not just the href defines how to execute the interaction, href of multiple Forms MAY point to the same resource, where the different Forms represent different operations on that resource."
+"td-form-href-unicode-usage","null","In cases of protocols where Universal Character Set (Unicode) [[ISO10646]] characters are to be sent in the protocol message, the resource MAY be represented internally as an IRI[[!RFC3987]] and can be converted to a URI following Section 3.1 of [[RFC3987]] before placing it in the href."
```
TallTed:
```suggestion
"td-form-href-unicode-usage","null","In cases of protocols where Universal Character Set (Unicode) [[ISO10646]] characters are to be sent in the protocol message, the resource MAY be represented internally as an IRI [[!RFC3987]] and can be converted to a URI following Section 3.1 of [[RFC3987]] before placing it in the href."
```

testing/assertions.csv
```diff
@@ -310,8 +311,8 @@
 "td-vocab-format--DataSchema","null","formatAllows validation based on a format pattern such as ""date-time"", ""email"", ""uri"", etc. (Also see below.)optionalstring"
 "td-vocab-forms--InteractionAffordance","null","formsSet of form hypermedia controls that describe how an operation can be performed. Forms are serializations of Protocol Bindings. The array cannot be empty.mandatoryArray of Form"
 "td-vocab-forms--Thing","null","formsSet of form hypermedia controls that describe how an operation can be performed. Forms are serializations of Protocol Bindings. Thing level forms are used to describe endpoints for a group of interaction affordances.optionalArray of Form"
-"td-vocab-href--Form","null","hrefTarget IRI of a link or submission target of a form.mandatoryanyURI"
-"td-vocab-href--Link","null","hrefTarget IRI of a link or submission target of a form.mandatoryanyURI"
+"td-vocab-href--Form","null","hrefTarget URI[[RFC3986]] of a submission target of a Form. If relative, it is combined with the value in the base in Thing-level to create an absolute URI.mandatoryanyURI"
```
TallTed:
```suggestion
"td-vocab-href--Form","null","hrefTarget URI [[RFC3986]] of a submission target of a Form. If relative, it is combined with the value in the base in Thing-level to create an absolute URI.mandatoryanyURI"
```

validation/td-validation.ttl
```diff
@@ -605,8 +605,7 @@
     sh:order 7 ;
     sh:property [
         sh:path hctl:hasTarget ;
-        skos:definition """Target IRI of a link or submission target of a
-                form."""^^rdf:HTML ;
+        skos:definition """Target URI[[RFC3986]] of a submission target of a <a>Form</a>. If relative, it is combined with the value in the <code>base</code> in Thing-level to create an absolute URI."""^^rdf:HTML ;
```
TallTed:
```suggestion
        skos:definition """Target URI [[RFC3986]] of a submission target of a <a>Form</a>. If relative, it is combined with the value in the <code>base</code> in Thing-level to create an absolute URI."""^^rdf:HTML ;
```

validation/td-validation.ttl
```diff
@@ -747,6 +746,15 @@
     ] ;
     skos:scopeNote """
 
+    <p class="note">
+        Please note that the value of <code>href</code> is of of type URI[[RFC3986]] and what is sent as a protocol message
+        is protocol-specific and can be decoded from <code>href</code>.
+        <span class="rfc2119-assertion" id="td-form-href-unicode-usage">
+            In cases of protocols where Universal Character Set (Unicode) [[ISO10646]] characters are to be sent in the protocol message,
+            the resource MAY be represented internally as an IRI[[!RFC3987]] and can be converted to a URI following Section 3.1 of [[RFC3987]] before placing it in the <code>href</code>.
```
TallTed:
```suggestion
        Please note that the value of <code>href</code> is of of type URI [[RFC3986]] and what is sent as a protocol message
        is protocol-specific and can be decoded from <code>href</code>.
        <span class="rfc2119-assertion" id="td-form-href-unicode-usage">
            When Universal Character Set (Unicode) [[ISO10646]] characters are to be sent in the protocol message,
            the resource MAY be represented internally as an IRI [[!RFC3987]] and can be converted to a URI following Section 3.1 of [[RFC3987]] before placing it in the <code>href</code>.
```

ontology/hctl.html
ektrah: The Target URI is *always* be resolved against the Base URI, not only when it's relative.

validation/td-validation.ttl
ektrah: What does "the resource MAY be represented internally as an IRI" mean?

ontology/hctl.html
egekorkan: That is true. Absolute href on absolute base should return the input absolute href

validation/td-validation.ttl
egekorkan: I was struggling to find good wording here. So the implementation (driver sending the request) is what I mean by internal. The unicode characters can be used there but need to transformed to URI for using in TDs.

ontology/hctl.html
ektrah: No, not exactly. Reference resolution also includes removing dot segments. So, to obtain the Target URI, a TD consumer resolves the `form.href` unconditionally against the Base URI, regardless of whether `form.href` is relative or not.

For the Base URI: The spec doesn't say it, but I assume that `td.base` must not be a relative URI reference. So, to obtain the Base URI, the TD consumer checks that `td.base` is not relative and then resolves it. (It does not matter against which URI `td.base` is resolved because it is not relative.)

validation/td-validation.ttl
ektrah: Maybe explain that a form contains all the necessary information to perform some protocol interaction. This information is encoded in JSON-LD and, in case of addressing information, often in the components of the URI reference in the `href` field of the form. This information needs to be translated into suitable protocol elements, etc.

egekorkan: @ektrah have another look :)

egekorkan:
TD Call:

- @danielpeintner recommends putting a definition for absolute and relative URIs. Probably referring to URI specs
- The wording of `Target URI of a submission target of a Form. It MUST be resolved with the value in the base in Thing-level.`  should be changed to `Target URI of a submission target of a Form. The target URI MUST be resolved using the base URI at the Thing-level.`
- @mahdanoura should update the toolchain repo
- Type issue: Fixes can be 
  - string but adding a regex/ABNF -> not interoperable with implementations in different languages
  - string but adding `(specifically RFC3986)` after it
  - anyURI but adding `(here it is restricted to an "ASCII URI")` after it

The first 2 points will be fixed by @egekorkan . The other two, we can wait for further feedback.

ontology/hctl.html
```diff
@@ -148,27 +148,27 @@ <h2>Terminology</h2>
...
 <section id="Link"><h4>Link</h4><p>IRI: <code>https://www.w3.org/2019/wot/hypermedia#Link</code></p><span>A link can be viewed as a statement of the form "<b><em>link context</em></b>  has a <b><em>relation type</em></b> resource at <b><em>link target</em></b>", where the optional <b><em>target attributes</em></b>  may further describe the resource.</span><table class="def numbered"><tbody><tr><td>In the domain of</td><td><code><a href="#hasAnchor">hctl:hasAnchor</a></code><br><code><a href="#hasHreflang">hctl:hasHreflang</a></code><br><code><a href="#hasRelationType">hctl:hasRelationType</a></code><br><code><a href="#hasSizes">hctl:hasSizes</a></code><br><code><a href="#hasTarget">hctl:hasTarget</a></code><br><code><a href="#hintsAtMediaType">hctl:hintsAtMediaType</a></code></td></tr><tr><td>In the range of</td><td><code><a href="https://www.w3.org/2019/wot/td#hasLink">td:hasLink</a></code></td></tr></tbody></table></section></section><section><h3>Object Properties</h3><section id="additionalReturns"><h4>additionalReturns</h4><p>IRI: <code>https://www.w3.org/2019/wot/hypermedia#additionalReturns</code></p><span>This optional term can be used if additional expected responses
```
mahdanoura, line 150: For all the other definitions of the TD concepts we have natural language text describing them, I find the Link definition provided here hard to understand, it mixes natural language with some other forms of semantics.

ontology/hctl.html
```diff
@@ -148,27 +148,27 @@ <h2>Terminology</h2>
...
...em>operation type</em></b> operation on <b><em>form context</b></em>, make a <b><em>request method</b></em> request to <b><em>submission target</b></em>" where the optional <b><em>form fields</b></em> may further describe the required request. In Thing Descriptions, the <b><em>form context</b></em> is the surrounding Object, such as Properties, Actions, and Events or the Thing itself for meta-interactions.</span><table class="def numbered"><tbody><tr><td>In the domain of</td><td><code><a href="#additionalReturns">hctl:additionalReturns</a></code><br><code><a href="#forContentCoding">hctl:forContentCoding</a></code><br><code><a href="#forContentType">hctl:forContentType</a></code><br><code><a href="#forSubProtocol">hctl:forSubProtocol</a></code><br><code><a href="#hasOperationType">hctl:hasOperationType</a></code><br><code><a href="#hasTarget">hctl:hasTarget</a></code><br><code><a href="#returns">hctl:returns</a></code><br><code><a href="https://www.w3.org/2019/wot/td#hasSecurityConfiguration">td:hasSecurityConfiguration</a></code></td></tr><tr><td>In the range of</td><td><code><a href="https://www.w3.org/2019/wot/td#hasForm">td:hasForm</a></code></td></tr></tbody></table></section>
```
mahdanoura, line 149: Similar to the Link definition, the Form definition is inconsistently defined, mixing NL with some semantics making the text hard to understand

validation/td-validation.ttl
```diff
@@ -747,6 +746,16 @@
...
+        <span class="rfc2119-assertion" id="td-form-href-unicode-usage">
+            As the <a>Form</a> contains all the necessary information to send a protocol message, when Universal Character Set (Unicode) [[ISO10646]]
+            characters are used to represent the resource and are to be sent in the protocol message, they MUST be converted to a URI following Section
+            3.1 of [[RFC3987]] before placing it in the <code>href</code>.
+        </span>
+    </p>
+
     <!-- https://www.w3.org/TR/wot-architecture11/#arch-media-type-extra -->
     <p>
         While this table defines the terms available in a form, <span class="rfc2119-assertion" id="td-form-additional-mediatype-metadata">, Bindings MAY have additional information that specifies representation formats in more detail than the media type alone.</span>
```
mahdanoura:
```suggestion
        While this table defines the terms available in a form, <span class="rfc2119-assertion" id="td-form-additional-mediatype-metadata"> Bindings MAY have additional information that specifies representation formats in more detail than the media type alone.</span>
```

index.html
```diff
@@ -3812,6 +3815,17 @@ <h3><code>Form</code></h3>
...
+              Please note that the value of <code>href</code> is of of type URI [[RFC3986]] and what is sent as a
+              protocol message is protocol-specific and can be decoded from <code>href</code>.
+              <span class="rfc2119-assertion" id="td-form-href-unicode-usage">
+                As the <a>Form</a> contains all the necessary information to send a protocol message, when Universal
+                Character Set (Unicode) [[ISO10646]] characters are used to represent the resource and are to be sent in
+                the protocol message, they MUST be converted to a URI following Section 3.1 of [[RFC3987]] before
+                placing it in the <code>href</code>.
+              </span>
+            </p>
+
```
mahdanoura: Since we currently don’t have a better alternative to `xsd:anyURI` that is also widely supported by existing tooling and implementations, and many TD vocabulary terms already rely on `anyURI`, we will likely need similar clarification for those terms as well.
Instead of repeating this note for each individual term, it might make sense to add a single, general note in the Preliminaries section (where XML datatypes are introduced). This note could state that any vocabulary term typed as xsd:anyURI is intended to be interpreted as a URI per RFC 3986, and that identifiers containing non-ASCII characters must be converted to a URI (i.e., percent-encoded as needed).
The NOTE for href in Forms could then extend this general rule by additionally specifying that the value of `href` must be resolved against the `base`.

index.html
```diff
@@ -3677,7 +3677,10 @@ <h3><code>Form</code></h3>
               <tbody>
                 <tr class="rfc2119-table-assertion" id="td-vocab-href--Form">
                   <td><code>href</code></td>
-                  <td>Target IRI of a link or submission target of a form.</td>
+                  <td>
+                    Target URI [[RFC3986]] of a submission target of a <a>Form</a>. It MUST be resolved with the value
```
mahdanoura: Minor consistency: Maybe we move this MUST statement in the Note below that explains about href.

index.html
egekorkan: In the call today, we have agreed on this direction as it will keep the tables leaner as well and establish a hierarchy (between base and href), thus not repeat the mechanism in multiple places.

index.html
egekorkan: This will be handled by the note mentioned in the conversation above.

index.html
```diff
@@ -3812,12 +3848,20 @@ <h3><code>Form</code></h3>
               </tbody>
             </table>
 
+            <p class="note">
+              The value of the <code>href</code> property in a <a>Form</a> instance can be a <a>relative URI</a> or an
+              <a>absolute URI</a>. To avoid any misinterpretation,
```
danielpeintner: The part "To avoid any misinterpretation" confuses me a bit. I think I know what you mean, but maybe there is a better way to say it.

You refer to https://github.com/w3c/wot-thing-description/pull/2172/#discussion_r2664512921, right? 

What about saying something like this ...

To properly build the target URI, the href value MUST be always resolved using the base URI in Thing-level.

Note: "in" Thing-level reads also wrong to me. Isn't it "base URI from the Thing-level" ?

egekorkan: https://github.com/w3c/wot-thing-description/pull/2172#discussion_r2664512921 is what I mean specifically. Otherwise adapting :)

Issue #2087, opened 2025-03-05
Proposal: Move Forms out of the TD (TD2.0)


hspaay:
24-Mar-2025 Update 1: Changed title and intro based on comments. 

This is an outline of the proposal to solve problems that Forms bring by moving them out of TD2.0.

## Why move Forms out of the TD
Forms are overkill, complex, unnecessary, and very very annoying. 

All protocols bindings need to exchange the same information: operation, thingID, affordance name, correlationID, messageID, message type (request, response, notification), action status, and input or output data.

Thats it. It always boils down to this set of information. Describing the miriad ways of how protocols can do this using forms in every single affordance is unnecessary. There is no need to spread this out over the TD, especially since it depends on the protool used.  Right now forms pretty much let you build a separate protocol binding for every single affordance. This makes the words 'utter overkill' look inadequate. 

Having to specify a form in every single affordance can double the TD size, cost extra processing power and resources to handle, without any benefit. It adds unnecessary complexity, increases the risk of bugs, makes testing more difficult and hinders adoption. It is the wrong solution for a simple problem. Lets get past it in TD2.0 while we can. 

What is described in forms can be moved to the protocol binding. The TD only needs to reference which protocol bindings are supported and what endpoint to use. 

This is IoT, not a multi-media server chat, telephony, broadcasting system that interfaces with every known system known to mankind in the past and future.

## Proposal 

Leverage operations and *let the protocol binding define how to transfer the information for an operation*. 

The TD only needs to specify the protocol binding used and how to establish a secure connection. A Thing level protocols section in the TD is all that is needed. For example:
```
{
  protocols: {
      "websocket": {
         "endpoint": "wss://address:port/path",
         "security": "stuff goes here",
         "encoding": "json",
      }
      "mqtt": {
         "endpoint": "mqtts://address:port/"
         "security": "stuff goes here",
         ...
      }
  }
}
```

The TD defines how to connect. The protocol defines how to exchange messages for operations. 

All protocols exchange the same information: operation, thingID, affordance name, correlationID, messageID, message type (request, response, notification) and input or output data. Let the protocol specify how to exchange this information.

The latest [websocket proposal](https://github.com/w3c/web-thing-protocol/issues/42) (not strawman) is a good example. It works. 

... but what about ... Yep, leave them in the comments. I'm sure there are valid concerns so lets take a look. The benefits are too great to ignore this.

## Bonus oppertunity

[...]

PR #2173, opened 2026-01-08
Common Definitions Simplification

### LINKED ISSUE #2165: Review of Common Definitions Proposal
I finally got to review the [common definitions proposal](https://github.com/w3c/wot-thing-description/tree/main/proposals/common-definitions) (sorry for being late to the party). I've read through the proposal and all of the examples and I have the following feedback.

Firstly I can see a lot of hard work has gone into this and there are a lot of clever ideas for reducing redundancy in Thing Descriptions. I also love that we would finally have inline security definitions!

If the goal is to remove all redundant/repetitive metadata in a Thing Description and make them more concise, then the proposal is a good one.

However, side effects of the current proposal include:
- There are lots of ways of expressing the same thing, with different types of members that can contain each other or inherit from each other
- There is an increased risk of circular references or conflicting definitions
- It is more difficult to validate whether a TD is valid using JSON Schema alone
- It would require significantly more pre-processing for Consumers to arrive at a usable TD
- I imagine trying to write an algorithm to expand and parse a TD using all of these features (whilst detecting circular references and conflicting terms) would be very challenging

For me this raises the question of what exactly we are optimising for, and what the right trade-offs are between metadata redundancy, human readability and Consumer implementation complexity.

Taking all of those factors into account, I wonder whether the current proposal goes a bit too far and whether there could be a middle ground which optimises out some of the redundancy whilst not creating too many layers of interdependency which complicate Consumer implementation complexity.

## Proposal

I've noticed that the existing `schemaDefinitions` member provides common metadata for data schemas but the proposed `securityDefinitions`, `connectionDefinitions` and `formDefinitions` members are all essentially providing common metadata for Forms. That got me wondering whether a good compromise would be to reduce the proposal down to:
- `schemaDefinitions` for data schemas
- `formDefinitions` for Forms, which would also include `base` and `security` members

This would mean collapsing `securityDefinitions` and `connectionDefinitions` into a flatter structure as part of `formDefinitions` and also collapsing the top level `security` and `connection` members into a single `formDefaults` member (as opposed to `form` which I think is a confusing name at the top level of the TD alongside `forms`). I would also seriously consider getting rid of the `inherit` term which is a potential source of a lot of parsing complexity and circular references, and also reconsider how useful the top level `schema` member is.

If I've understood all of this correctly then the main cost of this significant simplification would be that you would not be able to share a common `base` or `security` definition between multiple `formDefinitions` which could result in a tiny bit more redundancy in certain scenarios, but in return it would make parsing Thing Descriptions far more straightforward.

I will include some example Thing Descriptions following this proposal in a follow-up comment.

## Shared Connections

One other piece of feedback that I have is that the [main use case](https://github.com/w3c/wot-thing-description/issues/878) I originally had for this feature was to provide a single WebSocket endpoint which could be connected to once and then shared between all affordances of a Thing. I note that this proposal doesn't actually fully solve that problem since it's still necessary to create a Form for every InteractionAffordance and there are no clear semantics to tell a Consumer to re-use a single connection.

Having spent many months working on a [WebSocket sub-protocol for the Web of Things](https://w3c.github.io/web-thing-protocol/) I am no longer sure this is such an important feature of Thing Descriptions, since it can actually be useful to have Forms for each InteractionAffordance to be explicit about which operations are supported. We also got around the connection re-use issue by simply defining the circumstances under which a socket should be re-used in the sub-protocol specification.

However, since it might be relevant to other protocols I thought I would point out that as far as I can tell this original requirement has not actually been met.

egekorkan:
## Description of Changes

I am taking the proposal in #2165 and adapting to it. The best is to look at the diff in the tds.js file. Here are my takeaways:

- Amount of possible examples (the first test cases) has drastically reduced. That means, there is less ways to express the same concept. This is a big win.
- Simple examples have basically no impact. I only had to replace `form` with `formDefaults`. 
- When all affordances have multiple protocols but the same security or same contentType, the TD gets longer. From a "mechanical" perspective this is negative.
- Extending the above point, longer TDs get easier to understand from a human perspective. They can be better described I would say. That means an implementation who can only support this case and above is easier to implement.
- When all affordances and forms DO NOT use the same form definition (i.e., events with mqtt, writeproperty requiring basic instead of nosec, etc.) the TDs are still weird. They were weird before anyways. An implementation that understands this is going to be more difficult to implement. These are less common but I would not call them edge cases.

As a result, I like this approach more.

## Related Issue

Closes #2165

## Type of Change

### DIFF
--- a/proposals/common-definitions/tooling/schema-new.schema.json
+++ b/proposals/common-definitions/tooling/schema-new.schema.json
@@ -64,12 +64,13 @@
         {
           "type": "array",
           "items": {
-            "type":"string"
+            "type": "string"
           }
-        },{
+        },
+        {
           "$ref": "#/definitions/singleFormDefinition"
         }
-     ]
+      ]
     },
     "formDefinitions": {
       "type": "object",


proposals/common-definitions/tooling/tds.js
```diff
@@ -771,42 +401,49 @@ const recommendedTDs = [
   },
   // 7. readproperty and writeproperty defaults that are not the defaults of the binding (GET and POST)
```
egekorkan: Feedback from @relu91 in the call: it would be helpful for this case to allow `formDefaults` in the affordance level. all properties would have `formDefaults:[read,write]` and all actions would have `formDefaults:[invoke]`

proposals/common-definitions/README.md
egekorkan, line 1: feedback from @relu91 : What about having `href` in a form definition and in none of the forms in the compact TD. That would be relevant for WebThing WS subprotocol or any protocol where the resource identifier is the same as affordance name. This can be relevant for defining the default for http binding where the Thing has pattern of `properties/myPropertyName`. Relevant to address https://github.com/w3c/wot-thing-description/issues/2087

benfrancis:
I think I may be missing something about the way defaults can be overridden.

@egekorkan wrote:
>> When a Form in an InteractionAffordance overrides a value, does it always have to override that value in all of the default set of Forms?
> No.

How would you override a value in only one of the default Forms from an individual Form instance? How does the Form identify which default Form to override?

>> What if an InteractionAffordance has two Forms with different values for the same member, which one wins in the expanded TD?
> No winner :) Both are visible in the expanded TD. This goes in the matrix multiplication discussion.

Same question as above, how does an individual Form specify which default Form from `formDefaults` to override?

> If there 100 http properties and actions and 1 mqtt event, I think it would make sense to use formDefaults and override for mqtt event.

How would you prevent the EventAffordance being populated with the HTTP-based Form upon expansion?


>> If there is one InteractionAffordance which should not have one of the default set of Forms, presumably there is no way to express that? So this only works when every InteractionAffordance needs the full set of Forms?
> Kind of but you can override with form in a Form.

Same question as above. How would you specify in an InteractionAffordance that one of the default Forms should not be included?

> You would still need the href or a way to identify the resource. In case of WebThing WS Subprotocol, it would look rather empty unless href has a meaning.

Right, it's a bit odd in that use case.

Based on my current understanding, in all of the above use cases I would suggest it would be clearer to use `formDefinitions` rather than `formDefaults` so that:
1. You can be explicit in an individual Form instance about which base Form you are extending/overriding
2. Default forms are not automatically populated in every InteractionAffordance even if they're not needed
3. A Form definition can be referenced without any overrides

I think this provides weight to the idea that `formDefaults` should be a single Form, and any use cases where multiple re-useable Forms are needed should use `formDefinitions`.

Am I missing something?

proposals/common-definitions/tooling/tds.js
benfrancis: Or just use `formDefinitions` instead of `formDefaults` in this use case...

proposals/common-definitions/README.md
benfrancis: > What about having href in a form definition and in none of the forms in the compact TD.

Yes, currently the `forms` member of an InteractionAffordance is mandatory, can not be an empty array and `href` is a mandatory member of Form. This means that even if the `href` is the same as the default for every Form you would have to include a `href` with the same value in every Form. If `formDefaults` is specified then maybe `forms` and `href` should not be mandatory.

> That would be relevant for WebThing WS subprotocol or any protocol where the resource identifier is the same as affordance name.

This is usually not the case in the Web Thing Protocol WebSocket sub-protocol, where the same URL is usually used for all affordances of a Thing.

> This can be relevant for defining the default for http binding where the Thing has pattern of properties/myPropertyName

The URL structure is (intentionally) not specified in the HTTP Basic Profile, `href` can not be auto-generated from an affordance name. I don't think the default HTTP binding should prescribe a particular URL structure either.

proposals/common-definitions/README.md
```diff
@@ -107,119 +107,91 @@ Related Issues:
...
-- All of them contain `inherit`, which has type `String`
-- Form can refer to a connection via `Form::connection`
-- Connection can refer to a security `Connection::security`
-- Form can refer to a schema via `Form::additionalResponse` (no change)
-- Security can be overridden in a form using an inline definition for connection. No `security` term in the form level.
-- Thing can refer to some forms in `Thing::forms`
-- Security cannot refer to a connection, form or schema
-- Connection cannot refer to a form or schema
-- Schema cannot refer to anything beside itself, i.e. inheriting or `oneOf` etc.
+### Elementss
```
danielpeintner:
```suggestion
### Elements
```

proposals/common-definitions/tooling/tds.js
egekorkan: I did not see this feedback before. Could you elaborate?

egekorkan:
With the latest push, I think this PR is ready to merge. We have two feedbacks from @relu91 and some from @benfrancis that I will move out of this PR as they are not against this but asking for more simplification or usability improvements.

To be discussed in today's call.

proposals/common-definitions/tooling/tds.js
benfrancis: Rather than complicate `formDefaults` usage further, as a rule of thumb where there are multiple default Forms involved I suggest that a Thing Description author should use `formDefinitions` rather than `formDefaults`. In this case there would be one definition for reading and writing properties and one definition for invoking actions.

egekorkan: OK I see but without `form` in individual forms?

benfrancis: No, _with_ form in individual Forms (though I think there might be a better name for that term).

Issue #2175, opened 2026-01-29
[Common Definitions] Only allow one default Form in `formDefaults`


benfrancis:
PR #2173 will significantly simplify the Common Definitions proposal by flattening all definitions for Forms into a single `formDefinitions` member, and all defaults for Form objects into a single `formDefaults` member. I think this is a big improvement.

However, currently `formDefaults` can contain multiple default Forms (either as an array of strings referencing Form definitions, or possibly as an array of Forms). When `formDefaults` is an array rather than a Form it becomes more like `defaultForms` (i.e. a default set of Forms for every InteractionAffordance) rather than `formDefaults` (i.e. the default values for the Form object). When used like this, what appears to be a single Form in an InteractionAffordance can actually be expanded out into multiple Forms in non-obvious ways when parsed by a Consumer.

In my opinion having multiple defaults is confusing, non-obvious for a human reader of the Thing Description, makes expansion very complicated, and results in lots of weird [edge cases](https://github.com/w3c/wot-thing-description/pull/2173#issuecomment-3791440595) when overriding Forms.

I would suggest that as a further simplification `formDefaults` (or `defaultForm`) should be a single default Form only.

Then as a rule of thumb for Thing Description authors:
- If there is one default Form for all Forms in the Thing Description then define a default Form in `formDefaults` (either as a Form or a string referencing a form definition)
- If there are multiple common Forms then define multiple definitions using `formDefinitions` and reference them from Forms

Although not always as compact in terms of total Thing Description size, this makes it much more explicit that an InteractionAffordance has multiple Forms, makes it more obvious which Forms are being overridden at the InteractionAffordance level, simplifies expansion, and removes a bunch of weird edge cases.

See https://github.com/w3c/wot-thing-description/issues/2165#issuecomment-3553915346 for example Thing Descriptions.

egekorkan:
Sorry I did not see the second comment.

I will make a "forever draft" PR to show the changes. I am not 100% sure but the changes seem to on examples 8-10. I don't see any benefit of doing what you are proposing though. Makes it more explicit but not able to express that there are always multiple ways to use each affordance. Something that came to my mind is to propose something like:

- defaultForm : object. Use when there is a single form everywhere. overwrite terms as needed in each form
- defaultForm**s** : array of objects. Use when there are multiple default forms everywhere -> expressing that there are always multiple ways to use each affordance
[...]

PR #2179, opened 2026-02-04
More explicit form usage

egekorkan:
Basically the proposal in https://github.com/w3c/wot-thing-description/issues/2175 as a PR for change tracking


egekorkan:
TD Call today:
- We will keep it the way it is and let implementation experience decide if this expansion is not wanted.

PR #2198, opened 2026-04-01
Add assertions on link relation types (#2151 13, 14)

### LINKED ISSUE #2151: Adding relevant Arch Assertions
The relevant assertions are below. They were categorized as A3 or A2. They will be assigned to people tomorrow

---

@k-toumura  the next 4
- [ ] 1. In W3C WoT, the description metadata for a Thing instance MUST be available as a WoT Thing Description (TD) [WOT-THING-DESCRIPTION]. ([arch-td-metadata](https://www.w3.org/TR/wot-architecture11/#arch-td-metadata))
  - Add to [Behavioral Assertions](https://w3c.github.io/wot-thing-description/#behavior)?)
 
- [ ] 3. To be considered a Thing, however, at least one TD representation MUST be available. ([arch-td-mandatory](https://www.w3.org/TR/wot-architecture11/#arch-td-mandatory))  - Can be merged with 1
  - As informative note)

- [ ] 13. Extension relation types MUST be compared as strings using ASCII case-insensitive comparison, (c.f. ASCII case insensitive). (If they are serialized in a different format they are to be converted to URIs). ([arch-rel-types](https://www.w3.org/TR/wot-architecture11/#arch-rel-types))
  - We should change the TD spec as the https://www.rfc-editor.org/rfc/rfc8288.html#section-2.1.1 requires case insensitive comparison. The receiver should be permissive

- [ ] 14. Nevertheless, all-lowercase URIs SHOULD be used for extension relation types [RFC8288]. ([arch-rel-type-lowercase](https://www.w3.org/TR/wot-architecture11/#arch-rel-type-lowercase))
  - Conflict at `controlledBy`

----
@egekorkan next 4
- [x] 16. Form context and submission target MAY point to the same resource or different resources, where the submission target resource implements the operation for the context. ([arch-form-iris2](https://www.w3.org/TR/wot-architecture11/#arch-form-iris2))

  - Explanation: basically GET and PUT can point to the same resource but we differentiate with methods. This is not explicit in TD spec

- [x] 17. The request method MUST identify one method of the standard set of the protocol identified by the submission target URI scheme. ([arch-op-request-method](https://www.w3.org/TR/wot-architecture11/#arch-op-request-method))

  - After https://w3c.github.io/wot-thing-description/#td-bindings-server-accept
  - Rewording that linked assertion can be also enough. Basically, this is saying that the TD cannot contain an HTTP method like `GIT`.

- [x] 20. Interaction Affordances MUST include one or more Protocol Bindings. ([arch-hypermedia](https://www.w3.org/TR/wot-architecture11/#arch-hypermedia))

- [x] 24. Protocol Bindings MAY have additional information that specifies representation formats in more detail than the media type alone. ([arch-media-type-extra](https://www.w3.org/TR/wot-architecture11/#arch-media-type-extra))

----
@k-toumura next 4

- [ ] 11. Events MAY be triggered through conditions that are not exposed as Properties. ([arch-event-trigger](https://www.w3.org/TR/wot-architecture11/#arch-event-trigger))

- [ ] 25. Thus, the Interaction Affordance for structured data types SHOULD be associated with a data schema to provide more detailed syntactic metadata for the data exchanged. ([arch-schema](https://www.w3.org/TR/wot-architecture11/#arch-schema))

- [ ] 28. Authentication and authorization SHOULD be established based on separately managed Private Security Data. ([arch-security-consideration-auth-private-data](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-auth-private-data))

  - Fits to https://w3c.github.io/wot-thing-description/#behavior-security

- [ ] 29. Producers of TDs MUST ensure that no Private Security Data is included in TDs. ([arch-security-consideration-no-private-security-data](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-no-private-security-data))
  - Fits to https://w3c.github.io/wot-thing-description/#behavior-security
  - Should be merged with 27 and 28. 29 covers the other two. There is no way to violate one but fullfil the other two. Then 27 and 28 will be A2.

---
@k-toumura  for the rest

- [x] 46. When secure transport over TCP is appropriate, then at least TLS 1.3 [RFC8446] SHOULD be used. ([arch-security-consideration-tls-1-3](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-tls-1-3))

  - This should be a generic assertion to all bindings that can support TCP. The implementation enforcement can only happen in the binding, which is informative. However, we should recommend all implementers to do that. In the binding documents, there should be informative notes about this that can point back to TD.
  - **Note: the same concern seem to apply to OPC UA, which can use other mechanisms that TLS for security. We need to change the 3 following assertions to not constrain to a specific transport layer security. QUIC is over UDP for example.**

- [x] 47. If TLS 1.3 cannot be used for compatibility reasons but secure transport over TCP is appropriate, TLS 1.2 [RFC5246] MAY be used. ([arch-security-consideration-tls-1-2](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-tls-1-2))

- [x] 48. If DTLS 1.3 cannot be used for compatibility reasons but secure transport over UDP is appropriate, then DTLS 1.2 [RFC6347] MAY be used. ([arch-security-consideration-dtls-1-2](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-dtls-1-2))

- [x] 49. Versions of DTLS or TLS earlier than 1.2 MUST NOT be used for new development. ([arch-security-consideration-no-earlier-tls-or-dtls](https://www.w3.org/TR/wot-architecture11/#arch-security-consideration-no-earlier-tls-or-dtls))

- [x] 50. Storage of explicit PII in TDs SHOULD be minimized as much as possible. ([arch-privacy-consideration-min-explicit-pii](https://www.w3.org/TR/wot-architecture11/#arch-privacy-consideration-min-explicit-pii))

- [x] 53. Things returning data or metadata (such as TDs) associated with a person SHOULD use some form of access control. ([arch-privacy-consideration-access-control-mandatory-person](https://www.w3.org/TR/wot-architecture11/#arch-privacy-consideration-access-control-mandatory-person))

k-toumura:
## Description of Changes

This PR introduces new assertions regarding the comparison and casing of Link relation types to align
with Architecture document, while addressing conflicts with the TD Information Model's strict
case-sensitivity and exisiting relation types (such as `controlledBy`).

## Related Issue

Closes #2151

## Type of Change

***
<a href="https://pr-preview.s3.amazonaws.com/k-toumura/wot-thing-description/pull/2198.html" title="Last updated on Apr 14, 2026, 12:11 AM UTC (a5e01ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/2198/6589c95...k-toumura:a5e01ae.html" title="Last updated on Apr 14, 2026, 12:11 AM UTC (a5e01ae)">Diff</a>


testing/assertions.csv
```diff
@@ -215,6 +215,8 @@
 "td-property-names_unit","null","The vocabulary term unit MUST be serialized as a JSON name within a Property object."
 "td-property-names_uriVariables","null","The vocabulary term uriVariables MUST be serialized as a JSON name within a Property object."
 "td-property-names_writeOnly","null","The vocabulary term writeOnly MUST be serialized as a JSON name within a Property object."
+"td-rel-type-case-insensitive","null","To ensure interoperability, relation types MUST be compared as string using ASCII case-insensitive comparison."
```
danielpeintner: I wonder whether we need to mention ASCII at all?


```suggestion
"td-rel-type-case-insensitive","null","To ensure interoperability, relation types MUST be compared as string using case-insensitive comparison."
```

EDIT: I see you refer to https://infra.spec.whatwg.org/#ascii-case-insensitive
Maybe it is fine...

validation/td-validation.ttl
```diff
@@ -1168,6 +1168,26 @@
...
+    for link relations.  TD Consumer should be permissive and handle mixed-case relation types gracefully.
+    </p>
+    <p>
+    <!-- arch-rel-type-lowercase -->
+    <span class="rfc2119-assertion" id="td-rel-type-lowercase">
+    Furthermore, all-lowercase URIs SHOULD be used for extension relation types.
+    </span>
+    Note that while this is the recommended practice for newly defined extensions,
+    some existing WoT-specific relation types in the table below (such as <code>controlledBy</code>) have historically used mixed casing.
+    Due to the case-insensitive compaison requirement, these legacy mixed-case types remain fully valid and interoperable.
```
egekorkan:
```suggestion
    Due to the case-insensitive comparison requirement, these legacy mixed-case types remain fully valid and interoperable.
```

validation/td-validation.ttl
```diff
@@ -1168,6 +1168,26 @@
...
     is an instance of a specific Thing Model), or to further documentations information (e.g., device manual of a Thing). It is recommended to reuse existing and established
     <a href="https://www.iana.org/assignments/link-relations">Link Relation definitions from IANA.</a>
+    </p>
+    <p>
+    As specified in Web Linking [[RFC8288]], link relation type can be ether registered relation types or extension relation types.
+    <!-- arch-rel-types -->
+    <span class="rfc2119-assertion" id="td-rel-type-case-insensitive">
+    To ensure interoperability, relation types MUST be compared as string using ASCII case-insensitive comparison.
+    </span>
+    (c.f. <a href="https://infra.spec.whatwg.org/#ascii-case-insensitive">ASCII case insensitive</a>.
```
egekorkan: I think that https://datatracker.ietf.org/doc/html/rfc8288#section-2.1.1 is a better reference here.

> Registered relation type names MUST conform to the reg-rel-type rule (see [Section 3.3](https://datatracker.ietf.org/doc/html/rfc8288#section-3.3)) and MUST be compared character by character in a case-insensitive fashion.

validation/td-validation.ttl
```diff
@@ -1168,6 +1168,26 @@
...
+    If they are serialized in a different format they are to be converted to URIs).
+    Because the general TD Information Model is strictly case-sensitive, this rule act as an explicit exception
+    for link relations.  TD Consumer should be permissive and handle mixed-case relation types gracefully.
+    </p>
+    <p>
+    <!-- arch-rel-type-lowercase -->
+    <span class="rfc2119-assertion" id="td-rel-type-lowercase">
+    Furthermore, all-lowercase URIs SHOULD be used for extension relation types.
```
egekorkan: First I was a bit confused but then saw it is specified like this in RFC8288. I think the initial link to the RFC should be more specific and say that we strictly follow that. So these assertions are not really our creation. Then we can give these as examples. So something like: "The link relation types follow RFC8288, meaning -> a bullet list of examples like URIs for extension, case insensitive comparion etc.

egekorkan (changes requested):

k-toumura:
Thank you for the suggestion!  I completely agree that we should make it clear these assertions come directly from RFC8288 and are not our own new creations.

I've updated the PR to explicitly state that the link relation types follow Web Linking,
and restructured the assertions into a bullet list.

I also made two minor adjustments:
- I put the reference to Section *2.1 Relation Type* as a whole, rather than linking to 2.1.1 Registered Relation Types or 2.1.2 Extension Relation Types.  This is because the assertions related to both sections.
- I removed the word "ASCII" from "case-insensitive comparison" because the specific term is not explicitly used in RFC8288.

(side note: Is it possible to reference a specific section of RFC directly in ReSpec?
[ReSpec document](https://respec.org/docs/#examples-4) mentions that the format `[[[SHORTNAME#id]]]` (e.g. `[[[RFC8288#link-relation-types]]]`) can be used, but trying this resulted in an error in my local environment... )

egekorkan (approved):
Thank you @k-toumura ! The changes are on point.

> (side note: Is it possible to reference a specific section of RFC directly in ReSpec?
[ReSpec document](https://respec.org/docs/#examples-4) mentions that the format [[[SHORTNAME#id]]] (e.g. [[[RFC8288#link-relation-types]]]) can be used, but trying this resulted in an error in my local environment... )

Same happened to me before and there is warning on respec that this is not recommended

TallTed:
> Is it possible to reference a specific section of RFC directly in ReSpec?

Try using [`data-cite`](https://respec.org/docs/#data-cite). Note that this belongs inside an `<a ...>`, rather than a `[[[ ... ]]]`

egekorkan (approved):
TD Call today: Decided to merge.

To be 100% formally correct, this means we introduce a new string type that is case insensitive but modeling this would be too cumbersome so we ignore for now.

PR #2199, opened 2026-04-14
TM Placeholder Fix

### LINKED ISSUE #2194: Unclear algorithm for placeholder replacement
Hi!

Thank you for the excellent work on WoT TD 1.1.

I'm implementing section 9.3.5 of the spec: https://www.w3.org/TR/wot-thing-description11/#thing-model-td-placeholder

It defines a placeholder regex: `{{2}[ -~]+}{2}`

When I apply this regex on something like:

```
{{FIRST}} {{LAST}} -- {{POSITION}} at Acme Corp
```

I get a single match, spanning `{{FIRST}} {{LAST}} -- {{POSITION}}`, which means that the placeholder identifier is `FIRST}} {{LAST}} -- {{POSITION`.

Is this intended? If so, the spec (in my opinion) should say that only one placeholder per string is allowed.

Otherwise, if this is not intended, I suggest clarifying that multiple placeholders per string are allowed, and the regex should be modified to `{{2}[ -~]+?}{2}` (add a `?` after `+` to get a lazy matcher).

egekorkan:
## Description of Changes

This is addressing all the problems in #2194. I am not changing the regex used in the JSON Schema as there is no issue with validation. Specifically, it clearly mentions that multiple placeholders are allowed, if an object is used in a string, it should be stringified with no spaces etc. and the regex that was proposed by @Ostrzyciel is integrated but as a note, not as an assertion. 

## Related Issue

Closes #2194

## Type of Change

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/2199.html" title="Last updated on Apr 15, 2026, 9:54 AM UTC (75e28c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/2199/2b076cf...75e28c6.html" title="Last updated on Apr 15, 2026, 9:54 AM UTC (75e28c6)">Diff</a>


danielpeintner (changes requested):
The code snippet showing multiple placeholders causes issues with ReSpec rendering even though ReSpec should understand that it is part of `<code>`

`<code>"mqtt://{{MQTT_BROKER_ADDRESS}}:{{PORT}}"</code>`

looks like this which is not what we want.

Image

You might need to change the line to something like

`<code>"mqtt://</code><code>{{</code><code>MQTT_BROKER_ADDRESS</code><code>}}</code><code>:</code><code>{{</code><code>PORT</code><code>}}</code><code>"</code>`

egekorkan:
@danielpeintner yup you are right. Fixed :)

egekorkan:
TD Call Today: We are going to merge this as the group agrees.

PR #2205, opened 2026-05-27
Fixes to JSON Schema to allow a single URI for the context value

### LINKED ISSUE #2204: JSON Schema rejects single string context
The `thing-context` definition in the [JSON Schema](https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json) only accepts `@context` as an array:

 ```
 "thing-context": {
    "type": "array",
    "items": [{ "$ref": "#/definitions/thing-context-td-uri-temp" }],
    "additionalItems": { ... }
  }
```

  Per the spec, `@context  can also be a single URI string.

  This is rejected by the schema because thing-context enforces `"type": "array"` without a `oneOf` alternative for the `string` form.

egekorkan:
## Description of Changes

Reverting the problem caused in #2128 

## Related Issue

Closes #2204 

## Type of Change

### DIFF
--- a/validation/td-json-schema-validation.json
+++ b/validation/td-json-schema-validation.json
@@ -76,27 +76,27 @@
           "$ref": "#/definitions/thing-context-td-uri-temp"
         },
         {
-        "$comment": "Experimental context URI with other vocabularies after it",
-        "type": "array",
-        "items": [
-          {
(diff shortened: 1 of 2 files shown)


egekorkan:
Done the testing as well and should be good to go.

Issue #1541, opened 2022-06-15
Bringing back writeable and readable


egekorkan:
Automatically deferred to 2.0

We know what the discussion will be.

See also: https://github.com/w3c/wot-thing-description/issues/854#issuecomment-1156634978

egekorkan:
There is a new use case to bring this back. There are protocols (such as lorawan and mqtt) where the protocol itself is completely async, you cannot ask for the new values via readproperty. However, gateway implementations want to provide the readproperty behavior, so they "retain" the last value. In MQTT, that can be based on the publish settings too. So the Consumer of the gateway uses readproperty. For this workflow, it would be possible to turn an event to an observable and readable property but the mindset is to add a `readproperty` operation as events seem to reserved to more "alarm" like semantics, like signaling that something has happened, not for streaming data. We can discuss whether events should be restricted to "alarms" or not but if the implementers want to do it in a certain as a majority, that should be respected. Thus, it would be very beneficial to have observeonly properties, meaning we have readable:false, writeable:false and observable: true

egekorkan:
However, we need to discuss if we move these terms to the property affordance level or should they stay in data schema? We have https://w3c.github.io/wot-thing-description/#sensor-example-webhook where we can see data schema usage of these keywords. However, I am not sure if implementations really do this behavior.

danielpeintner:
Clarification question: does "bring back writeable and readable" mean adding those terms or replacing readOnly and/or writeOnly to avoid conflicts?

egekorkan:
Regarding data schema we have:

- not allow readability annotation for data schema -> no readability in actions and events' schemas.
  - We also don't have observability annotation in data schema
- allow readable writeable in data schema -> same behaviour as now but other keywords. 
- allow readable writeable in property and readOnly and writeOnly in data schema

An alternative is to remove `writeOnly` and `readOnly` completely and not introduce an alternative like readable. The Consumer applications need to compute based on the forms' op values.

danielpeintner:
> An alternative is to remove `writeOnly` and `readOnly` completely and not introduce an alternative like readable. The Consumer applications need to compute based on the forms' op values.

I start liking this proposal suggested by @relu91 👍

egekorkan:
Regarding remove them completely: I understand that it removes redundant information but it is pushes humans to look into the forms, which is something they should never need to do. Those keywords influence the behavior of the application significantly and need to be visible.

siddharthbaleja7:
[...]

PR #2206, opened 2026-06-18
Interaction Semantics Section

egekorkan:
## Description of Changes

I am addressing `Explaining how and where they are used today (without the redesign) would help in the first place.` comment in #1541. There is a bit of overlap between a couple of sections on this.

After putting the overall content, I am not sure about some aspects:

- There are linting level considerations 
- Even though it was meant to be informative, it is becoming more normative
- While the work is needed, I am not sure if we should do this kind of work until we have more features out there. I feel that this section would result in a rather philosophical discussion. Still we need to merge this before the toolchain change

## Related Issue

not closing but related to #1541 

## Type of Change

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/pull/2206.html" title="Last updated on Aug 19, 2026, 4:17 PM UTC (0882ddd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-thing-description/2206/18b7ca0...0882ddd.html" title="Last updated on Aug 19, 2026, 4:17 PM UTC (0882ddd)">Diff</a>


egekorkan:
We had a discussion on this today. This makes it clear that we have a weird situation with the possible usage with `readOnly` and `writeOnly`. The idea for this PR is to move the tables and json examples to a markdown since it is not very clear for readers what to do with them. Even for me it is not clear what to do with them :D

After this analysis, with the current terminology, `readOnly` and `writeOnly` only makes sense for the root of the property. In other places, it is either implied, makes no sense or another wording would have made more sense (like readOnly in event subscription). With @relu91 , one intuitive direction was orienting towards adding input and output everywhere.

egekorkan:
I have moved the confusing table to analysis folder and we should pick up on it later on. The changes in this PR are still relevant as I move some text around.

egekorkan:
There are some editorial work to be done in the new section. Then we can have another look tomorrow and merge it.

A separate discussion to be had. Another issue will be created.

- The default for properties can be simple like today but should not have readOnly, writeOnly anymore
- Events with input and output in subscription and cancellation
- SDF has input and output for events and actions. Not for properties

egekorkan:
https://github.com/w3c/wot-thing-description/issues/2215 created

PR #2226, opened 2026-08-26
Tests fix to JSON Schema tests only

### LINKED ISSUE #2210: Fixing and Splitting CI tests for validation
We have reenabled the CI tests but they are not working right now. Also, it would be better to split the tests for JSON Schema. Only this one should be in another test https://github.com/w3c/wot-thing-description/blob/c9ec652b7d256e0a5c25128022bf556447afe88f/validation/td-validation.test.js#L42

egekorkan:
## Description of Changes

Had to stop testing for SHACL and JSON-LD expansion. Only JSON Schema tests now. Mocha is bumped to 11, node.js 26 is used now.

## Related Issue

Closes #2210

## Type of Change

- Prepatory work such analyses, user stories etc.


package.json
```diff
@@ -7,7 +7,7 @@
     "render": "bash ./render.sh && prettier . --write",
     "assertions": "node assertions.js",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "test:validation": "mocha validation/**/*.test.js",
+    "test:schema": "mocha validation/td-validation-schema.test.js",
```
danielpeintner, line 10: What seems to be missing is the `test:shacl` task

egekorkan: Added it but running it means failing all the tests for now as JSON-LD to RDF does not work. Not sure what is causing the failure but json-ld playground is working fine.

egekorkan: Ok bumping jsonld package to 9.0.0 improved it

PR #2230, opened 2026-09-02
Removing Resources after wot-resources migration

egekorkan:
## Description of Changes

To avoid mistakenly changed resources here, we should basically delete them and point to where people should look at.

## Related Issue

Part of https://github.com/w3c/wot-resources/issues/3#issuecomment-1759133621

## Type of Change

- Prepatory work such analyses, user stories etc.


egekorkan:
Approved in today's call but leaving it around for a bit since it is very fresh now.

publication/ver11/7-resources/README.md
```diff
@@ -0,0 +1,4 @@
+# Resources with TD 1.1
+
+The resources have been moved to <https://github.com/w3c/wot-resources/tree/main/td/v1.1> and are maintained over there.
+If you need to generate resources for TD 1.0, make sure to switch to the relevant tag at <https://github.com/w3c/wot-thing-description/tree/REC1.1>.
```
danielpeintner:
```suggestion
If you need to generate resources for TD 1.1, make sure to switch to the relevant tag at <https://github.com/w3c/wot-thing-description/tree/REC1.1>.
```

Copy and paste error... I guess

PR #2233, opened 2026-09-10
Add data mapping user story 4 analysis

wiresio:
Analysis data data mapping user story 4 - Enum mapping

Issue #1936, opened 2023-12-13
Supporting complex/structured types in simple protocols


egekorkan:
Relates to #1930.

Some devices do not support complex/structured types but device manufacturers do some "adjustments" to accommodate that. Something we see in Modbus devices is that object-like data is split into multiple registers where one set contains one "key" of the object and the other contains another "key". There is an example at the end where 3 registers contain the voltage value and another 3 contain a timestamp in Unix time.

This is part of the data mapping discussion, but we need an object data schema that maps one key to one set of registers and the other key to the others.

![modbus-complex](https://github.com/w3c/wot-thing-description/assets/20195407/b6496b51-4a09-4be1-9b79-22b36447058b) 

Also available publicly at https://cache.industry.siemens.com/dl/files/595/34261595/att_951630/v1/manual_pac4200_en-US_en-US.pdf on page 161 (Table A- 17)

lu-zero:
This is a good example of a data map in the read direction, if we consider adding an expression-filter on top of mapping we could provide conversions from bytes to float and integer.

Kaz040:
This issue is being discussed in relations to PROFINET also. a resource in a profinet device can be an object. the definition of the object payload is always provided in the device manual. an example is this SITOP device,  [https://cache.industry.siemens.com/dl/files/415/84977415/att_906547/v1/A5E37775406-7-76_MANUAL_SITOP-UPS1600-UPS1100_en-US.pdf](url) on page 130. 

The first parameter --> Buffering parameters 

![image](https://github.com/w3c/wot-thing-description/assets/67549739/327d190e-54ed-49c2-a540-0a1d753b43c6)


is a complex payload. inside it are parameters defined in the table below

![image](https://github.com/w3c/wot-thing-description/assets/67549739/a83b2ade-c8ec-46dc-afd1-7c9f692cc75a).

so to be able to capture what each byte or byte stream from the payload mean, terms like **profinet : offset**, **profinet : byte or profinet : length** (decision not made yet on term name)  and **profinet : bit, can be profinet : bitmask**  are to be introduced. To do this, the definition has to be at dataSchema level. 

My question now is, should these terms be protocol specific or generic for payloads in byte streams?

relu91:
> My question now is, should these terms be protocol specific or generic for payloads in byte streams?

[...]

Issue #1930, opened 2023-11-30
Supporting bitmaps 


egekorkan:
There is a need to represent bitmaps, i.e. a list/array of bits, where each bit corresponds to a boolean value like alarm 1, alarm 2 or status like `in-alarm` (bit 0), `fault` (bit 1), `overridden` (bit 2), `out-of-service` (bit 3). This is very common in Modbus and BACnet and possibly other protocols. 

The big question is how this looks in the data schema. This relates a lot to the recent discussions we had about uri variables in BACnet and HTTP (also slides from @lu-zero in TPAC23). My initial thought is that it should be an array or an object and then we map this to specific bits via forms.

Note: This is split from https://github.com/w3c/wot-binding-templates/issues/101 and https://github.com/w3c/wot/issues/119#issuecomment-219436032 . By splitting that binding template issue into multiple smaller ones, I will close that.

mahdanoura:
**Source:** PAC4200 System Manual, Appendix A.2.1  

<img width="1489" height="1985" alt="Image" src="https://github.com/user-attachments/assets/cb9ab5eb-c501-46c6-83fc-f4bfe1e9859e" />

Offset 203, 2 registers, `Unsigned long` — "Limit Violations". The value encodes 12 limit flags and 5 logic-result flags across a 32-bit word in **non-contiguous** byte-and-bit positions:

```
| Byte | Bits | Semantics |
|------|------|-----------|
| Byte 3 | Bits 0–7 | Limits 0–7 (one boolean per bit) |
| Byte 2 | Bits 0–3 | Limits 8–11 (one boolean per bit) |
| Byte 2 | Bits 4–7 | *(unused)* |
| Byte 1 | — | *(unused)* |
| Byte 0 | Bit 0 | Limit logic flag |
| Byte 0 | Bits 1–4 | Logic results 1–4 |
```

In this case, each bit position maps to a distinct named boolean property, but they are spread across non-contiguous bytes with byte-level gaps

derwehr:
> In this case, each bit position maps to a distinct named boolean property, but they are spread across non-contiguous bytes with byte-level gaps

Byte sequences like this are the reason i added this issue (and a PR for it) to node-wot: https://github.com/eclipse-thingweb/node-wot/issues/1099

Many [Bluetooth devices](https://www.bluetooth.com/specifications/specs/) transmit data that is not byte-aligned.
[...]

PR #2234, opened 2026-09-17
Data mapping user story 5

wiresio:
Include PR comments from https://github.com/w3c/wot-thing-description/pull/2233#pullrequestreview-5223398785

PROFINET examples still to be refined


planning/work-items/analysis/data-mapping/US5-details.md
```diff
@@ -42,7 +41,6 @@
 - `bitCompose` - convert a structured object of fields into one integer.
```
Kaz040, line 41: I am not sure if the bitExtract and bitCompose will cover all application/octet-streams payload, especially byte(s) extraction and byte(s) composition
I tried to produce example on two parameters from the SENTRON PAC4200 : https://github.com/w3c/wot-thing-description/issues/1936#issue-2039515535 . and SITOP device: https://github.com/w3c/wot-thing-description/issues/1936#issuecomment-1880817227, I could not.

planning/work-items/analysis/data-mapping/US5-details.md
Kaz040: I am assuming these scopes are related to bitExtract and bitCompose. If yes, then maybe we should change the wording a little bit because the value that is transmitted as application/octet-stream on the wire will be array of bytes(bytearray) on the application side. I am suggesting, 

- Decomposing packed byte fields into structured boolean or integer values.
- Composing structured fields back into packed byte values.

This itself also does not cover payload that might string, timestamps, floats. I would also suggest we extend map:type to have more that boolean and integer type.

Kaz040: is the fieldValue = application value and wireValue = protocol payload? 

For the composition, the wireValue is included in the logical expression, where are we going to get it from to execute the composition logical expression?

planning/work-items/analysis/data-mapping/US5-details.md
```diff
@@ -256,46 +246,45 @@ The following JSON-LD context defines the structural conversion terms not covere
...
-| ------------ | -------------------------- | ---------------------------------------------------------------------- |
-| `map:fields` | `bitExtract`, `bitCompose` | Ordered list of field definitions.                                     |
-| `map:name`   | Field definition           | Name used as the structured object's key.                              |
-| `map:mask`   | Field definition           | Non-zero integer mask selecting the field's bits.                      |
-| `map:shift`  | Field definition           | Non-negative right shift for extraction or left shift for composition. |
-| `map:type`   | Field definition           | `boolean` or `integer`; default is `integer`.                          |
+| Term | Used by | Description |
+|---|---|---|
+| `map:fields` | `bitExtract`, `bitCompose` | Ordered list of field definitions. |
+| `map:name` | Field definition | Name used as the structured object's key. |
```
Kaz040: can we include a clause to restrict the map:name to match the property affordance name it is mapping? in the enum and enumRange operation, we can then remove mapFrom, mapTo and introduce only the map:name that will be consistence along all map:proc.

planning/work-items/analysis/data-mapping/US5-details.md
```diff
@@ -256,46 +246,45 @@ The following JSON-LD context defines the structural conversion terms not covere
...
+| `map:mask` | Field definition | Non-zero integer mask selecting the field's bits. |
+| `map:shift` | Field definition | Non-negative right shift for extraction or left shift for composition. |
```
Kaz040, line 287: While these might work for boolean and integer operations as intended from the scope definition. we need more that boolean and integer types because some payload can be string, timestamp, float e.t.c

planning/work-items/analysis/data-mapping/US5-details.md
```diff
@@ -256,46 +246,45 @@ The following JSON-LD context defines the structural conversion terms not covere
...
+| `map:type` | Field definition | `boolean` or `integer`; default is `integer`. |
```
Kaz040: my suggestion on this term is to use xsd typ. Then, in each protocol bindings, their protocol specific type are mapped to the xsd type. luckily,  there are many similarities between the protocol types of these low level protocols and xsd.

planning/work-items/analysis/data-mapping/US5-details.md
```diff
@@ -537,41 +559,101 @@ This example decomposes a 16-bit status register into an application-level objec
...
+        "limit2": { "type": "boolean" },
+        "limit3": { "type": "boolean" },
+        "limit4": { "type": "boolean" },
+        "limit5": { "type": "boolean" },
+        "limit6": { "type": "boolean" },
+        "limit7": { "type": "boolean" },
+        "limit8": { "type": "boolean" },
+        "limit9": { "type": "boolean" },
+        "limit10": { "type": "boolean" },
+        "limit11": { "type": "boolean" },
```
Kaz040, line 585: I do not have access to the device manual though but I think the limits are mutually exclusive values. Maybe an investigation should be done on that register to see the payload output.

Kaz040 (commented):
Thank you so much for this extensive work on data mapping @wiresio. This opens the path to general mapping solution instead of protocol dependent solutions. 

For protocols like PROFINET, IO-LINK and CAN. their data-mapping principle follow similar pattern.This pattern is mostly addressed in the Bitfield operator. 

In general, I think we can extend the Field Definition part to cover the types on the wire and data metadata for protocols with octet-streams payload.

planning/work-items/analysis/data-mapping/US5-details.md
wiresio: The register information is here: https://github.com/w3c/wot-thing-description/issues/1930#issuecomment-4342467719

planning/work-items/analysis/data-mapping/US5-details.md
wiresio: Would be possible to use, e.g., "xsd:boolean" instead. However, for the sake of consistency to not express the same concept in two ways, I would stick to the JSON schema type string.

planning/work-items/analysis/data-mapping/US5-details.md
wiresio: Agreed.
- restrict map:name to always equal the target property/affordance name, and remove map:mapFrom/map:mapTo
- enum, when applied to one member of a structured object, will use the same map:name term as bitExtract/bitCompose, acting in place (read and write the same key)

WIll update Example 4 accordingly.

planning/work-items/analysis/data-mapping/US5-details.md
wiresio: Will correct and clarify this.

planning/work-items/analysis/data-mapping/US5-details.md
wiresio: Will add an explanation: 
- bitExtract/bitCompose can only produce/consume an unsigned integer magnitude (boolean being the 1-bit case)
- other application-facing types are reached by composing a follow-up numeric (US3) or enum (US4) operator on the extracted integer

We can discuss whether a future operator is needed for reinterpreting extracted bits as an IEEE-754 float rather than an unsigned magnitude. Also we can can think about formatting an extracted epoch integer as an application-facing timestamp string.

planning/work-items/analysis/data-mapping/US5-details.md
wiresio: Will clarify the wording and apply the changes as written in: https://github.com/w3c/wot-thing-description/pull/2234#discussion_r4081452801

planning/work-items/analysis/data-mapping/US5-details.md
wiresio: You're right — bitExtract/bitCompose can't express either example, but that's because neither is a bitfield problem. Will add a clarification and another "open issues" bullet ;-)

wiresio:
> Thank you so much for this extensive work on data mapping @wiresio. This opens the path to general mapping solution instead of protocol dependent solutions.
> 
> For protocols like PROFINET, IO-LINK and CAN. their data-mapping principle follow similar pattern.This pattern is mostly addressed in the Bitfield operator.
> 
> In general, I think we can extend the Field Definition part to cover the types on the wire and data metadata for protocols with octet-streams payload.

Thanks for you review comments @Kaz040!

Please see my answers inline. Will update the document and put the open issues here.

planning/work-items/analysis/data-mapping/US5-details.md
Kaz040: Ok. Now I could see the document. the limits are not mutually exclusive.

planning/work-items/analysis/data-mapping/US5-details.md
Kaz040: Alright, then bindings will need to map protocol types to JSON schema type right? We can also state in the data mapping document that format from JSON schema could be used at affordance level for strings that need formatting.

planning/work-items/analysis/data-mapping/US5-details.md
```diff
@@ -256,46 +246,45 @@ The following JSON-LD context defines the structural conversion terms not covere
 ```
 
 **Notes:**
-
 - `valueMapping`, `fromWire`, and `toWire` are shared pipeline attachment and direction terms.
 - `op` and the structural operation identifiers are proprietary execution step identifiers.
-- `map:path` is deliberately narrower than a general query language: it addresses object keys and array indexes only.
+- `map:path` uses [JSON Pointer](https://www.rfc-editor.org/rfc/rfc6901) syntax. It addresses object members and array elements only; it does not support filtering, wildcards, unions, or expressions.
```
danielpeintner: Maybe we can make it clearer that **we** limit JSON Pointer.
Currently it reads a bit as if JSON Pointer does not support filtering, etc.

wiresio: Thanks, will make it clearer.

planning/work-items/analysis/data-mapping/US5-details.md
wiresio: Bindings may use XSD datatypes or their own protocol-specific type vocabulary. However, for user stories 3, 4, and 5 and also the TD core, I would recommend using JSON Schema type consistently.

wiresio:
## Some open issues / next steps

1. Align `map:bitExtract` and `map:bitCompose` with binding-specific byte-order and integer-width metadata.
2. Decide whether a future operator is needed for (a) reinterpreting extracted bits as an IEEE-754 float rather than an unsigned magnitude, and (b) formatting an extracted epoch integer as an application-facing timestamp string. Packed multi-byte string/text fields remain out of scope for `bitExtract`/`bitCompose` and are a payload-binding concern.
3. Track whether a generic, protocol-independent byte-offset/byte-length/type field-extraction mechanism (distinct from `bitExtract`/`bitCompose`'s bit-level masking) should be defined for octet-stream payloads without a native binding equivalent, referencing the SENTRON PAC4200 voltage/timestamp and SITOP "Buffering parameters" examples in [issue #1936](https://github.com/w3c/wot-thing-description/issues/1936).

